Issue #537, opened 2014-12-16
Remove Rock-Paper-Scissors


adamkrellenstein:
- [ ] Disable RPS
  - [ ] Remove most functionality
  - [ ] Hard-code existing games
  - [ ] Update API
  - [ ] Update docs

See #487 for inspiration.
##

ouziel-slama:
done here https://github.com/CounterpartyXCP/counterparty-core/pull/1821

PR #569, opened 2014-12-26
Disable Rock‐Paper‐Scissors with Block ???

adamkrellenstein:
This is the first step in solving #537.


adamkrellenstein:
Replaced by #604, which is simpler.

PR #595, opened 2015-01-05
Unit test additions

Millz0r:
Added several tests, as well as fixing old util function names destroy and send. Also fixed multi-line docstring according to PEP.


adamkrellenstein:
Looks good. (Just a couple mistakes re: `block_index` that'll fix in a second.)

Issue #621, opened 2015-01-11
Excessive logging problem on testnet


rippler:
Using `develop` cloned 3-4 hours ago. I made one send and here's what appeared in the log (the line kept repeating many times until the next block 317849).

```
[INFO] Block: 317847 (1.68s)←[0m
[INFO] Block: 317848 (1.13s)←[0m
[INFO] Send: 0.01 XCP from mvBrgbMAucRUQJFJ4Ddjog6FvL4JDQtoPZ to myG2jGS146Lntch
bNHaAHqPZJyqwqdorUX (141d2517dcef3d95cdc7a88212e58d2a7ecfe6649fa3e01ec9ab3b92876
9020a) [valid]←[0m
[INFO] Send: 0.01 XCP from mvBrgbMAucRUQJFJ4Ddjog6FvL4JDQtoPZ to myG2jGS146Lntch
bNHaAHqPZJyqwqdorUX (141d2517dcef3d95cdc7a88212e58d2a7ecfe6649fa3e01ec9ab3b92876
9020a) [valid]←[0m
[INFO] Send: 0.01 XCP from mvBrgbMAucRUQJFJ4Ddjog6FvL4JDQtoPZ to myG2jGS146Lntch
bNHaAHqPZJyqwqdorUX (141d2517dcef3d95cdc7a88212e58d2a7ecfe6649fa3e01ec9ab3b92876
9020a) [valid]←[0m
[INFO] Send: 0.01 XCP from mvBrgbMAucRUQJFJ4Ddjog6FvL4JDQtoPZ to myG2jGS146Lntch
bNHaAHqPZJyqwqdorUX (141d2517dcef3d95cdc7a88212e58d2a7ecfe6649fa3e01ec9ab3b92876
9020a) [valid]←[0m
[INFO] Send: 0.01 XCP from mvBrgbMAucRUQJFJ4Ddjog6FvL4JDQtoPZ to myG2jGS146Lntch
bNHaAHqPZJyqwqdorUX (141d2517dcef3d95cdc7a88212e58d2a7ecfe6649fa3e01ec9ab3b92876
9020a) [valid]←[0m
[INFO] Send: 0.01 XCP from mvBrgbMAucRUQJFJ4Ddjog6FvL4JDQtoPZ to myG2jGS146Lntch
bNHaAHqPZJyqwqdorUX (141d2517dcef3d95cdc7a88212e58d2a7ecfe6649fa3e01ec9ab3b92876
[...]

PR #623, opened 2015-01-11
[#621] fix excessive logging problem

ouziel-slama:
block_index should be passed to `log.message()` to detect mempool transactions


adamkrellenstein:
@ouziel-slama, couldn't we make this a simple optional flag instead (i.e. `mempool=False`)?

ouziel-slama:
@adamkrellenstein: do you mean something like this:

`log.message(db, 'update', 'bets', bindings, mempool=(True if block_index==config.MEMPOOL_BLOCK_INDEX else False))`
?

adamkrellenstein:
Right, that's worse. Nevermind!

ACK on rebase.

PR #628, opened 2015-01-15
Check for Null Data Chunks with Block ???

adamkrellenstein:
**Untested**


ouziel-slama:
Is that really necessary with this two lines: https://github.com/CounterpartyXCP/counterpartyd/blob/develop/counterpartylib/lib/blocks.py#L581 ?

PR #629, opened 2015-01-16
Fix for overflow in bet.compose and broadcast.compose and new tests

Millz0r:
Fixed integer overflow in deadline and timestamp in bet.compose and broadcast.compose. The struct pack used uint instead of ulonglong.
Also added tests for bet.cancel_bet and broadcast_parse.


Millz0r:
I'll add commit to deal with test errors soon.

Millz0r:
Error is within test suite data, not the codebase.

Issue #632, opened 2015-01-18
"ImportError: cannot import name util" / apsw?


guaka:
While trying out `counterpartyd` I got `ImportError: cannot import name util`.
I was a bit surprised by this. Emptying lib/util.py gave me the same error for api.py, and then for database.py. Finally figured out uncommenting `import apsw` is enough to get a step further.

Now trying to figure out how to install apsw. `sudo pip3 install apsw` gives me some errors though.

I do think it would make sense to mention `apsw` somewhere, e.g. in pip-requirements.txt and/or README.md...

```
$ grep -aire apsw *                                              (git)-[master]
grep: lib/.#__init__.py: No such file or directory
lib/__pycache__/database.cpython-33.pyc:uChecking database integrity.uPRAGMA integrity_checkiuokuIntegrity check failed.u"Could not perform integrity check.FT(loggerudebuguformatuconfigDATABASEusplitupopuapswu
lib/api.py:#import apsw
lib/database.py:# import apsw
lib/database.py:        db = apsw.Connection(config.DATABASE, flags=0x00000001)
lib/database.py:        db = apsw.Connection(config.DATABASE)
lib/util.py:#import apsw
test/conftest.py:import json, binascii, apsw
test/conftest.py:    db = apsw.Connection(util_test.CURR_DIR + '/fixtures/rawtransactions.db')
test/util_test.py:import apsw, pytest, requests
test/util_test.py:    filecon = apsw.Connection(db_filename)
test/util_test.py:    shell = apsw.Shell(stdout=output, args=(db_filename,))
test/util_test.py:    #shell = apsw.Shell(stdout=output, db=db)
test/util_test.py:    # apsw oddness: follwing sentence not always generated!
test/util_test.py:    db = apsw.Connection(database_filename)
test/util_test.py:    prod_db = apsw.Connection(prod_db_path)
```

guaka:
`src/apsw.c:62:2: error: Your SQLite version is too old.  It must be at least 3.8.5`

`brew install sqlite` fixed that.

After that `sudo pip3 install apsw` worked fine.

PR #633, opened 2015-01-18
use pip-requirements.txt - see also #632

guaka:
pip3 install -r pip-requirements.txt


adamkrellenstein:
@guaka, thanks for the PR, but all PRs should be made against the branch `develop` (in which `pip-requirements.txt` no longer exists.

PR #634, opened 2015-01-18
add apsw to pip-requirements.txt #632

guaka:
not sure which version would make sense


adamkrellenstein:
We're not using `pip-requirements.txt` in `develop`.

PR #645, opened 2015-01-20
Improve setup.py

gotcha:
- Enable installation in a virtualenv
- Be more explicit about sudo when installing serpent


adamkrellenstein:
@ouziel-slama?

ouziel-slama:
yeah sure! thanks @gotcha!

PR #646, opened 2015-01-20
Setup for CircleCI

gotcha:
CircleCI is a very good service
Among other things, quicker than TravisCI

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.org/CounterpartyXCP/counterpartyd.svg?branch=develop)](https://travis-ci.org/CounterpartyXCP/counterpartyd)
+[![Build Status](https://circleci.com/gh/CounterpartyXCP/counterpartyd.svg?&style=shield)](https://circleci.com/gh/CounterpartyXCP/counterpartyd)
 [![Coverage Status](https://coveralls.io/repos/CounterpartyXCP/counterpartyd/badge.png?branch=develop)](https://coveralls.io/r/CounterpartyXCP/counterpartyd?branch=develop)
 
 # Description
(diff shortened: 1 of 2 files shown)


gotcha:
This would imply that one member of CounterpartyXCP organization makes a free account at https://circleci.com/ with GH account

adamkrellenstein:
I'm willing to try this out. (Speed might be an issue in the near future, if we're going to be doing full DB reparses: https://github.com/CounterpartyXCP/counterpartyd/issues/642)

@gotcha: can you squash this down to one commit, and change it to use Python 3 (https://circleci.com/gh/CounterpartyXCP/counterpartyd/1)?

gotcha:
@adamkrellenstein There is no way of testing 3 by itself. But they can run tests for multiple versions.
I guess you want 3.2 and 3.3 as well. But not 3.1 ?

adamkrellenstein:
@gotcha, not 2.7? I think Python 3.2, 3.3 and 3.4 would be good.

gotcha:
@adamkrellenstein 
- Why 2.7 ? counterpartyd README explicitly defines Python 3 as a dependency. Is that wrong ?
- I thought CircleCI was able to test multiple versions of Python but it seems I was wrong. Nevertheless, you only test 3.4 on Travis

gotcha:
@adamkrellenstein squashed to one commit

adamkrellenstein:
@gotcha, right now CircleCI is running 2.7, and it's failing... It should be 3.4, say.

gotcha:
@adamkrellenstein I am not sure I understand what you mean. Did you setup on CircleCI ? If yes and build fails, I guess there is an issue with master/develop branch.

I have setup circleCI for counterpartyd develop branch in a copy with my account and it does work:

https://circleci.com/gh/gotcha/testcounter

adamkrellenstein:
Ah, my bad. I was running it without your `circle.yml`, so it was defaulting to Python 2.

Thanks for the PR!

PR #647, opened 2015-01-20
Update `master` to v9.49.4

adamkrellenstein:
This is a lot of commits, and a lot of code movement, but nothing very controversial or likely to cause stability or consensus issues.

The priority is making sure that there is a smooth upgrade path to this version, in particular for those using the build system. **The whole CLI has been moved to https://github.com/CounterpartyXCP/counterparty-cli.**


adamkrellenstein:
@btcdrak, @rippler, @ouziel-slama, @robby-dermody, thoughts?

PR #652, opened 2015-01-22
Update `master` to v9.49.4

adamkrellenstein:
Sorry @ouziel-slama, @btcdrak, @rippler, @robby-dermody, I moved this from #647.

This is a lot of commits, and a lot of code movement, but nothing very controversial or likely to cause stability or consensus issues.

The priority is making sure that there is a smooth upgrade path to this version, in particular for those using the build system. **The whole CLI has been moved to https://github.com/CounterpartyXCP/counterparty-cli.**


counterpartylib/server.py
```diff
@@ -0,0 +1,350 @@
...
+class LockingError(Exception):
+    pass
+def get_lock():
+    logger.info('Acquiring lock.')
+
+    # Cross‐platform.
+    if os.name == 'nt' or platform.system() == 'Darwin':    # Windows or OS X
+        # Not database‐specific.
+        socket_family = socket.AF_INET
+        socket_address = ('localhost', 8999)
```
btcdrak: Maybe this should be configurable?

adamkrellenstein: Good question, but realistically, how many Windows and Mac users are doing something else on port 8999 (that won't patch the code themselves)? (If people complain, we can always add the configuration parameter in later.)

adamkrellenstein:
I think this is ready to be released. Will do as soon as I am able.

Issue #567, opened 2014-12-24
Backend HTTP REST API Support


adamkrellenstein:
That is, be able communicate with the backend over a REST API.
- See https://github.com/bitcoin/bitcoin/pull/2844, merged in Bitcoin Core 0.10.0, and enabled with `-rest`.
- Will btcd have this, too? https://github.com/conformal/btcd/issues/200
- And `searchrawtransactions`?
##

Issue #566, opened 2014-12-24
HTTP REST API


adamkrellenstein:
Copy Bitcoin Core's: https://github.com/bitcoin/bitcoin/pull/2844

Millz0r:
The API is almost complete, it just needs few clarifications:

1. There is a problem making rest at /. The problem is with URL routing in Flask, in order to get a / match you would need to match all URLs (including API) and then send it it's way (so /API/ would also be served from that). I don't think it's a good idea to couple things like that - I think it would be better to keep API at /api/ (although atm it also listens to /, maybe we should add a info page at / saying you should either poll /api/ or /rest/?) and keep REST at /rest/.

2. Mapping URL arguments to get_rows one. get_rows has a lot of arguments, such as filters, filter operators, order, etc. Do we want to support all of them? At the moment I have only seen simplest examples like this:
/rest/balances/address/349sdn84j3b5s9djh4983
As I understand that should resolve to SQL query:
SELECT * FROM balances WHERE address=349sdn84j3b5s9djh4983
Do we want to keep it like that (only WHERE statement match for URLs) or do we want to upgrade to more functionality?
Personally I think we can start with the simple approach, once we release this and get some feedback from people actually using it we might consider extending it.

ouziel-slama:
1. Can you do something like this ?
```
@app.route('/', defaults={'path': ''})
@app.route('/<path:path>')
def catch_all(path):
    if path == '' or path.split('/')[0] == 'api':
        return rpc_query(path)
    else:
        return rest_query(path)
```

2. Yeah, I agree. The `op` operator (OR or AND), seems enough to me for a start.

Millz0r:
1. Yip, I'll do it like that.

2. How would the operator be described - between queries, eg.?
```
/rest/balances/address/349sdn84j3b5s9djh4983/OR/asset/BTC/AND/quantity/2
```
This would be slightly more problematic with current code, but it is doable.

ouziel-slama:
The `op` parameter is for all field/value in the query, you just need it once, for instance like this: 
```
/balances/address/349sdn84j3b5s9djh4983/asset/BTC/quantity/2?op=1
```

PR #686, opened 2015-02-02
Added HTTP REST API. Implements #567 and #566.

Millz0r:
Added support for two REST calls:
- /rest/block/BLOCK-HASH.format
- /rest/tx/TX-HASH.format

Format can be .json for JSON-encoded block data from XCP database; .dat for binary data from backend; and .txt for hex encoded data from backend. The functionality uses the same server code as JSON-RPC API.
Also some minor comment fixes.


counterpartylib/lib/api.py
```diff
@@ -694,6 +746,68 @@ def handle_post():
             _set_cors_headers(response)
             return response
 
+        ######################
+        # HTTP REST API
+        ######################
+        # Handle blocks route.
+        @app.route('/rest/block/<block_hash>.<format>', methods=["GET",])
+        def handle_get_block(block_hash, format):
+            # Check for tx_hash validity.
```
adamkrellenstein: Should say `block_hash` here.

adamkrellenstein: And should mention SQL injection (though the db connection *is* read-only).

adamkrellenstein:
Nice! Very clean code! :)

I'm not actually sure if we should return txs and blocks in `.txt` and `.dat` formats. It's a little surprising  and a little repetetive, IMO, that our JSON blocks, e.g., are different from Bitcoin Core's, but our hex and binary ones aren't? @ouziel-slama, thoughts?

counterpartylib/lib/api.py
Millz0r: True, however an attacker could still read the contents of the whole database.

adamkrellenstein: Yes, but the db contains no private information (I guess it could leak the db file path, or something?). :) With v9.49.4, the server doesn't even know about wallets/privkeys. Of course, it's a good idea in any case.

counterpartylib/lib/api.py
```diff
@@ -284,6 +286,58 @@ def compose_transaction(db, name, params,
...
+    """Return JSON data of specified transaction."""
+    cursor = db.cursor()
+    txs = list(cursor.execute('''SELECT * FROM transactions WHERE tx_hash = ?''', (tx_hash,)))
+    cursor.close()
+    if len(txs) == 0:
+        raise exceptions.DatabaseError('No transaction found for hash %s.' % tx_hash)
+    elif len(txs) > 1:
+        raise exceptions.DatabaseError('More than one transaction found for hash %s.' % tx_hash)
+    tx_info = txs[0]
+    tx_json = json.dumps(tx_info, use_decimal=True)
```
Millz0r: Valid point, I removed it. It seems it's deprecated anyway.

adamkrellenstein:
@Millz0r, feel free to chime in as well, re: what formats you think we should provide.

ouziel-slama:
hum.. sorry but I'am not convinced:

- I agree there is no reason to have an hex output for the block or transaction: it's a backend feature we don't need to reimplement it. 

- IMHO we need to have only Json and XML outputs.

- Methods `get_*_tx` and `get_*_block`, `handle_get_block` and `handle_get_tx` are too much similars. If we want to cover all tables with the REST API, we have to copy/past this code for each table and it's not the good way I think. IMHO the best way to have a DRY code is to use the existing code which generates the functions `get_*` and `create_*`. For this you could for instance:
	
	* create a method that catch all HTTP queries that are not RPC calls.
	* In this method parse the used route and generate the correspondings argument for the functions `api.get_*` or `api.create_*`.

```
GET /tables/{table_name}/{field1}/{value1}/{field2}/{value2}/.../?op=OR

=> api.get_rows(db, table_name, filters=[(field1, '==', value1), (field2, '==', value2), ..], filterop='OR')
```
and
```
GET /transactions/{tx_name}?param1=value1&param2=value2&...

=> api.compose_transaction(db, tx_name, {param1:value1, param2=value, ..})
```

Millz0r:
I'm not sure about which formats we should support. It's true that we are duplicating backend's capabilities for hex and binary data (and actually calling it to get the data). Having it in JSON along with XML could be useful.
@ouziel-slama:
That's actually my plan - to have a generic handler for all REST (or one for everything else than /rest/tx/ and /rest/block) that would then call the dynamically generated get_{} functions. This way we keep JSON-RPC stuff, as well as exposing it through REST without much effort.
The only two problems with that I have are:
- Need some kind of standardization for this kind of query, ie. how to handle fields and operators
```GET /tables/{table_name}/{field1}/{value1}/{field2}/{value2}/.../?op=OR```
- Introspection feature to make this possible in Python. Normally we could use getattr, however the needed get_{} functions are subfunctions of run() and I can't find  a reliable way to get those subfunction objects. Any ideas on that?

IMHO we should keep the /rest/tx/ /rest/block/ APIs I implemented and add a generic handler for all other get_{} functions from JSON-RPC.

Another question - /rest/tx/ and /rest/block/ do not require authorization - should the advanced get_{} functionality be unauthorized too? Or maybe only POST queries should require it?

adamkrellenstein:
Let's just do JSON and XML, then, I think.

I have no idea how to do the introspection.

Why keep the `/rest/tx/` and `/rest/block/`, but generate all of the other functions dynamically? (Maybe this is obvious.)

@ouziel-slama and I think that, unlike Bitcoin Core's REST API, ours (and our JSON-RPC API) should allow for authentication, but just have it off by default, in both cases. (See #687.)

Also, the ChangeLog should be updated in this PR.

ouziel-slama:
> Need some kind of standardization for this kind of query, ie. how to handle fields and operators

IMHO, to be as close as possible in the "spirit" of the REST standard, only fields/value should be part of the url, and other parameters should be get parameters. In general REST URL are in the format:
`/{ressource_name}/{ressource_id}/{linked_ressource_name}/{linked_ressource_id}/../...`
it corresponds "well":
`tables/balances/address/mfh3tGUeeoXfRsqrwL4JSHuS653vu2TLtg`

> Introspection feature to make this possible in Python

I think you don't need to use introspection. You can just use `api.compose_transaction()` and `api.get_rows`. These 2 functions are outside of the `APIServer` class.

Millz0r:
Yeah, all of that makes sense. I'll start working on singlular handler for JSON and XML and open a new PR.

Issue #689, opened 2015-02-03
Remove major version for the database file name


ouziel-slama:
We could maybe just launch a `rollback START_BLOCK` when there is a major version change, like we are launching a `reparse` when there is a minor version change.

adamkrellenstein:
ACK!

adamkrellenstein:
We'd have to store the major version number in the db somehow, however.

ouziel-slama:
yep! so maybe we should include this issue in the `v9.49.4`? then the `setup.py` will move an rename the database directly with the final name.

adamkrellenstein:
Yeah, I guess. Can you implement and test it?

ouziel-slama:
sure!

adamkrellenstein:
Haha! That's an awful hack, but sure. :) It's still much better than using the filename.

ouziel-slama:
:-) ok go for it!

adamkrellenstein:
@ouziel-slama, je crois que tu voulais dire "OK, **let's** go for it". ;)

ouziel-slama:
:-)) yeah thank you! fixed :-)

PR #693, opened 2015-02-03
[#689] Remove major version for the database file name

ouziel-slama:
For review. Not tested yet.

### DIFF
--- a/counterpartylib/lib/blocks.py
+++ b/counterpartylib/lib/blocks.py
@@ -875,19 +875,15 @@ def follow(db):
     else:
         block_index = util.CURRENT_BLOCK_INDEX + 1
         version_major, version_minor = util.get_user_version()
-        version_changed = False
         # Rollback database if major version has changed.
         if version_major != config.VERSION_MAJOR:
             logger.info('Client major version number mismatch ({} ≠ {}).'.format(version_major, config.VERSION_MAJOR))
             reparse(db, block_index=config.BLOCK_FIRST, quiet=False)
-            version_changed = True
+            util.set_user_version(config.VERSION_MAJOR, config.VERSION_MINOR)
         # Reparse all transactions if minor version has changed.
         elif version_minor != config.VERSION_MINOR:
             logger.info('Client minor version number mismatch ({} ≠ {}).'.format(version_minor, config.VERSION_MINOR))
             reparse(db, quiet=False)
-            version_changed = True
-        # Update version if necessary
-        if version_changed:
             util.set_user_version(config.VERSION_MAJOR, config.VERSION_MINOR)
 
         check.version()
--- a/counterpartylib/lib/blocks.py
+++ b/counterpartylib/lib/blocks.py
@@ -882,7 +882,7 @@ def follow(db):
     # Check database version.
     try:
         check.database_version(db)
-    except util.DatabaseVersionError as e:
+    except check.DatabaseVersionError as e:
         logger.info(str(e))
         # no need to reparse or rollback a new database
         if block_index != config.BLOCK_FIRST:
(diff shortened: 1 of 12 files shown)


counterpartylib/lib/blocks.py
```diff
@@ -874,11 +874,21 @@ def follow(db):
         block_index = config.BLOCK_FIRST
     else:
         block_index = util.CURRENT_BLOCK_INDEX + 1
+        version_major, version_minor = util.get_user_version()
+        version_changed = False
```
adamkrellenstein: Do we really need to keep track of whether the version actually changed? Can't we just update `user_version` every time?

counterpartylib/lib/util.py
```diff
@@ -600,4 +600,16 @@ def parse_id(match_id):
...
 
+
+def get_user_version():
+    user_version = cursor.execute('PRAGMA user_version').fetchall()[0]['user_version']
+    version_minor = user_version % 1000
+    version_major = int((user_version - version_minor) / 1000)
+    return version_major, version_minor
+
+def set_user_version(version_major, version_minor):
+    user_version = (int(version_major) * 1000) + int(version_minor)
```
adamkrellenstein: Won't `version_major` and `version_minor` always be `int`s already?

counterpartylib/lib/util.py
```diff
@@ -600,4 +600,16 @@ def parse_id(match_id):
     assert match_id[64] == ID_SEPARATOR
     return match_id[:64], match_id[65:] # UTF-8 encoding means that the indices are doubled.
 
+
+def get_user_version():
+    user_version = cursor.execute('PRAGMA user_version').fetchall()[0]['user_version']
+    version_minor = user_version % 1000
+    version_major = int((user_version - version_minor) / 1000)
```
adamkrellenstein: Could we use `//` here instead?

adamkrellenstein:
This line should be changed as well: https://github.com/CounterpartyXCP/counterpartyd/blob/master/test/util_test.py#L360

adamkrellenstein:
Would it be too much work to add test coverage for this code now, before v9.49.4? Just so we're extra sure it works well. :)

counterpartylib/lib/blocks.py
ouziel-slama: yeah you right, but in this case it becomes difficult to add a message in the log when there is a version change.

counterpartylib/lib/util.py
ouziel-slama: yeah right! no need this `int()`

counterpartylib/lib/util.py
ouziel-slama: yeah!

ouziel-slama:
:-) sure! I will add tests for this.

counterpartylib/lib/blocks.py
adamkrellenstein: Right. I think this is the best way to do it, then: 92bdc45

ouziel-slama:
> This line should be changed as well: https://github.com/CounterpartyXCP/counterpartyd/blob/master/test/util_test.py#L360
already done here https://github.com/CounterpartyXCP/counterpartyd/blob/dbfile/counterpartylib/test/util_test.py#L382

ouziel-slama:
@adamkrellenstein: I "slightly" modified the code. I created a function `check.database_version()`. What do you think to rename `check.version()` to `check.codebase_version()` ?

adamkrellenstein:
`check.software_version()` is better, I think.

ouziel-slama:
@adamkrellenstein: ok tests written and done. I think it's ok to merge.

adamkrellenstein:
My computer's on the fritz. Feel free to merge.

ouziel-slama:
Good luck!

Issue #673, opened 2015-01-26
Make Install Smoother


atweiden:
Recently, PyPi dependency resolution has gotten out of hand. During linux builds, python has been manually downloading and resolving dependency trees when I build counterparty pkgs. I don't know where this strange behavior came from, today is the first time it's happened. I'm having to do an unholy amount of sed to make things work:

```sh
msg 'Fixing setup.py...'
sed -i "/^# In Windows.*/,/^- Serpent.*/d" setup.py
sed -i "s/python-bitcoinlib==.*/python-bitcoinlib'/" setup.py
sed -i "s/json-rpc==.*/json-rpc'/" setup.py
sed -i "s/python-dateutil==.*/python-dateutil'/" setup.py
```

From my perspective it warrants a different approach to linux packaging. I'm thinking it's necessary to package everything inside virtualenvs.

What are your thoughts on using [spotify/dh-virtualenv](https://github.com/spotify/dh-virtualenv) and creating native Debian packages for Counterparty with a [ppa](http://askubuntu.com/questions/71510/how-do-i-create-a-ppa)?

robby-dermody:
+1 for utilizing virtualenv ... I haven't looked much into dh-virtualenv but it looks interesting. .deb packages would be a plus, but there has to be a path for developers working on the source, where they can quickly edit a file (under version control) and test the changes. (note: that's one of the issues with the current situation...if you check out the source, you really need to do a "python setup.py install" again before you test your changes.... not a super big deal, but it's one more step in the edit-build-test loop.

adamkrellenstein:
I don't understand what the problem is here. Do you just want Debian packages from virtualenvs? That's for a Debian package maintainer to take up. What's the cause of the dependency hell? Is it that we're being too specific with the version requirements? (Versions which Debian doesn't like?) I'm OK with loosening up the restrictions there.

@robby-dermody, very often in testing I'll execute the source files in place, instead of installing them first.

atweiden:
By using `pip` or `npm` to fetch and build programming language deps, you don't have to worry at all about dependency hell, and you can still use the native linux package manager for actual installation. A good example is [nodejs-bitcoin](https://github.com/atweiden/pkgbuilds/blob/master/nodejs-sx/PKGBUILD). It specifies [sx](https://github.com/spesmilo/sx) as a dependency while simultaneously allowing `npm` to resolve NPM deps, and the resulting package is tracked for installation/uninstallation by `pacman`.

[...]

PR #701, opened 2015-02-06
solves #673

atweiden:
reference: https://github.com/CounterpartyXCP/counterpartyd/issues/673#issuecomment-72727530


atweiden:
This should've been PR'd against `develop`

PR #702, opened 2015-02-06
make install smoother

atweiden:
for #673


adamkrellenstein:
Breaks Travis with `dateutil` dependency.

Issue #707, opened 2015-02-11
Dividend Fee Check


adamkrellenstein:
> Here, the following check only checks if there is at least DIVIDEND amount of XCP, and FEE amount of XCP, but not whether the balance contains DIVIDEND+XCP.

```
if dividend_asset != config.BTC:
        balances = list(cursor.execute('''SELECT * FROM balances WHERE (address = ? AND asset = ?)''', (source, dividend_asset)))
        if not balances or balances[0]['quantity'] < dividend_total:
            problems.append('insufficient funds ({})'.format(dividend_asset))

    fee = 0
    if not problems and dividend_asset != config.BTC:
        holder_count = len(set(addresses))
        if block_index >= 330000 or config.TESTNET: # Protocol change.
            fee = int(0.0002 * config.UNIT * holder_count)
        if fee:
            balances = list(cursor.execute('''SELECT * FROM balances WHERE (address = ? AND asset = ?)''', (source, config.XCP)))
            if not balances or balances[0]['quantity'] < fee:
                problems.append('insufficient funds ({})'.format(config.XCP))
```

> This will cause a Excepion in the debit class, when too much XCP is being debited than physically in the users account when the account contained exactly DIVIDENT amount of XCP.

Issue #369, opened 2014-10-25
Mainnet Integration Test


adamkrellenstein:
Why wasn't this bug caught by the test suite? 1ca58b2cf87b678359408b9662818d25c90080f1
##

adamkrellenstein:
See https://github.com/CounterpartyXCP/counterparty-core/issues/2167

PR #708, opened 2015-02-11
[#707] Fix Dividend Fee Check

ouziel-slama:


adamkrellenstein:
Should this be a protocol change, or does this patch *only* avoid the triggering of a fatal exception?

ouziel-slama:
hum... good question. I would say the latter, but not 100% sure.. What do you think?

adamkrellenstein:
We need to be 100% sure to merge it now. :-/ If we can't be, then we should just include this in v9.50.0.

ouziel-slama:
hum.. If the case occurs, no transactions will be recorded for the block and following blocks, all `counterparty-server` instances will just crash. So we can not say that is really a protocol change. If you agree with this reasoning so yes I think we can merge now and cherry-pick to master.

adamkrellenstein:
I looked it over carefully, and I think that indeed if and only if this condition is satisfied with the current version, `util.debit()` will throw an uncaught exception. However, I think we should have full test coverage here before this gets merged in.

ouziel-slama:
I added a test for this edge case, and seems there is already a test for all existing errors in `vectors.py`. @Millz0r, can you please confirm we have a test for every kind of error in`dividend.py` ?

ouziel-slama:
OK, I just add test for the last untested case in `dividend.validate().
In `dividend.py` only this case is no tested: https://github.com/CounterpartyXCP/counterpartyd/blob/develop/counterpartylib/lib/messages/dividend.py#L159
AND old protocol. It's a general issue, the test suite only test the last version of the protocol, so this kind of block is never tested: https://github.com/CounterpartyXCP/counterpartyd/blob/develop/counterpartylib/lib/messages/dividend.py#L85

Millz0r:
As far as I can tell there are five unit tests still not covered:
- old protocol mentioned above
- two cases of BTC send in compose and parse
- invalid struct in parse
- one unpack case in parse where len(message) == LENGTH_1 . I'm not sure if this would be affected? https://github.com/CounterpartyXCP/counterpartyd/blob/707/counterpartylib/lib/messages/dividend.py#L152

Besides that I think we are covered well enough.

adamkrellenstein:
None of that's relevant except for the test of the old protocol, which is difficult. :/ (#369)

adamkrellenstein:
@ouziel-slama, can you do the cherry-pick?

ouziel-slama:
@adamkrellenstein, just did a PR on master. I will merge as soon tests will finish on circle.

Issue #687, opened 2015-02-03
Disable Authentication for JSON-RPC API by Default


adamkrellenstein:

Millz0r:
Since I am working on REST/RPC I also added this to my backlog, I got it implemented via a conditional decorator - however what should I base the condition on? Should we add another configuration option, eg. api_auth = 0?

adamkrellenstein:
Shouldn't we just disable authentication if there's no password set?

Millz0r:
That is reasonable, would just need to change that:
https://github.com/CounterpartyXCP/counterpartyd/blob/master/counterpartylib/server.py#L244

PR #712, opened 2015-02-12
[#566] [#687] HTTP REST API

Millz0r:
Solves #566 and #687.

- All API calls go through a single handler now.
- New GET /rest/ call as an interface to get_rows database queries.
- New POST /rest/ call as an interface to compose_transaction calls.
- JSON-RPC doesn't listen on / anymore (that access route not documented anyway).
- RPC_PASSWORD config setting is no longer mandatory.
- Adds xmltodict requirement.

The API is documented here for now:
http://docs.counterpartylib.apiary.io/#

### DIFF
--- a/counterpartylib/lib/api.py
+++ b/counterpartylib/lib/api.py
@@ -31,6 +31,7 @@ from tornado.ioloop import IOLoop
 import jsonrpc
 from jsonrpc import dispatcher
 import inspect
+from xmltodict import unparse as serialize_to_xml
 
 from counterpartylib.lib import config
 from counterpartylib.lib import exceptions
@@ -283,6 +284,14 @@ def compose_transaction(db, name, params,
     # except:
         # import traceback
         # traceback.print_exc()
+
+def conditional_decorator(decorator, condition):
+    """Checks the condition and if True applies specified decorator."""
+    def gen_decorator(f):
+        if not condition:
+            return f
+        return decorator(f)
+    return gen_decorator
 
 def init_api_access_log():
     """Initialize API logger."""
@@ -615,21 +624,50 @@ class APIServer(threading.Thread):
                 response.headers['Access-Control-Allow-Methods'] = 'GET, POST, OPTIONS'
                 response.headers['Access-Control-Allow-Headers'] = 'DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type'
 
-        @app.route('/', methods=["OPTIONS",])
-        @app.route('/api/', methods=["OPTIONS",])
-        def handle_options():
+        @app.route('/', defaults={'args_path': ''}, methods=['GET', 'POST', 'OPTIONS'])
+        @app.route('/<path:args_path>',  methods=['GET', 'POST', 'OPTIONS'])
+        # Only require authentication if RPC_PASSWORD is set.
+        @conditional_decorator(auth.login_required, hasattr(config, 'RPC_PASSWORD'))
+        def handle_root(args_path):
+            """Handle all paths, decide where to forward the query."""
+            if args_path.startswith('api/') or args_path.startswith('API/') or \
+               args_path.startswith('rpc/') or args_path.startswith('RPC/'):
+                if flask.request.method == 'POST':
+                    # Need to get those here because it might not be available in this aux function.
+                    request_json = flask.request.get_data().decode('utf-8')
+                    response = handle_rpc_post(request_json)
+                    return response
+                elif flask.request.method == 'OPTIONS':
+                    response = handle_rpc_options()
+                    return response
+                else:
+                    error = 'Invalid method.'
+                    return flask.Response(error, 405, mimetype='application/json')
+            elif args_path.startswith('rest/') or args_path.startswith('REST/'):
+                if flask.request.method == 'GET' or flask.request.method == 'POST':
+                    # Pass the URL path without /REST/ part and Flask request object.
+                    rest_path = args_path.split('/', 1)[1]
+                    response = handle_rest(rest_path, flask.request)
+                    return response
+                else:
+                    error = 'Invalid method.'
+                    return flask.Response(error, 405, mimetype='application/json')
+            else:
+                # Not found
+                return flask.Response(None, 404, mimetype='application/json')
+
+        ######################
+        # JSON-RPC API
+        ######################
+        def handle_rpc_options():
             response = flask.Response('', 204)
             _set_cors_headers(response)
             return response
 
-        @app.route('/', methods=["POST",])
-        @app.route('/api/', methods=["POST",])
-        @auth.login_required
-        def handle_post():
-
+        def handle_rpc_post(request_json):
+            """Handle /API/ POST route. Call relevant get_rows/create_transaction wrapper."""
             # Check for valid request format.
             try:
-                request_json = flask.request.get_data().decode('utf-8')
                 request_data = json.loads(request_json)
                 assert 'id' in request_data and request_data['jsonrpc'] == "2.0" and request_data['method']
                 # params may be omitted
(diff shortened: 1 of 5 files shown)


adamkrellenstein:
Minor point: the ChangeLog should note that authentication for the JSON-RPC API is disabled by default and the password parameter is no longer required.

The syntax for multiple parameters looks strange to me, but maybe it's the standard way to do it? E.g. `sends/source/mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc/destination/mtQheFaSfWELRB2MyMBaiWjdDm6ux9Ezns/AND` In that case, could `source` and `mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc` be separated by something other than a `/`? Could `source/mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc` and `destination/mtQheFaSfWELRB2MyMBaiWjdDm6ux9Ezns` be separated by a third character as well?

Can we redirect `/rpc/` to `/api/`, so the docs can just list `/rpc/` and `/rest/`?

Otherwise, on a first read-through, it looks pretty good to me!

ouziel-slama:
Some nitpicking:

1) REST url should contains resource names and ID's, and the last part can eventually be an "action". The action parameters should be in the query string. If you consider an address as a resource, we can have something like:
`/rest/source/mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc/compose?destination=mtQheFaSfWELRB2MyMBaiWjdDm6ux9Ezns&quantity=3&asset=XCP`
To be really REST, this query, to compose a transaction, should be used with GET method (POST is used when there is something written in the database). BUT in our case I think we should ALSO accept POST for privacy (POST query string can be crypted with SSL).

2)  `handle_rest_get` and `handle_rest_post` are very similar! Do you think it's possible to DRY a little bit this 2 functions ?

Millz0r:
- I just added more ChangeLog changes.
- Redirection to /rpc/ will be easy, I'll add that today.
- Considering URL building, the more RESTful way would be to keep most data as params, however I thought it would be more readable current way. I agree it's better to stick with the standard and use params as not to surprise users at any time.
- They are indeed very similar, I'll see if I could consolidate it into one function, espcially if we are gonna support GET on both queries.

Millz0r:
@ouziel-slama:
1) What about table-wide selects for get_rows (with no additional arguments)? For instance `/rest/balances/get`, which would return `SELECT * FROM balances`? The resource name would then be either table name or transaction type, and all other arguments would be passed as action parameters.

ouziel-slama:
@Millz0r :
Yes, `GET /rest/balances` should return the first 1000 (the default/max limit is 1000) rows in balances table. 
Also the `filterop` should be in the query string, not in the url: `/rest/sends/source/mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc/destination/mtQheFaSfWELRB2MyMBaiWjdDm6ux9Ezns?op=AND`
In general, only things that can be considered as resource or as resource ID should be part of the URL.

adamkrellenstein:
@ouziel-slama, shouldn't that then be `/rest/sends?source=mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc&destination=mtQheFaSfWELRB2MyMBaiWjdDm6ux9Ezns&op=AND`

@Millz0r, I think that's more readable anyway.

ouziel-slama:
@adamkrellenstein: :-) yeah you right, if we want to be really REST the only resource we have here is the `sends` table.. so the cleanest way is:
`sends/{tx_index or tx_hash}` to get one row, and `sends?source=..&dest=..` to get rows.
... but it's very usual to see REST APIs that passes search parameters in the URL.. I'am not against the two formats of the url.. is up to you and @Millz0r ..

adamkrellenstein:
I strongly prefer keeping the parameters out of the URL.

Millz0r:
Changes above, here are some example URLs that work now:

```
curl -X GET "localhost:14000/rest/send/compose?source=mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc&destination=mtQheFaSfWELRB2MyMBaiWjdDm6ux9Ezns&asset=BTC&quantity=100000"

"01000000025c5ea576702be46253a239bd1aa6b2dbed1f5ace830bbaa9939e80209c76eee4000000001976a9144838d8b3588c4c7ba7c1d06f866e9b3739c6303788acffffffff4986e0b01495bcd00dcf9e5d112c386f5bf0d5d8d926b9811e3ac9d31657f902000000006951210282b886c087eb37dc8182f14ba6cc3e9485ed618b95804d44aecc17c300b585b0210378ee11c3fb97054877a809ce083db292b16d971bcdc6aa4c8f92087133729d8b21037af2e06061b54cdfe3657bbc8496d69000b822e2db0c86ccbe376346a700b83353aeffffffff02a0860100000000001976a9148d6ae8a3b381663118b4e1eff4cfc7d0954dd6ec88ac850ae011000000001976a9144838d8b3588c4c7ba7c1d06f866e9b3739c6303788ac00000000"
```

```
curl -X GET "localhost:14000/rest/balances/get?address=mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc&quantity=10000&op=OR"`

[{"asset": "SLVBARTEN", "quantity": 10000, "address": "n3PQ3R4Hc4C7FnDoYTBSfo3wuVPxXxvKRW"}, {"asset": "BIZBAZQUX", "quantity": 50, "address": "mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc"}, {"asset": "A11001756367432399000", "quantity": 10000, "address": "muv4ujHQmbW8G7YtVmAFboN8Y431ZpvQyK"}]
```

```
curl -H "Accept: application/xml" -X GET "localhost:14000/rest/balances/get?address=mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc&quantity=10000&op=OR"

<?xml version="1.0" encoding="utf-8"?>
<balances><item><asset>SLVBARTEN</asset><quantity>10000</quantity><address>n3PQ3R4Hc4C7FnDoYTBSfo3wuVPxXxvKRW</address></item><item><asset>BIZBAZQUX</asset><quantity>50</quantity><address>mn6q3dS2EnDUx3bmyWc6D4szJNVGtaR7zc</address></item><item><asset>A11001756367432399000</asset><quantity>10000</quantity><address>muv4ujHQmbW8G7YtVmAFboN8Y431ZpvQyK</address></item></balances>
```

counterpartylib/server.py
```diff
@@ -210,6 +210,9 @@ def initialise(database_file=None, log_file=None, api_log_file=None,
     else:
         config.RPC_HOST = 'localhost'
 
+    # The web root directory for API calls, eg. localhost:14000/api/
+    config.RPC_WEBROOT = '/api/'
```
adamkrellenstein: This should be changed to `/rpc/`.

counterpartylib/server.py
```diff
@@ -240,10 +243,9 @@ def initialise(database_file=None, log_file=None, api_log_file=None,
     #  counterpartyd API RPC password
     if rpc_password:
         config.RPC_PASSWORD = rpc_password
+        config.RPC = 'http://' + urlencode(config.RPC_USER) + ':' + urlencode(config.RPC_PASSWORD) + '@' + config.RPC_HOST + ':' + str(config.RPC_PORT) + config.RPC_WEBROOT
     else:
-        raise ConfigurationError('RPC password not set. (Use configuration file or --rpc-password=PASSWORD)')
-
-    config.RPC = 'http://' + urlencode(config.RPC_USER) + ':' + urlencode(config.RPC_PASSWORD) + '@' + config.RPC_HOST + ':' + str(config.RPC_PORT)
+        config.RPC = 'http://' + config.RPC_HOST + ':' + str(config.RPC_PORT) + config.RPC_WEBROOT
```
adamkrellenstein: We need to use `urlencode` here, too. (as just above)

counterpartylib/lib/api.py
```diff
@@ -653,6 +691,101 @@ def handle_post():
...
+                return flask.Response(error, 400, mimetype='text/plain')
+
+            # Get all arguments passed via URL.
+            url_args = path_args.split('/')
+            try:
+                query_type = url_args.pop(0).lower()
+            except IndexError:
+                error = 'No query_type provided.'
+                return flask.Response(error, 400, mimetype='text/plain')
+            # Check if message type or table name are valid.
```
ouziel-slama: Error should be returned in Json by default.

counterpartylib/lib/api.py
```diff
@@ -615,21 +624,50 @@ def _set_cors_headers(response):
...
             _set_cors_headers(response)
             return response
 
-        @app.route('/', methods=["POST",])
-        @app.route('/api/', methods=["POST",])
-        @auth.login_required
-        def handle_post():
-
+        def handle_api_post(request_json):
+            """Handle /API/ POST route. Call relevant get_rows/create_transaction wrapper."""
```
ouziel-slama: I think should be renamed to `handle_rpc_post` or `handle_rpc`.

counterpartylib/lib/api.py
```diff
@@ -653,6 +691,101 @@ def handle_post():
...
+                except (script.AddressError, exceptions.ComposeError, exceptions.TransactionError, exceptions.BalanceError) as error:
+                    error_msg = str(error.__class__.__name__) + ': ' + str(error)
+                    return flask.Response(error_msg, 400, mimetype='text/plain')                        
+            else:
+                # Need to de-generate extra_args to pass it through.
+                query_args = dict([item for item in extra_args])
+                try:
+                    operator = query_args.pop('op')
+                except KeyError:
+                    operator = 'AND'
```
ouziel-slama: Can be replaced by : `query_args.pop('op', 'AND')`

adamkrellenstein:
ACK, with the caveat that the docs say the API isn't yet finalized.

@ouziel-slama?

adamkrellenstein:
Also, needs a rebase.

ouziel-slama:
ACK

adamkrellenstein:
Merged here: https://github.com/CounterpartyXCP/counterpartyd/commit/0695ac8f645411767d71b11327f323467254866e

PR #725, opened 2015-02-22
Check if `parsed_source` is not a 1 of X multisig address

### LINKED ISSUE #721: `Constructed transaction does not parse correctly`
<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8779523-constructed-transaction-does-not-parse-correctly?utm_campaign=plugin&utm_content=tracker%2F494307&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F494307&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ouziel-slama:
Inelegant but quick and fast way to fix #721 : because the new `is_scriptpubkey_spendable()`, `parsed_source` can be a 1 of X multisig address.
Note: maybe there is several cases to get the `Constructed transaction does not parse correctly` error, but this patch fixes at least one of them.

### DIFF
--- a/counterpartylib/lib/transaction.py
+++ b/counterpartylib/lib/transaction.py
@@ -481,7 +481,7 @@ def construct (db, tx_info, encoding='auto',
     # Check desired info against parsed info.
     desired = (desired_source, desired_destination, desired_data)
     parsed = (parsed_source, parsed_destination, parsed_data)
-    if desired != parsed:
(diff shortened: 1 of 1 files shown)


adamkrellenstein:
I'd rather we have a cleaner, more elegant fix than this. There's no rush with this (`is_vout_spendable()` is only in `develop`).

PR #728, opened 2015-03-05
Some refactoring to make possible transactions checking client-side

ouziel-slama:
Move sub functions outside `get_tx_info2()` because `counterpary-cli` needs the function `decode_output()` to check transactions generated by a server.

Add `raw_unpack_data()` method. This method don't use the database or the backend.
It is used by `counterparty-client/gui` to check transactions generated by `counterparty-server`.

see: https://github.com/CounterpartyXCP/counterparty-cli/pull/34


counterpartylib/lib/api.py
```diff
@@ -285,6 +285,36 @@ def compose_transaction(db, name, params,
...
+                    'message': message_type
+                }
+                if hasattr(message_module, 'FORMAT'):
+                    format = message_module.FORMAT() if callable(message_module.FORMAT) else message_module.FORMAT
+                    unpacked = struct.unpack(format, message)
+                    for p in range(len(unpacked)):
+                        param_name = message_module.FORMAT_DESCRIPTION[p]
+                        params[param_name] = unpacked[p]
+                    return params
+    except Exception as e:
```
adamkrellenstein: This line is no good.

ouziel-slama: why? here I want to catch all possible errors. No matter what type of error, if there is one, that means that the data is incorrect.

adamkrellenstein: If there's a filesystem read error, e.g., that's not a problem with the unpacking, and it doesn't mean that the data is invalid.

ouziel-slama: yeah you are right! I will fix that.

ouziel-slama:
I moved transaction checking from `counterparty-cli` to `counterparty-lib`. After several attempts I have not managed to use the same functions at the end of `transactions.construct()`. Checks are too different in the two cases, and IMO wanting to absolutely use the same functions sounds like a solution in search of a problem.

counterpartylib/lib/transaction.py
```diff
@@ -487,4 +491,199 @@ def construct (db, tx_info, encoding='auto',
...
+            if invalid_param:
+                raise exceptions.TransactionError('constructed transaction contains invalid parameter: {}'.format(param))
+
+        # TODO: check all messages
+        # check destination
+        if asked_message_type == 'bet':
+            params['destination'] = params['feed_address']
+
+        # No destination
+        if asked_message_type in ['order', 'dividend', 'broadcast', 'cancel', 'destroy', 'execute', 'publish', 'rps', 'rpsresolve']:
```
adamkrellenstein: Can't we do this whole section more automatically with `FORMAT_DESCRIPTIONS`?

counterpartylib/lib/transaction.py
```diff
@@ -487,4 +491,199 @@ def construct (db, tx_info, encoding='auto',
...
+            'give_id': 'give_asset', 
+            'get_id': 'get_asset',
+            'dividend_asset_id': 'dividend_asset'
+        }
+        for asset_param in asset_param_list:
+            if asset_param in data:
+                asset_id = data.pop(asset_param)
+                param_name = asset_param_list[asset_param]
+                try:
+                    data[param_name] = util.generate_asset_name(asset_id, block_index=333501)
```
adamkrellenstein: Why do we need to try both ways? Shouldn't all newly created transactions use the new asset ids?

counterpartylib/lib/transaction.py
```diff
@@ -487,4 +491,199 @@ def construct (db, tx_info, encoding='auto',
...
+                    raise exceptions.TransactionError('constructed transaction contains invalid data outputs')
+                outputs_info['data_output_type'] = output_type
+        outputs_info['total_value'] += vout.nValue
+
+    return outputs_info
+
+def check_outputs(method, params, tx_hex):
+    tx_info = get_outputs_info(tx_hex)
+    #print(tx_info)
+    asked_message_type = method.split('_')[1]
```
adamkrellenstein: Can we call this `desired_message_type`?

ouziel-slama:
@adamkrellenstein, there is problem to check `btcpay` transaction without database: it's impossible to get the address and amount corresponding to the order hash without querying the database.
To enable full client side validation, maybe we should add two mandatories parameters for `btcpay`: `destination` and `amount`. 
Or we can show for confirmation to the user, these 2 parameters when the transaction is generated.
thought?

counterpartylib/lib/transaction.py
adamkrellenstein: Or rather


And create new variables `parsed_destinations = tx_info['destinations'].keys()` and `desired_destination = params[destination']`, I think.

counterpartylib/lib/transaction.py
```diff
@@ -487,4 +491,199 @@ def construct (db, tx_info, encoding='auto',
...
+        elif tx_info['data_output_type'] == 'OP_CHECKMULTISIG':
+            dust_size = params['multisig_dust_size'] if 'multisig_dust_size' in params and params['multisig_dust_size'] else config.DEFAULT_MULTISIG_DUST_SIZE
+            max_data_value = tx_info['data_output_count'] * dust_size
+        if tx_info['data_value'] > max_data_value:
+            raise exceptions.TransactionError('invalid dust quantity for data outputs')
+
+    else:
+        # only BTC sends don't contain data
+        if (method != 'create_burn') and (method != 'create_send' or params['asset'] != 'BTC'):
+            raise exceptions.TransactionError('constructed transaction does not contains data')
```
adamkrellenstein: `does not contain data` (typo)

counterpartylib/lib/transaction.py
ouziel-slama: hum.. seems difficult.. I need yet to well test each transaction.. when done I will give a try.

ouziel-slama: EDIT: I think the code changed since you message, not sure if I understood well your remark.

counterpartylib/lib/blocks.py
```diff
@@ -491,59 +561,6 @@ def get_tx_info2(tx_hex, block_parser=None):
     # Decode transaction binary.
     ctx = backend.deserialize(tx_hex)
 
-    def arc4_decrypt(cyphertext):
```
adamkrellenstein: This change is significant enough that we need to do a complete `kickstart` to make sure it doesn't change the way any transactions are parsed.

counterpartylib/lib/transaction.py
ouziel-slama: hum.. I don't think, for instance we can have a `send` transaction for an "old" asset.

counterpartylib/lib/transaction.py
ouziel-slama: sure!

counterpartylib/lib/transaction.py
```diff
@@ -487,4 +491,199 @@ def construct (db, tx_info, encoding='auto',
 
     return unsigned_tx_hex
 
+
+"""
+Client side checking: never uses database or backend.
+"""
+
+def raw_unpack_data(data_hex):
+    from counterpartylib.lib.api import API_TRANSACTIONS
```
adamkrellenstein: Let's add a detailed docstring to this function.

counterpartylib/lib/transaction.py
```diff
@@ -487,4 +491,199 @@ def construct (db, tx_info, encoding='auto',
...
+                    format = message_module.FORMAT(message) if callable(message_module.FORMAT) else message_module.FORMAT
+                    unpacked = struct.unpack(format, message)
+                    for p in range(len(unpacked)):
+                        param_name = message_module.FORMAT_DESCRIPTION[p]
+                        params[param_name] = unpacked[p]
+                    return params
+    except struct.error as e:
+        raise exceptions.UnpackError(e)
+
+    raise exceptions.UnpackError("Invalid data")
```
adamkrellenstein: Under what circumstances is this exception raised?

counterpartylib/lib/blocks.py
ouziel-slama: yep!

counterpartylib/lib/transaction.py
```diff
@@ -487,4 +491,199 @@ def construct (db, tx_info, encoding='auto',
...
+
+        # check parameters
+        optional_params = {
+            'call_price': 0.0,
+            'description': b'',
+            'call_date': 0,
+            'callable_': False,
+            'divisible': True
+        }
+        for param in data:
```
adamkrellenstein: This block could be clearer, I think.

counterpartylib/lib/transaction.py
```diff
@@ -487,4 +491,199 @@ def construct (db, tx_info, encoding='auto',
...
+        if (method != 'create_burn') and (method != 'create_send' or params['asset'] != 'BTC'):
+            raise exceptions.TransactionError('constructed transaction does not contains data') 
+        if method == 'create_burn':
+            params['destination'] = config.UNSPENDABLE
+
+        # check change address (change in destinations)
+        if tx_info['change'][0] is not None:
+            raise exceptions.TransactionError('invalid change address')
+
+        # check destination
```
adamkrellenstein: All of these comments need to be more descriptive. Check destination *how*, *why*, *against what*?

adamkrellenstein: The error messages need to be more specific. 'Invalid how'?

counterpartylib/lib/transaction.py
ouziel-slama: if there is no return before, that means if there is no corresponding message ID.

counterpartylib/lib/transaction.py
adamkrellenstein: Ah, right. (I made that mistake before.)

adamkrellenstein:
You mean mandatory parameters to the `create_btcpay` call? Yes, I'm fine with adding those.

counterpartylib/lib/transaction.py
ouziel-slama: done.

ouziel-slama:
yep! OK.

counterpartylib/lib/transaction.py
ouziel-slama: fixed

counterpartylib/lib/transaction.py
adamkrellenstein: 'k, then the exception should say `unknown message ID`.

ouziel-slama: done

counterpartylib/lib/transaction.py
ouziel-slama: I added more verbose comment and error message.

counterpartylib/lib/transaction.py
ouziel-slama: I broke the block in several checks.

adamkrellenstein:
@ouziel-slama, what's the status of this?

ouziel-slama:
@adamkrellenstein, the codebase is finished, but I need to add unit test for every RPC call.

ouziel-slama:
@adamkrellenstein, there is now enough tests for RPC calls... however the `kickstart` failed, not sure it's link with this PR, but I will check that tonight.

adamkrellenstein:
@ouziel-slama, why/how did the kickstart fail? Are we sure it won't be an intermittent problem?

ouziel-slama:
@adamkrellenstein: the parsing didn't stop at BLOCK_START, he continued until block 1.. I don't understand how it's happened (that mean that the kickstart just skip the first block, see here: https://github.com/CounterpartyXCP/counterpartyd/blob/master/counterpartylib/lib/blocks.py#L835)
I ran two kickstarts without problems.. I will launch more to be sure.

adamkrellenstein:
Created this issue #749.

chiguireitor:
This Pr although useful, has diverged too much from the codebase, closing this and saving on branch pr728

PR #732, opened 2015-03-18
HOTFIX: Check maximum quantity also for invalid `send`

ouziel-slama:
@adamkrellenstein, IMO this is not a protocol change, but needs a mandatory update. What do you think the cleanest way to merge this one in `master` ?


adamkrellenstein:
@ouziel-slama: as long as every client is crashed because of the bug, it's OK to just merge it and release it as v9.49.5.

Are you OK with that?

adamkrellenstein:
Or we could release it as v9.50.0 with a mandatory protocol change whose minimum block number is 0.

PR #750, opened 2015-04-10
Batch query to get transactions hex

ouziel-slama:


counterpartylib/lib/backend/__init__.py
```diff
@@ -34,6 +34,8 @@ def searchrawtransactions(address, unconfirmed=False):
     return BACKEND().searchrawtransactions(address, unconfirmed=unconfirmed)
 def getrawtransaction(tx_hash, verbose=False):
     return BACKEND().getrawtransaction(tx_hash, verbose=verbose)
+def getrawtransactions(txhash_list, verbose=False):
```
adamkrellenstein: This is potentially confusing. Let's call this `getrawtransaction_batch`.

counterpartylib/lib/config.py
```diff
@@ -88,7 +88,7 @@
 
 
 DEFAULT_REQUESTS_TIMEOUT = 20  # 20 seconds
-
+DEFAULT_RPC_BATCH_SIZE = 300
```
adamkrellenstein: How did we arrive at this number? Trial and error?

counterpartylib/lib/backend/__init__.py
ouziel-slama: :-) done.

counterpartylib/lib/config.py
ouziel-slama: It's a completely arbitrary number. Batch query works well even with 2425 transactions in the block (the max I saw). This parameter is important for people who are using Bitcoin Core with another application (like Counterblock) and should be tweaked with the `rpcthreads` Bitcoin core parameter.

adamkrellenstein: Shouldn't we then have the default value be bigger than the max number of transactions in a block? (So most users only see one query per block.) Except that we tell everyone to set `rpcthreads` to 1000. Should we increase that number? (It seems right now we should just set this to 1000. Then all *current* blocks will be done in one query: https://blockchain.info/charts/n-transactions-per-block)

ouziel-slama: yeah make sense! I increased the default value to 1000. I don't think it's necessary to tell to everyone to set `rpcthreads` to 1000: the default value is 3, and even if the batch query takes time, there is two other threads.

ouziel-slama: Most important is to have this parameter in case of..

adamkrellenstein: Hm, actually, should we just make this 5000 (I think a 1MB block can hold about 4200 transactions), so we won't have to increase it if blocks start filling up more?

Issue #752, opened 2015-04-13
Slow Asset Conservation Check


adamkrellenstein:
See https://github.com/CounterpartyXCP/counterpartyd/issues/748.

adamkrellenstein:
Solved: https://github.com/CounterpartyXCP/counterpartyd/pull/754

PR #754, opened 2015-04-13
[#752] Optimize and make optional asset conservation checking

ouziel-slama:
In my configuration, with the new method, `check.asset_conservation()` takes less than 2s instead more than 155s.

### DIFF
--- a/counterpartylib/lib/blocks.py
+++ b/counterpartylib/lib/blocks.py
@@ -1010,8 +1010,8 @@ def follow(db):
                 parse_block(db, block_index, block_time)
 
             # When newly caught up, check for conservation of assets.
-            if block_index == block_count:
-                check.asset_conservation(db)
+            if block_index == block_count and config.CHECK_ASSET_CONSERVATION:
+              check.asset_conservation(db)
 
             # Remove any non‐supported transactions older than ten blocks.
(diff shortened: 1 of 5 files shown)


adamkrellenstein:
Awesome

deweller:
woo hoo!

mtbitcoin:
woot !

mtbitcoin:
Can you clarify on the "make optional asset conservation checking" ?? 

What is the conservation checking for ?

adamkrellenstein:
@mtbitcoin, it's a sanity check to make sure that the number of assets in balances and in escrow sum to the total number issued. If there were ever a mistake in the protocol that deleted on created an asset unintentionally, clients running the conservation check would halt with an error.

mtbitcoin:
ok. Noted.

As a tradeoff, have you considered just running the sanity checks say after XX number of block (once a day, etc) ?

side note: I have found that with the conservation checks off, the mempool transactions shows up really quick and works well for checking the pending transactions. I've always had this issue (which wasn't an issue in ealier on).  So in hindsight, it looks like the issue might have been due to the sanity check and increase volume of assets registered instead

adamkrellenstein:
There's no best option. The optimal setting depends on the particular use case. Feel free to patch the code to run it only every, say, 100 blocks. I think the likelihood of the sanity check catching an error is pretty low.

PR #758, opened 2015-04-14
Faster Rollbacks with Savepoints

adamkrellenstein:
A little kludgy, but simple enough to be reliable and maintainable, I think.

Not sure if this will be a memory/storage hog. All savepoints are currently released 1) after every reorg. and 2) every `N = 100` blocks.

This will make reorgs. instantaneous most of the time. Reorgs. immediately following the clearing of all savepoints will work as they always have before.


adamkrellenstein:
Savepoints prevent the API from being up-to-date.

PR #764, opened 2015-04-20
Update `master` to v9.51.1

adamkrellenstein:
Not mandatory. Nothing controversial. Most notably includes bug fixes and major performance improvements.


adamkrellenstein:
I'd like to get this out ASAP. @ouziel-slama, thoughts?

ouziel-slama:
@adamkrellenstein: just need to make more tests with BTCD and it's ok for me.

ouziel-slama:
OK, I fixed the BTCD wrapper (unfortunately BTCD don't manage batch RPC, see https://github.com/btcsuite/btcd/issues/329). Also I improved the system to clean the raw transactions cache and avoid memory leak (simple FIFO).

PR #777, opened 2015-05-10
Always default to multisig when encoding == 'auto'

### LINKED ISSUE #730: RPS in REST API
Since RPS was deprecated, it might be a good idea to remove it in the future (or document that it's deprecated)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9654443-rps-in-rest-api?utm_campaign=plugin&utm_content=tracker%2F494307&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F494307&utm_medium=issues&utm_source=github).
</bountysource-plugin>

arhuaco:
In my tests this fixes https://github.com/CounterpartyXCP/counterwallet/issues/730.
Probably useful as workaround.

### DIFF
--- a/counterpartylib/lib/transaction.py
+++ b/counterpartylib/lib/transaction.py
@@ -329,10 +329,7 @@ def construct (db, tx_info, encoding='auto',
     # Data encoding methods (choose and validate).
     if data:
         if encoding == 'auto':
-            if len(data) <= config.OP_RETURN_MAX_SIZE:
-                encoding = 'opreturn'
-            else:
-                encoding = 'multisig'
+            encoding = 'multisig'
         elif encoding not in ('pubkeyhash', 'multisig', 'opreturn'):
             raise exceptions.TransactionError('Unknown encoding‐scheme.')
 


btcdrak:
That doesn't make sense. Why on earth would you default to multisig when opreturn is desirable whenever possible?

btcdrak:
NACK

arhuaco:
Hi there. Two notes:

1) I hope this helps debug #730 . With this change I am not longer getting the error {"result":null,"error":{"code":-26,"message":"64: scriptpubkey"},"id":0}.

I was stuck unable to perform operations. So this was useful to me.

2) The documentation should be updated because it says "at this time, auto is effectively the same as multisig", and it is not currently true. It should say that Counterparty it chooses the most efficient encoding available.

arhuaco:
I mean, this issue:

https://github.com/CounterpartyXCP/counterwallet/issues/730

It also affects counterparty-cli, which I think does not allow to set the encoding.

ouziel-slama:
Problems with OP_RETURN should be fixed here: https://github.com/CounterpartyXCP/counterparty-lib/commit/8e3590ee5295cc646793ba8bbbd28bcb79ddb43a
40 bytes is not enough for cancel order.

PR #782, opened 2015-07-04
Enable API During Reparses

adamkrellenstein:
**Untested**


adamkrellenstein:
@deweller, can you try this out?

adamkrellenstein:
Ah, just did some testing myself... this approach won't work.

adamkrellenstein:
Something similar might work. The reparse would have to be completely atomic, or the database that the API connects to would have to be frozen for the duration of the reparse. (The issue is that while the reparse works through the entire history of Counterparty, the API returns *that* data, i.e. the very old data.)

robby-dermody:
no longer necessary. closing

PR #793, opened 2015-09-07
Increase OP_RETURN to network defaults of 80 bytes

btcdrak:


adamkrellenstein:
Do 80-byte `OP_RETURN` transactions get relayed and mined as fast as 40-byte ones?

This breaks the tests.

btcdrak:
@adamkrellenstein One easy way to find out. Make two separate TXs (not linked in any way), one with a 40 byte OP return and the other with an 80 and see what happens.

robby-dermody:
This is pending additional testing as to how rapidly these > 40 byte OP_RETURNS are mined on mainnet

btcdrak:
I suggest, wait until CLTV gets to around 90% deployment, that will mean 90% of the hash power are running 0.11.2 which has the 80 byte policy.

You can check uptake here: https://data.bitcoinity.org/bitcoin/block_version/5y?c=block_version&r=week&t=a

btcdrak:
OT: I am sorry I dont have time to fix tests etc. I was just bringing this to attention because I am concerned CP use the most efficient encoding rather than all these workarounds. I think it's probably not necessary to obfuscate outputs after this either, because the landscape has changed a lot and OP_RETURN is much more accepted for meta data.

rubensayshi:
ping @robby-dermody @adamkrellenstein @btcdrak 

I think it would be safe to make the switch now.
Any idea on what kind of testing you guys want to see before making this change? (or not?)

I can update the tests to work with 80 bytes if that's all that is blocking this.

btcdrak:
All pools have upgraded now so you absolutely should switch to OP_RETURN 80.

robby-dermody:
@rubensayshi yes please create/update tests and we will merge

robby-dermody:
closing, since #828 was merged in

PR #798, opened 2015-09-23
Update fixtures

ouziel-slama:
@robby-dermody, I updated here the fixtures. 3 tests still don't pass.. trying to fix them..


robby-dermody:
excellent, thanks @ouziel-slama

PR #799, opened 2015-10-06
custom_inputs parameter for controlling exact utxos to use in transactions

cryptonaut420:
expects an array of same data as retrieved by bitcoind's listunspent() function


ouziel-slama:
+1!! definitely the faster way to generate a transactions! Even perhaps that should be the default mode..
(could you please make the PR on `develop` branch?)

adamkrellenstein:
ACK

Isn't this already made against `develop`?

> expects an array of same data as retrieved by bitcoind's listunspent() function

Can that be documented in the code?

cryptonaut420:
Yeah, the PR is to the develop branch but its coming from my master branch, mostly because I messed up and was trying not to include some other unnecessary commits lol

ouziel-slama:
oh! sorry I misread!

PR #800, opened 2015-10-10
Undolog support

robby-dermody:
Have done a good deal of testing (numerous rollbacks, running about a week on mainnet and testnet, comparing database table dumps). looks good. I feel pretty confident about the method as well (simple SQL triggers that work on all applicable tables -- nothing more fancy or table-specific than that).

test suite is still a bit broken @ouziel-slama will need to fix before merge


counterpartylib/lib/blocks.py
```diff
@@ -340,6 +363,35 @@ def initialise(db):
...
+    cursor.execute('''CREATE TABLE IF NOT EXISTS undolog(
+                        seq INTEGER PRIMARY KEY AUTOINCREMENT,
+                        sql TEXT)
+                   ''')
+    cursor.execute('''CREATE TABLE IF NOT EXISTS undolog_block(
+                        block_index INTEGER PRIMARY KEY,
+                        first_seq INTEGER)
+                   ''')
+    # Create undolog triggers
+    for table in TABLES + ['balances']:
```
adamkrellenstein: Why does `balances` have to be added here manually?

counterpartylib/lib/blocks.py
```diff
@@ -653,40 +703,105 @@ def reinitialise(db, block_index=None):
...
+                    return block_index - 1
+            else:
+                return next(reversed(seqs)) #the last inserted block_index
+
+        with db:
+            # Check if we can reparse from the undolog
+            seqs_rows = list(undolog_cursor.execute(
+                '''SELECT block_index, first_seq FROM undolog_block WHERE block_index >= ? ORDER BY block_index ASC''', (block_index,)))
+            seqs = collections.OrderedDict()
+            for e in seqs_rows:
```
adamkrellenstein: this is a bit opaque

what's `seqs`?

counterpartylib/lib/blocks.py
```diff
@@ -340,6 +363,35 @@ def initialise(db):
...
+                        sql TEXT)
+                   ''')
+    cursor.execute('''CREATE TABLE IF NOT EXISTS undolog_block(
+                        block_index INTEGER PRIMARY KEY,
+                        first_seq INTEGER)
+                   ''')
+    # Create undolog triggers
+    for table in TABLES + ['balances']:
+        columns = [column['name'] for column in cursor.execute('''PRAGMA table_info({})'''.format(table))]
+        cursor.execute('''CREATE TRIGGER IF NOT EXISTS _{}_it AFTER INSERT ON {} BEGIN
```
adamkrellenstein: `_it` -> `_insert_?`, etc.

adamkrellenstein:
This is simple enough and well-enough tested I think this can be merged into `develop` pretty soon.

The biggest issue is just that some variables could have longer and more perspicacious names.

robby-dermody:
@adamkrellenstein @ouziel-slama adjusted naming (I like this better too), but since some of the SQL trigger and column names changed, it broke the tests again. ouziel will need to fix until he and I can sit down and troubleshoot my testing environment and why it can't properly update tests. I'll doublecheck his revised test output to make sure it looks good

rubensayshi:
without looking at the code I figured I'd mention; during some 'block storms' (when diff resets to 1) on testnet I've seen reorgs of up to 30 blocks.

I guess it's fine to just require a full rebuild in that case, but keep in mind this does happen (on testnet).

adamkrellenstein:
@rubensayshi thanks for the heads-up. @robby-dermody if it doesn't require a significant amount of space to store the undo log (and I don't know why it would!) we can increase the limit to, say 100 blocks.

robby-dermody:
@rubensayshi @adamkrellenstein The additional size requirements are trivial. Just committed a change to increase to 100 blocks.

robby-dermody:
@adamkrellenstein are we good to merge?

PR #802, opened 2015-10-22
9.52.0

robby-dermody:
See `ChangeLog.md`.

I just checked the consensus hashes on `cw01` (after letting it run more than a week) and compared them to those of a node running 9.51.4 for the latest block. They match.

Let's discuss what additional testing (if any) we need to perform to get this pushed to `master`.


adamkrellenstein:
I think if this thing is running for a few weeks (and can do from-scratch reparses) without a consensus hash mismatch, then it's pretty safe.

The question is just which hashes are we checking, and are they sufficient? AFAIK, the two we have are enough! (If there were a problem, it would show up just in `txlist_hash`, because transactions would be retrieved out of order.)

robby-dermody:
@adamkrellenstein @ouziel-slama I'm feeling pretty confident about this patch. I have it running on `cw01` and my dev box. After multiple mainnet reorgs, several 90+ block rollbacks, and a full reparse operation, all hashes still match up, and match the hashes of the current `master`. From the message feed hash, all database tables are matching up as well.

I think we're ready to merge into `master`, after @ouziel-slama is able to get to fixing the circleci testsuite run so it passes.

ouziel-slama:
Tests fixed. Was just a typo error.

PR #803, opened 2015-10-27
Adding message hash functionality into `develop`

robby-dermody:
Alright, this should enhance `develop` so that for every block, a hash is calculated from database updates (this is calculated in `database.rowtracer()`). Should allow us to be more confident with `undolog`, and anything related in the future.

Why is this good/necessary? `ledger_hash` only really covers debits and credits. In contrast, message hashes from different clients can be compared to concisely spot differences/diversion in essentially any table in the underlying database, beyond just the `transactions` table (`txhash`) and debits and credits (`ledger_hash`).

Currently, I'm not _enforcing_ this hash via checkpoints and other mechanics we use in the `check` module when we parse a new block, but it is being generated and shown for informational purposes. I think this is a good, conservative choice, given that this hash could change if, for instance, we add an additional data parameter to one of the message types (`send`), which would probably not affect "consensus", per se. Moreover, if the hash overrides a different value for the same block (e.g. on the reparse), that will be shown in the parse output for the new block. 

In addition I found and fixed a few minor bugs in `blocks`.


counterpartylib/lib/blocks.py
```diff
@@ -179,10 +182,11 @@ def parse_block(db, block_index, block_time, previous_ledger_hash=None,
...
 
     # Calculate consensus hashes.
-    new_txlist_hash = check.consensus_hash(db, 'txlist_hash', previous_txlist_hash, txlist)
-    new_ledger_hash = check.consensus_hash(db, 'ledger_hash', previous_ledger_hash, util.BLOCK_LEDGER)
+    new_txlist_hash, found_txlist_hash = check.consensus_hash(db, 'txlist_hash', previous_txlist_hash, txlist)
+    new_ledger_hash, found_ledger_hash = check.consensus_hash(db, 'ledger_hash', previous_ledger_hash, util.BLOCK_LEDGER)
+    new_messages_hash, found_messages_hash = check.consensus_hash(db, 'messages_hash', previous_messages_hash, database.BLOCK_MESSAGES)
 
-    return new_ledger_hash, new_txlist_hash
+    return new_ledger_hash, new_txlist_hash, new_messages_hash, found_messages_hash
```
adamkrellenstein: What's with the asymmetry here? Why do we need to return `found_messages_hash`?

counterpartylib/lib/config.py
```diff
@@ -6,8 +6,8 @@
 
 # Versions
 VERSION_MAJOR = 9
-VERSION_MINOR = 51
-VERSION_REVISION = 5
+VERSION_MINOR = 52
```
adamkrellenstein: v9.51.5

ChangeLog.md
```diff
@@ -1,9 +1,10 @@
 ## Client Versions ##
-* v9.51.5 (2015-10-22)
+* v9.52.0 (2015-10-28)
```
adamkrellenstein: v9.51.5

counterpartylib/lib/blocks.py
robby-dermody: I'm doing it to be able to display in the `Block:` printout if the hash has changed. Eliminating that feature (which I think is pretty helpful) would remove the need to return this value.

adamkrellenstein: Do we not want to do this for all hashes?

ChangeLog.md
robby-dermody: Initially that's what it was, but this functionality requires a reparse to populate `messages_hash`, so I incremented the minor version.

counterpartylib/lib/blocks.py
robby-dermody: Not necessary for the other hashes, as we will throw an exception if they don't match up. With `messages_hash`, _currently_ we are simply displaying that it's a mismatch, for the reasons I outlined in the PR description.

adamkrellenstein: Or are we planning on replacing all of the other hashes with just the messages hash? (I think that'd be nice?)

In any case, there should be a note in the code for this.

robby-dermody: We always could throw an exception with `messages_hash`, but I'm not sure that's the best option, long term....

robby-dermody: I thought about replacing `ledger_hash` with it. It's an option, but I think `ledger_hash` has value, as it's more strict (we throw exceptions), we do checkpoints on it. `txlist_hash` I'd probably leave as is, because IIRC, it's more of a hash on the input data (transaction list coming in off of the parsed blockchain).

_If_ we enforce `message_hash` with checkpoints and throwing exceptions, like we do with `ledger_hash`, then it might make sense to replace `ledger_hash` with it... BUT, keeping the two could make sense as well, as if `messages_hash` is off and `ledger_hash` is not, you can immediately tell that the database data is different, in a way that does not impact balances. With just `messages_hash`, you don't get that same specificity.

ChangeLog.md
adamkrellenstein: ah, okay, makes sense!

counterpartylib/lib/config.py
adamkrellenstein: nvm

adamkrellenstein:
ACK

Issue #811, opened 2015-11-18
Counterparty tries to spend an output already spent in unconfirmed tx


arhuaco:
Hello there.

I have a few unit tests with Counterparty in regtest. Today one of them failed and I could "fix" it by mining a block before doing a TX. I checked the transactions and I think counterparty-lib is trying to spend an output that has already been spent in the mempool (not confirmed yet).

This TX is in the mempool:

    {
       "lock_time":0,
       "size":264,
       "inputs":[
          {
             "prev_out":{
                "index":1,
                "hash":"b4a0e795ee17c56fb4c3886dff87f9463a1b927c1147a3b2a7ec52c7f08b89cf"
             },
    "script":"47304402206b8ebfee6dba49c913f58dec7bbf52b17816d2d95bc7c7978c72f41c2fb8024602206019913c3bd273d9a2c0daa6de3a9d1810ee960ed4bf216d47b822eeb341fe360121039b67d69621634b64a7d7f4be5f0f95daae6d1984a56bfc4857e5a169bffc93de"
          }
       ],
       "version":1,
       "vin_sz":1,
       "hash":"342f973147a005d4d2fc8113824c1a7130b5637193375d93f03a79ad1a26130c",
       "vout_sz":3,
       "out":[
          {
             "script_string":"OP_DUP OP_HASH160 ef2034e33c1f9c5b7a4bc919f21f583398fe2492 OP_EQUALVERIFY OP_CHECKSIG",
             "address":"1NoP5bqskewdT7XoPs8ckJJtjU6HbSxivf",
             "value":5430,
             "script":"76a914ef2034e33c1f9c5b7a4bc919f21f583398fe249288ac"
          },
          {
             "script_string":"OP_RETURN b7cc839c5701ed91e007dccd9b53d45bc5e24e245782d0b18c61beb0",
             "value":0,
             "script":"6a1cb7cc839c5701ed91e007dccd9b53d45bc5e24e245782d0b18c61beb0"
          },
          {
             "script_string":"OP_DUP OP_HASH160 e2752565f2366653646da2ecd0cb278790d0e751 OP_EQUALVERIFY OP_CHECKSIG",
         "address":"1MeQ3on1EusqjPdNc7cqrsGHe4ymNX6h2S",
         "value":9764567,
         "script":"76a914e2752565f2366653646da2ecd0cb278790d0e75188ac"
          }
       ]
    }

[...]

PR #812, opened 2015-11-18
WIP: backend should keep cache of full mempool.

rubensayshi:
!!! WORK IN PROGRESS !! INCOMPLETE !!!
===================================
fixes https://github.com/CounterpartyXCP/counterparty-lib/issues/811 (and indirectly also https://github.com/CounterpartyXCP/counterwallet/issues/626)

#### Easiest way to reproduce
Send BTC twice from the same address before a block is found.

--------------

when doing `backend.get_unspent_txouts(unconfirmed=true)` the backend uses it's in-memory `unconfirmed_transactions_cache` instead of the `searchrawtransactions` from bitcoind, because bitcoind `searchrawtransactions` only works for confirmed transactions.

A while back (I think here https://github.com/CounterpartyXCP/counterparty-lib/commit/db052758b07eda6bcbf52ec9fce19d2be5030a99) some optimizations made it so that `backend.refresh_unconfirmed_transactions_cache` is only called for XCP related transactions, as a result any unconfirmed plain old BTC transactions are ignored and can't be spend until confirmed.

This PR changes the behaviory again to call `backend.refresh_unconfirmed_transactions_cache` with all the mempool TXs and optimizes `backend.refresh_unconfirmed_transactions_cache` to avoid calling `extract_addresses` more than once for a transaction.


rubensayshi:
The biggest danger here is that we'll be storing more data in memory, instead of only keeping the raw transactions for unconfirmed counterparty-related transactions we're storing all raw transactions in memory.

I think a good solution could be to only store the tx_hash in the `unconfirmed_transactions_cache` and the extracted addresses and do `getrawtransaction` calls once the actual raw transaction is needed (when it's actually being spend from)?

counterpartylib/lib/blocks.py
```diff
@@ -1241,18 +1243,23 @@ def follow(db):
...
+            refresh_start_time = time.time()
+            backend.refresh_unconfirmed_transactions_cache(raw_mempool)
+            refresh_time = time.time() - refresh_start_time
 
             elapsed_time = time.time() - start_time
             sleep_time = config.BACKEND_POLL_INTERVAL - elapsed_time if elapsed_time <= config.BACKEND_POLL_INTERVAL else 0
 
-            logger.debug('Refresh mempool: %s CP txs seen, out of %s total entries (took %ss, next refresh in %ss)' % (
-                len(mempool), len(raw_mempool), "{:.2f}".format(elapsed_time, 3), "{:.2f}".format(sleep_time, 3)))
+            logger.info('Refresh mempool: %s XCP txs seen, out of %s total entries (took %ss (%ss was backend refresh), next refresh in %ss)' % (
```
rubensayshi: change to `logger.info` is just because if I `--verbose` there's too much spam

counterpartylib/lib/backend/addrindex.py
```diff
@@ -116,22 +117,69 @@ def unconfirmed_transactions(address):
...
+    new_tx_hash_list = mempool_txhash_list.difference(known_tx_hash_list)  # mempool_txhash_list - known_tx_hash_list
+    old_tx_hash_list = known_tx_hash_list.difference(mempool_txhash_list)  # known_tx_hash_list - mempool_txhash_list
+
+    intersect_time = time.time() - intersect_start_time
+
+    cleanup_start_time = time.time()
+
+    # cleanup the dropped txs
+    for tx_hash in old_tx_hash_list:
+        for address in reverse_unconfirmed_transactions_cache[tx_hash]:
```
rubensayshi: using `reverse_unconfirmed_transactions_cache` for easy lookup

counterpartylib/lib/backend/addrindex.py
```diff
@@ -116,22 +117,69 @@ def unconfirmed_transactions(address):
...
+
+    if unconfirmed_transactions_cache is None:
+        unconfirmed_transactions_cache = {}
+    if reverse_unconfirmed_transactions_cache is None:
+        reverse_unconfirmed_transactions_cache = {}
+
+    intersect_start_time = time.time()
+
+    # create diffs of new txs and txs that have been dropped
+    known_tx_hash_list = set(reverse_unconfirmed_transactions_cache.keys())
```
rubensayshi: `set`s are amazing performance wise for .diff or .intersect

rubensayshi:
Running testnet the performance seems to remain the same as the old code, each mempool refresh takes between 0.1s and 0.2s, which is what I'm also getting with old code.

gonna setup 2 VMs on EC2 to test code side-by-side ...

adamkrellenstein:
Very nice! utACK

@ouziel-slama, @robby-dermody?

robby-dermody:
utACK, can test tomorrow

Great work, and thank you

counterpartylib/lib/backend/addrindex.py
```diff
@@ -113,25 +114,77 @@ def extract_addresses(txhash_list):
     return tx_hashes_addresses, tx_hashes_tx
 
 def unconfirmed_transactions(address):
-    logger.debug("unconfirmed_transactions called: %s" % address)
+    logger.info("unconfirmed_transactions called: %s" % address)
```
robby-dermody: Change back to debug before commit

counterpartylib/lib/backend/addrindex.py
```diff
@@ -113,25 +114,77 @@ def extract_addresses(txhash_list):
...
     if unconfirmed_transactions_cache is None:
         raise Exception("Unconfirmed transactions cache is not initialized")
-    return unconfirmed_transactions_cache.get(address, [])
+
+    tx_hashes = unconfirmed_transactions_cache.get(address, set())
+
+    logger.info("unconfirmed_transcations found: %s" % ",".join(list(tx_hashes)))
```
robby-dermody: Same here (use debug())

counterpartylib/lib/backend/addrindex.py
```diff
@@ -113,25 +114,77 @@ def extract_addresses(txhash_list):
...
-    unconfirmed_transactions_cache = unconfirmed_txes
-    logger.debug('Unconfirmed transactions cache refreshed ({} entries, from {} supported mempool txes)'.format(
-        len(unconfirmed_transactions_cache), len(mempool_txhash_list)))
+            unconfirmed_transactions_cache.setdefault(address, set())
+            unconfirmed_transactions_cache[address].add(tx_hash)
+            reverse_unconfirmed_transactions_cache[tx_hash].add(address)
+
+    cache_time = time.time() - cache_start_time
+
+    logger.info('Unconfirmed transactions cache refreshed (from {} mempool txs, contained {} entries, {} new entries required parsing, {} were deleted)'.format(
```
robby-dermody: And same with these (debug)

robby-dermody:
Can you debase off of develop? I'll test tomorrow and if it looks good we can merge into develop, in preparation for the next release

rubensayshi:
bitcoin core v0.12.0 will include mempool limiting, so we'll be able to piggyback on that and won't have to build anything to limit our own mempool.

see https://github.com/bitcoin/bitcoin/pull/6722

rubensayshi:
okay reopening against `develop`

PR #814, opened 2015-11-19
backend should keep cache of full mempool.

rubensayshi:
fixes https://github.com/CounterpartyXCP/counterparty-lib/issues/811 (and indirectly also https://github.com/CounterpartyXCP/counterwallet/issues/626)

#### Easiest way to reproduce
Send BTC twice from the same address before a block is found.

--------------

when doing `backend.get_unspent_txouts(unconfirmed=true)` the backend uses it's in-memory `unconfirmed_transactions_cache` instead of the `searchrawtransactions` from bitcoind, because bitcoind `searchrawtransactions` only works for confirmed transactions.

A while back (I think here https://github.com/CounterpartyXCP/counterparty-lib/commit/db052758b07eda6bcbf52ec9fce19d2be5030a99) some optimizations made it so that `backend.refresh_unconfirmed_transactions_cache` is only called for XCP related transactions, as a result any unconfirmed plain old BTC transactions are ignored and can't be spend until confirmed.

This PR changes the behaviory again to call `backend.refresh_unconfirmed_transactions_cache` with all the mempool TXs and optimizes `backend.refresh_unconfirmed_transactions_cache` to avoid calling `extract_addresses` more than once for a transaction.

----------

The biggest danger here is that we'll be storing more data in memory, instead of only keeping the raw transactions for unconfirmed counterparty-related transactions we're storing all raw transactions in memory.

I think this is a good solution, to only store the tx_hash in the unconfirmed_transactions_cache and the extracted addresses and do `getrawtransaction` calls once the actual raw transaction is needed (when it's actually being spend from).

---------

bitcoin core v0.12.0 will include mempool limiting, so we'll be able to piggyback on that and won't have to build anything to limit our own mempool.

see bitcoin/bitcoin#6722

### DIFF
--- a/counterpartylib/lib/backend/addrindex.py
+++ b/counterpartylib/lib/backend/addrindex.py
@@ -16,6 +16,7 @@ from counterpartylib.lib import config, script, util
 
 raw_transactions_cache = util.DictCache(size=config.BACKEND_RAW_TRANSACTIONS_CACHE_SIZE) #used in getrawtransaction_batch()
 unconfirmed_transactions_cache = None
+reverse_unconfirmed_transactions_cache = None
 
 class BackendRPCError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


rubensayshi:
rebase version of https://github.com/CounterpartyXCP/counterparty-lib/pull/812 against `develop`

rubensayshi:
it's taking a while to get 2 new bitcoind's up with `-reindex -addrindex=1`, but I want to run the old and new code side-by-side to see how it affects performance and memory usage.

robby-dermody:
@rubensayshi let me know once you have those results, and I merge this into `develop` and test some more.

adamkrellenstein:
ACK

robby-dermody:
Merging now so I can perform some testing along with the rest of the changes in `develop`

PR #816, opened 2015-11-23
Bug in `get_tx_info1` handling `OP_CHECKSIG` for data encoding!?

rubensayshi:
so I wanted to take the test vector from `get_tx_info1` and run it through `get_tx_info2` (and I was adding a `get_tx_info3`) because considering they're both internal to `get_tx_info` their return values should be the same ...
but they weren't!

`get_tx_info1` parsing a transaction with the data encoded in `OP_CHECKSIG` output keeps the `PREFIX` in the `data` return value. 
so the `data` returned was: `b'\x00\x00\x00(\x00\x00R\xbb3d\x00TESTXXXX\x00\x00\x00\x02\xfa\xf0\x80\x00\x00\x00\x00TESTXXXX\x02\xfa\xf0\x80\x00\x00\x00\x00\x00\x00\x00TESTXXXX\x00\x00\x00;\x10\x00\x00\x00\n\x9b\xb3Q\x92(6\xc8\x86\x81i\x87\xe1\x0b\x03\xb8_8v\x8b'`

it also contains the last output, the change output!

I'm not sure how the `OP_CHECKSIG` encoding is supposed to work, but I think it's bugged ...
someone who actually knows how the `OP_CHECKSIG` data storage works could probably say what the correct working should be!?

is it supposed to be

vouts (what the test vector is):
 - destination
 - PREFIX . data[part1]
 - PREFIX . data[part2]
 - PREFIX . data[part3]
 - change

or vouts (what the code does right now):
 - destination
 - PREFIX . data[part1]
 - data[part2]
 - data[part3] . PREFIX
 - change

or a mix?:
 - destination
 - PREFIX . data[part1]
 - (optional)PREFIX data[part2]
 - data[part3] . PREFIX
 - change


robby-dermody:
paging @adamkrellenstein or @ouziel-slama

adamkrellenstein:
`get_tx_info1()` can be different from `get_tx_info2()`, because the former was only use until a certain block height and thereafter the latter was used. In any case, `get_tx_info1()` isn't used for any future blocks, so there's no reason to change it, unless I'm mistaken.

rubensayshi:
`get_tx_info1` is used `< 333500` and `OP_CHECKSIG` is used `>= 293000` so only TXs between those heights are affected.

but I'm worried that any TXs between that height are broken ... because the data returned from it is probably not readable ...

I'll use the blocktrail database to search for a transaction between those heights that might be affected and debug a bit ;)

adamkrellenstein:
If they are broken, there's nothing that can be done about them. They're in the past, and you can't change the past. ;)

rubensayshi:
can at least put down a comment that they're broken and supposed to be ;-)

robby-dermody:
@rubensayshi: +1 to minimally adding a comment for this. can you submit a PR for that, referencing this PR, and we can close this one out?

I wouldn't say they're "broken" unless you have firm evidence that's the case. Otherwise, just comment that the get_tx_info1() return value format is not consistent with get_tx_info2(), or whatever the exact situation is

rubensayshi:
I'm pretty sure it's broken, the `data` returned is `PREFIX + "data_chunk1" + PREFIX + "data_chunk2" + PREFIX + "data_chunk3"`

but because there's nothing hitting it afaik, I'm having a hard time finding transactions that use p2pkh encoding to prove it xD

PR #817, opened 2015-11-25
enable wildcard matching for `--function`, eg `--function=blocks.*`

rubensayshi:
makes life a little easier when you're developing and just want to run a few specific tests, but more then just 1 or 2 xD


rubensayshi:
hmm, maybe not that neccesary after all ... xD

robby-dermody:
@rubensayshi  are you finding this useful, or shall I close this PR?

rubensayshi:
no, in practice while working it wasn't very useful

PR #818, opened 2015-11-26
P2SH address support [WIP]

rubensayshi:


rubensayshi:
@adamkrellenstein @robby-dermody  @ouziel-slama 
having an issue here that requires your attention because I lack knowledge

when creating a transaction (`transaction.construct`)
when source=P2SH 
and encoding=multisig (or auto with data >= 40 bytes) 

there's no pubkey to use for the dust to be send to.

is this relevant? 
are there transactions with data >= 40 bytes? 
if/when we moved to 80 bytes for op_return, will the multisig encoding be required at all?
should I fallback to pubkeyhash encoding when source=P2SH and encoding=multsig? or raise an exception?

adamkrellenstein:
> are there transactions with data >= 40 bytes? 

Yes

> if/when we moved to 80 bytes for op_return, will the multisig encoding be required at all?

Yes, even 80 bytes is not enough for all use-cases

> should I fallback to pubkeyhash encoding when source=P2SH and encoding=multsig? or raise an exception?

I think if an encoding is explicitly specified, and you can't use it, you should raise an exception.

rubensayshi:
hmm okay.

I asume the `pubkeyhash` is [a lot] less space efficient than `multisig` encoding right? so I don't see a reason why I'd fall back to that, 
considering it's burning the dust anyway might as well just use `multisig` encoding with 3 data pubkeys and burn the dust that way?

or alternatively add an extra (optional) argument to `transaction.construct`, `dust_return_pubkey`, for when `source=p2sh`.
in that case the P2SH multisig wallet just needs to (eventually) build something to sweep all the dust it has accumulated back into a P2SH address to start using it again.

rubensayshi:
so I did the following now;

 - when `source==P2SH` and `encoding==auto` with > 40 bytes of data, `encoding=multisig` (nothing new here)
 - when `source==P2SH` and `encoding=multisig` (could be from auto + >40 bytes) the `dust_return_pubkey` argument is required to be set
    - this can either be a pubkey (or address) in which case that will be used as 3rd pubkey for the data outputs
    - or it can be explicitly set to FALSE so indicate the dust can be burned, in which case the 3rd pubkey is set to `config.UNSPENDABLE`

I was initially thinking of using the 3rd pubkey for data too when `dust_return_pubkey` is explicitly set to FALSE, considering it's being burned anyway, but that would require a lot more changes to the TX parsing code, though it might be worth it, considering:
 - it gives us 33 extra bytes
 - it's a bit weird to add `config.UNSPENDABLE` to each data output

adamkrellenstein:
Right now, all of the multi-sig outputs created by Counterparty by default are spendable precisely because the third public key is the sender's public key. It's important to preserver this design so as not to 'pollute' the UTXO set with unspendable outputs. (The optional `pubkeyhash` encoding is even worse.)

If we can't get the sender's public key from the source address when it's P2SH, then I think we have two options: 1) get it from somewhere else, or 2) just use `opreturn` encoding (and limit the data size for P2SH transactions to 80 bytes for now).

rubensayshi:
@adamkrellenstein I discussed a bit with @robby-dermody 

from a wallet developer point of view (ie. when I look at this from the point of view if I would want to use P2SH counterparty in a blocktrail wallet) 
it seems like a lot of hassle to specify a `return_dust_pubkey` (ie; probably at least a day of coding and more odd code)   
for just a bit of dust ... 

which is only waisted when the data doesn't fit in an opreturn (which will soon be 80 bytes) which from me - as a wallet developer - point of view wouldn't be worth it until users would start complaining about their dust being waisted.

so even if you'd make it a required argument, I - as a wallet developer - would probably just feed counterparty some bogus pubkey `randomBytes(33)`.

@robby-dermody seemed to agree on this conclusion about the mindset of a developer who would use P2SH for counterparty.

from that angle we might as well make it easy to not provide a pubkey and burn the dust, though we should still encourage (through documentation) specifying a pubkey.

our idea was to encourage (through documentation) P2SH applications to specified the `return_dust_pubkey`
but still allow a developer to explicitly set it to FALSE for a zero filled 3rd pubkey.

as you read in my previous comment, I suggested even use the 3rd pubkey for data too, to be more efficient, but maybe it's better not to, to encourage the developer providing a `return_dust_pubkey` a bit more.

I think you guys should make the final call on this, but for now I think I can continue coding stuff and adjust based on what you guys decide.
Also, I'll let you guys know once this stuff is ready for code review, atm this PR is just open for the discussion. 

PS. maybe you can enlighten me which messages tend to be >40 or even >80 bytes?

deweller:
How about we create a default address that is owned by the Counterparty foundation for multisig transactions coming from a p2sh source?

So when 1) a transaction does not fit into 80 bytes and 2) the user does not specify a `return_dust_pubkey`, we can use the public key of the Counterparty foundation as redeemable the public key for the multisig dust.

The foundation can redeem the dust once a month and use it for funding, or purchase XCP on the DEX and burn the XCP.

I think the amount of funds we are talking about here will be very small.

rubensayshi:
@deweller I considered that (I actually wanted to use the XCP burn address initially, but somehow I failed to comprehend that it was a burn address and that there's no pubkey for it xD)  
the only thing is, you guys added the encryption of the data at some point to obfuscate data, reusing a certain address for this purpose undos that.  
I assume the purpose was to make miners who would want to blacklist counterparty transactions to expend CPU to figure out what the data is, so that's no great either.  
I suppose zerofilling the 3rd pubkey could sort of give it away too, so instead of zerofilling putting random data in there might be better?

deweller:
The obfuscation is based on the transaction ID (hash) of the first spent UTXO in the transaction, so reusing the same public key as a multisig dust output would not affect the obfuscation.

If you generate a random (or use a zero-filled) public key for the the first multisig dust output, then those UTXOs will linger in the blockchain and never be spent.  Specifying a default Counterparty foundation controlled public key is preferable to growing the UTXO pool.

It can also be a config option if others want to specify their own default redeemable multisig dust public key.

rubensayshi:
if we reuse 1 address they can easily blacklist that 1 address, the obfuscation will no longer have any effect.
could do an counterparty foundation xpub I guess and create new derivation or smt...

robby-dermody:
I wouldn't worry about the threat of blacklisting too much. It's a very political topic that I think is behind us now, and we can always zero out (or randomize) the dust pubkey if it ever becomes an issue (of course leading to unspendable UTXOs, but only as a worse case side effect)...

rubensayshi:
there's 200 transactions with data > 80 bytes, I say we just burn it for now?

rubensayshi:
depends on https://github.com/CounterpartyXCP/counterparty-lib/pull/830 and https://github.com/CounterpartyXCP/counterparty-lib/pull/827 and https://github.com/CounterpartyXCP/counterparty-lib/pull/825 and https://github.com/CounterpartyXCP/counterparty-lib/pull/829 

because merging this first before those 2 just gives tons of merge conflicts (and I'm pretty sure they're gonna get merged).

rubensayshi:
now that #829 is merged this diff is readable ;-)

gotta go through it once more entirely myself, but afaik it's done

robby-dermody:
@rubensayshi is this ready on your end? (i.e. you went through it that one more time?)

adamkrellenstein:
We, as the Counterparty Foundation, have to be "good blockchain citizens", so we should make it at least a little bit difficult to pollute the UTXO pool with unspendable outputs. We also can't "enrich ourselves" with the dust. If wallet devs, however, want to use either random bytes or a donation address, that's entirely up to them.

ouziel-slama:
> We, as the Counterparty Foundation, have to be "good blockchain citizens", so we should make it at least a little bit difficult to pollute the UTXO pool with unspendable outputs. We also can't "enrich ourselves" with the dust. If wallet devs, however, want to use either random bytes or a donation address, that's entirely up to them.

+1
I can't see any good default value for `dust_return_pubkey`.
IMHO that is not a decision for the wallet developer, but it's up to the wallet user to decide if they want to use a donation address. As a wallet user I would like that, by default, the wallet generates an address from my master key, or let me pick from a select box this address. That said, it would be really nice if this select box contains an address clearly labelled "Donation For The Counterparty Foundation".

rubensayshi:
it's up to the wallet developer to enable such an option, the question is what would be the default

adamkrellenstein:
Right. My thought was that the default should be to not create an unspendable output.

rubensayshi:
Updated to no longer use a default P2SH_DUST_RETURN_PUBKEY, this means that to use P2SH either one of two things need to happen:

 - the node operator needs to set `--p2sh-dust-return-pubkey` and the api call needs to set `dust_return_pubkey=False`
 - or the api call needs to set `dust_return_pubkey=<pubkey>`

adamkrellenstein:
I'm okay with that.

robby-dermody:
@adamkrellenstein ok what are next steps in your mind before merging into `develop` (assuming @rubensayshi has completed his final review of this PR) -- have @ouziel-slama review it. Who else?

I've pinged @btcdrak about a review of the code as well

adamkrellenstein:
No one else. We should test it extensively on `testnet` of course, though. It'll have to stay in `develop` for a little while at least.

counterpartylib/lib/transaction.py
```diff
@@ -362,24 +395,21 @@ def construct (db, tx_info, encoding='auto',
...
 
     '''Inputs'''
 
-    # Source.
-        # If public key is necessary for construction of (unsigned)
-        # transaction, use the public key provided, or find it from the
-        # blockchain.
-    if source:
-        script.validate(source)
-    if encoding == 'multisig':
```
robby-dermody: not super familiar with this part of the code so you'll have to excuse my ignorance:

is there any reason you moved this up under `if data:`? How would this change impact construction of something like a testnet burn tx (which has no `data`) using multisig encoding?

robby-dermody:
also, can we have it so that if `dust_return_pubkey` is passed through the API, it will override any setting of `p2sh-dust-return-pubkey` in the node config? That way, the API caller/dev does not need to worry about whether the node operator has `p2sh-dust-return-pubkey` set or not, as long as they are setting `dust_return_pubkey` ...

robby-dermody:
past that, utACK. Can merge once these two things I raised are looked into, and the PR is rebased

robby-dermody:
@rubensayshi can you please fix the failing tests, and then I can merge?

PR #821, opened 2015-12-05
add back time into logformat

rubensayshi:
adding back time to logformat and 'custom' datetime format (to get rid of the milliseconds, improves readability a bit.


robby-dermody:
ACK
@adamkrellenstein ?

PR #822, opened 2015-12-05
Mempool optimizations

rubensayshi:
I've been trying to optimize a few things to get things running smoother on lower hardware, I'll put some comments inline about the changes.

### DIFF
--- a/counterpartylib/lib/backend/addrindex.py
+++ b/counterpartylib/lib/backend/addrindex.py
@@ -152,6 +152,8 @@ def refresh_unconfirmed_transactions_cache(mempool_txhash_list):
 
     intersect_time = time.time() - intersect_start_time
 
+    logger.debug("refresh_unconfirmed_transactions_cache: %d txs, %d new, %d dropped" % (len(mempool_txhash_list), len(new_tx_hash_list), len(old_tx_hash_list)))
+
     cleanup_start_time = time.time()
 
     # cleanup the dropped txs
--- a/counterpartylib/lib/backend/addrindex.py
+++ b/counterpartylib/lib/backend/addrindex.py
@@ -90,26 +90,34 @@ def rpc_batch(request_list):
     return list(responses)
 
 def extract_addresses(txhash_list):
+    logger.debug('extract_addresses, txs: %d' % (len(txhash_list), ))
     tx_hashes_tx = getrawtransaction_batch(txhash_list, verbose=True)
     tx_hashes_addresses = {}
-    tx_inputs_hashes = set() #use set to avoid duplicates
-    
+    tx_inputs_hashes = set()  # use set to avoid duplicates
+
     for tx_hash, tx in tx_hashes_tx.items():
         tx_hashes_addresses[tx_hash] = set()
         for vout in tx['vout']:
             if 'addresses' in vout['scriptPubKey']:
                 tx_hashes_addresses[tx_hash].update(tuple(vout['scriptPubKey']['addresses']))
+
         tx_inputs_hashes.update([vin['txid'] for vin in tx['vin']])
 
-    raw_transactions = getrawtransaction_batch(list(tx_inputs_hashes), verbose=True)
-    for tx_hash, tx in tx_hashes_tx.items():
-        for vin in tx['vin']:
-            vin_tx = raw_transactions[vin['txid']]
-            if vin_tx is None: #bogus transaction
-                continue
-            vout = vin_tx['vout'][vin['vout']]
-            if 'addresses' in vout['scriptPubKey']:
-                tx_hashes_addresses[tx_hash].update(tuple(vout['scriptPubKey']['addresses']))
+    logger.debug('extract_addresses, input TXs: %d' % (len(tx_inputs_hashes), ))
+
+    # chunk txs to avoid huge memory spikes
+    for tx_inputs_hashes_chunk in util.chunkify(list(tx_inputs_hashes), config.BACKEND_RAW_TRANSACTIONS_BATCH_SIZE):
+        raw_transactions = getrawtransaction_batch(tx_inputs_hashes_chunk, verbose=True)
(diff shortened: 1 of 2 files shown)


counterpartylib/lib/backend/addrindex.py
```diff
@@ -90,26 +90,34 @@ def make_call(chunk):
...
-            vin_tx = raw_transactions[vin['txid']]
-            if vin_tx is None: #bogus transaction
-                continue
-            vout = vin_tx['vout'][vin['vout']]
-            if 'addresses' in vout['scriptPubKey']:
-                tx_hashes_addresses[tx_hash].update(tuple(vout['scriptPubKey']['addresses']))
+    logger.debug('extract_addresses, input TXs: %d' % (len(tx_inputs_hashes), ))
+
+    # chunk txs to avoid huge memory spikes
+    for tx_inputs_hashes_chunk in util.chunkify(list(tx_inputs_hashes), config.BACKEND_RAW_TRANSACTIONS_CACHE_SIZE):
```
rubensayshi: `extract_addresses` is called for the full mempool (since my previous PR), because this 2nd batch is a lot bigger (at least x10 or more) doing this without chunking makes for a huge memory spike!

counterpartylib/lib/blocks.py
```diff
@@ -1200,53 +1206,62 @@ def follow(db):
...
+            # fetch raw for all transactions that need to be parsed
+            # Sometimes the transactions can’t be found: `{'code': -5, 'message': 'No information available about transaction'}`
+            #  - is txindex enabled in Bitcoind?
+            #  - or was there a block found while batch feting the raw txs
+            #  - or was there a double spend for w/e reason accepted into the mempool (replace-by-fee?)
+            try:
+                raw_transactions = backend.getrawtransaction_batch(raw_mempool)
+            except backend.addrindex.BackendRPCError as e:
+                logger.warning('Failed to fetch raw for mempool TXs, restarting loop; %s', (e, ))
+                continue  # restart the follow loop
```
rubensayshi: instead of raise `MempoolError` making it explicit that we are 'okay' with the -5 errors and just retry.

counterpartylib/lib/blocks.py
```diff
@@ -1200,53 +1206,62 @@ def follow(db):
...
+                                            block_index,
+                                            block_hash,
+                                            block_time) VALUES(?,?,?)''',
+                                       (config.MEMPOOL_BLOCK_INDEX,
+                                        config.MEMPOOL_BLOCK_HASH,
+                                        curr_time)
+                                      )
+
+                        tx_hex = raw_transactions[tx_hash]
+                        mempool_tx_index = list_tx(db, None, block_index, curr_time, tx_hash, tx_index=mempool_tx_index, tx_hex=tx_hex)
```
rubensayshi: use prefetch `tx_hex` to avoid a lot of singular calls, helps a lot!

counterpartylib/lib/blocks.py
```diff
@@ -1254,10 +1269,20 @@ def follow(db):
...
+            # let the backend refresh it's mempool stored data
+            # Sometimes the transactions can’t be found: `{'code': -5, 'message': 'No information available about transaction'}`
+            #  - is txindex enabled in Bitcoind?
+            #  - or was there a block found while batch feting the raw txs
+            #  - or was there a double spend for w/e reason accepted into the mempool (replace-by-fee?)
+            try:
+                backend.refresh_unconfirmed_transactions_cache(raw_mempool)
+            except backend.addrindex.BackendRPCError as e:
+                logger.warning('Failed to fetch raw for mempool TXs, restarting loop; %s', (e, ))
+                continue  # restart the follow loop
```
rubensayshi: again, 'accepting' the -5's and handling them

counterpartylib/lib/blocks.py
```diff
@@ -1066,6 +1065,8 @@ def follow(db):
                 continue
             else:
                 raise e
+
+        logger.debug('follow %d <= %d' % (block_index, block_count))
```
robby-dermody: place under noisy logger? or is this just transient debug code?

counterpartylib/lib/blocks.py
```diff
@@ -1181,12 +1182,17 @@ def follow(db):
             curr_time = int(time.time())
             mempool_tx_index = tx_index
 
+            xcp_mempool = []
+            raw_mempool = backend.getrawmempool()
+
+            logger.debug('Mempool size: %d' % (len(raw_mempool), ))
```
robby-dermody: Same as above comment

counterpartylib/lib/blocks.py
robby-dermody: could we end up with a situation where there's a persistent bad tx hash (-5) in the mempool and we just keep retrying over and over?  (I've seen reports that had issues due to a -5 result hanging around for a few calls, but I haven't been able to reproduce it myself) -- what have you seen in your testing?

robby-dermody:
utACK

counterpartylib/lib/blocks.py
rubensayshi: sounds impossible to me ...

maybe during bitcoind processing a block, `getrawmempool` returns the mempool prior to the block but the processing of the block is already dropping some of them?

though I doubt that, as far as I can tell, during processing of a block rpc calls are blocked ( I get timeouts every once in a while because the bitcoind on my laptop is still syncing ).

rubensayshi:
cleaned up the 2 debug log statements

counterpartylib/lib/blocks.py
robby-dermody: ok...with our discussion about 0.11.1+ automatically discarding high-S tx from its mempool (due to not matching `isstandard()` this makes sense now.

robby-dermody:
merging in and will test further

Issue #810, opened 2015-11-18
Uncaught exception in testnet ('invalid pushdata due to truncation').


arhuaco:
Hello. When I try to run:

    counterparty-client send --source mogTjt1DU1trpKsXgoesEL1J2PkXmyStnE  --destination mognqybiU68RwJExYscAaBhf9gkpVc1UAV  --quantity 1 --asset SPTEST

I get the following exception:

    counterpartycli.util.RPCError: {'data': {'type': 'DecodeError', 'message': 'invalid pushdata due to truncation', 'args': ['invalid pushdata due to truncation']}, 'code': -32000, 'message': 'Server error'}

I debugged a bit and found out that the exception happens when counterpartylib/lib/script.py:get_asm is processing TX [720b1a0a17cf19718a8bdfc65b0f2d3558e3e13e210c3c8da14496f9c05d00d9](http://testnet.blockscan.com/txInfo/3986).

What could be the issue?

Running counterparty-lib v9.52.0 and counterparty-cli v1.1.1 without modifications.

arhuaco:
Hi there. I still cannot log in into https://testnet.counterwallet.io/ because of this issue :-/

robby-dermody:
that transaction shows up on blockscan as a send (http://testnet.blockscan.com/txInfo/3986).

But it has an extra odd output: https://test-insight.bitpay.com/tx/720b1a0a17cf19718a8bdfc65b0f2d3558e3e13e210c3c8da14496f9c05d00d9 shows a 4th output with the following scriptPubkey: ```OP_RETURN OP_3 OP_VERIF OP_DEPTH OP_VERNOTIF OP_VERIFY OP_DUP OP_VERIF 7361797320686921```

Another send transaction on testnet (that processes fine) is https://test-insight.bitpay.com/tx/3eb26ebae2424d865dd726803c5d35e5a3ab382dbb37882e020899b3e816a78c (http://testnet.blockscan.com/txInfo/4384). That doesn't have that same output (only has 3 outputs)

Looking at it further, it looks like someone threw together a custom message with an extra OP_RETURN output: https://www.blocktrail.com/tBTC/tx/720b1a0a17cf19718a8bdfc65b0f2d3558e3e13e210c3c8da14496f9c05d00d9#tx_messages

The OP_RETURN decodes down to "Setfive says hi!"

[...]

PR #825, opened 2015-12-08
fixing `backend.get_unspent_outputs` raising exception when a transaction contains an output with garbage data

rubensayshi:
https://github.com/CounterpartyXCP/counterparty-lib/issues/810 occurs in `create_send` -> `backend.get_unspent_outputs` -> `backend.is_scriptpubkey_spendable` -> `script.scriptpubkey_to_address` path,  
blocks.get_tx_info is unaffected, so this isn't affecting consensus, only making certain transactions unspendable (and blocking any spending from a wallet).

`script.scriptpubkey_to_address` was raising an exception when it couldn't get the ASM from the `scriptPubKey`, this breaks `backend.get_unspent_outputs` and as a result breaks `create_send`.

I've only added a `try/except` for the `PushDataDecodeError` atm, but I think catching (the more generic) `DecodeError` would be even better,
I don't think this behavory of `script.scriptpubkey_to_address` raising an exception is desired, because we just want `None` returned for any scriptpubkeys that aren't spendable.

### DIFF
--- a/counterpartylib/lib/backend/__init__.py
+++ b/counterpartylib/lib/backend/__init__.py
@@ -14,6 +14,7 @@ from bitcoin.core import CBlock
 from counterpartylib.lib import util
 from counterpartylib.lib import script
 from counterpartylib.lib import config
+from counterpartylib.lib import exceptions
 
 from counterpartylib.lib.backend import addrindex, btcd
 
@@ -123,7 +124,12 @@ def get_btc_supply(normalize=False):
 
 def is_scriptpubkey_spendable(scriptpubkey_hex, source, multisig_inputs=False):
     c_scriptpubkey = bitcoinlib.core.CScript(bitcoinlib.core.x(scriptpubkey_hex))
-    vout_address = script.scriptpubkey_to_address(c_scriptpubkey)
+
+    try:
+        vout_address = script.scriptpubkey_to_address(c_scriptpubkey)
+    except exceptions.DecodeError:
+        return False
+
     if not vout_address:
         return False
 
--- a/counterpartylib/lib/blocks.py
+++ b/counterpartylib/lib/blocks.py
@@ -620,9 +620,9 @@ def get_tx_info2(tx_hex, block_parser=None):
 
         # Ignore transactions with invalid script.
         try:
-          asm = script.get_asm(vout.scriptPubKey)
+            asm = script.get_asm(vout.scriptPubKey)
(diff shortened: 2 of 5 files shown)


robby-dermody:
UtACK, @adamkrellenstein ?

arhuaco:
Hi there. Will this be part of the next release? Could I use this patch right now to work around the issue? Thanks.

robby-dermody:
Yes, if @adamkrellenstein is fine with this, we can merge it now.

adamkrellenstein:
> because we just want None returned for any scriptpubkeys that aren't spendable.
Where do we want that? Can we catch the error *there* (in the choosing of inputs) instead?

rubensayshi:
hmm idk I'd consider it `scriptpubkey_to_address` responsibility to either silently handle it or catch it and throw a new exception, it doesn't make sense that when you're calling `scriptpubkey_to_address` that you have to catch `DecodeError`.

it's used only for `is_scriptpubkey_spendable` so if you insist then I don't mind moving the trycatch

adamkrellenstein:
I'm okay with having it throw a different exception, but I don't think it should return `None`.

rubensayshi:
updated, exception is now caught in `is_scriptpubkey_spendable`

counterpartylib/lib/backend/__init__.py
```diff
@@ -123,7 +124,12 @@ def get_btc_supply(normalize=False):
 
 def is_scriptpubkey_spendable(scriptpubkey_hex, source, multisig_inputs=False):
     c_scriptpubkey = bitcoinlib.core.CScript(bitcoinlib.core.x(scriptpubkey_hex))
-    vout_address = script.scriptpubkey_to_address(c_scriptpubkey)
+
+    try:
+        vout_address = script.scriptpubkey_to_address(c_scriptpubkey)
+    except exceptions.DecodeError:
+        return False
```
adamkrellenstein: I'm saying that we shouldn't return `False` here. The behavior of such a change is, on the face of it, undefined.

rubensayshi:
we're calling ` is_scriptpubkey_spendable` with a valid `scriptpubkey_hex`, there's no reason why it should raise an Exception, it's result should be False imo

adamkrellenstein:
Ah, I'm okay with this.

@ouziel-slama ?

robby-dermody:
merging, as Adam is fine with this and we haven't heard from Ouziel

PR #827, opened 2015-12-12
padding byte used should be `x00` not the `version`.

rubensayshi:
I think this was excidently replaced because it looks like the `b'\x00'` is exidentlal hardcoding, but it's not!

this mistake/bug had no effect up until I've started implementing P2SH.

### DIFF
--- a/counterpartylib/lib/script.py
+++ b/counterpartylib/lib/script.py
@@ -106,7 +106,7 @@ def base58_check_decode(s, version):
             pad += 1
         else:
             break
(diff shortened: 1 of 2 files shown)


adamkrellenstein:
wow, weird error!

Why didn't we catch this before?

rubensayshi:
hmm, well only a small percentage of addresses have padding and I guess no1 ever noticed on testnet

rubensayshi:
still, this code looks so weird...

PR #829, opened 2015-12-17
placed all vectors `records` on multiple lines

rubensayshi:
while working on the p2sh addresses stuff it was becoming impossible to know what changes I was making to vectors, if each key is on it's own line it becomes a readable diff instead.

cuz of how much work it is to do this I used alphabetical sorting (plugin on pycharm) to sort them, eventhough some other sorting might make more sense, but this is a good start I think


robby-dermody:
@rubensayshi can you rebase, and then I can merge?

rubensayshi:
@robby-dermody done

robby-dermody:
The diff (https://github.com/CounterpartyXCP/counterparty-lib/commit/273fd2d148294501ff92509546f9364069a38b58.patch) is obviously difficult to visually parse. @adamkrellenstein are you good with merging this?

adamkrellenstein:
ACK

@ouziel-slama ?

robby-dermody:
Would like to merge this into `develop` by tomorrow (Sunday) and merge the 9.53.0 PR Sunday as well, if we don't hear from @ouziel-slama

rubensayshi:
that would be nice, then we can start reviewing the P2SH PR ^_^

PR #838, opened 2016-03-05
Partial Crash Prevention

adamkrellenstein:


ouziel-slama:
ACK

robby-dermody:
@adamkrellenstein Can we skip just the transaction itself instead of skipping all transactions in the block? That would be much preferable.

adamkrellenstein:
we can't because of this line: https://github.com/CounterpartyXCP/counterparty-lib/pull/838/files#diff-a7b98d56d05b9a595e317923cab93688L1125

robby-dermody:
closing in favor of #862

PR #844, opened 2016-03-14
use the data from the block instead of hitting RPC for raw TXs.

rubensayshi:
Sean brought up some questions about the raw TX cache and it made me think about why we're even querying the RPC for the raw TXs when parsing blocks.

Easy fix for a pretty big performance gain.

Improvement from 2~3s to < 0.1s for me


adamkrellenstein:
ACK

robby-dermody:
will merge this, but as we run it in `develop`, let's watch the consensus hashes to make sure they don't diverge from `master`. Don't see why they would, but just to be sure.

PR #847, opened 2016-03-15
add util.sizeof to debug size of a dict

rubensayshi:
small util function to debug size of dict.

result; getrawtransaction cache with 20k txs is ~130mb


robby-dermody:
is this being used anywhere in the code (the commit doesn't call it...does another PR depend on this one)?

rubensayshi:
no, but it's very useful for debugging memory usage

PR #848, opened 2016-03-16
add --verbosediff option for unit tests

rubensayshi:
imo should should be the default, but it's a --arg in case ppl don't like it ...

```
[...]

PR #849, opened 2016-03-16
updated error msg to make it more clear which ones are the expected outputs

rubensayshi:
depends on https://github.com/CounterpartyXCP/counterparty-lib/pull/848

making it a little bit cleared which of the 2 is the expected outputs, because even after a reasonable amount of hours coding on CP I still get confused xD


rubensayshi:
nvm, including it in the other PR

PR #850, opened 2016-03-16
XCP Dividends for PoS Voting

rubensayshi:
depends on: https://github.com/CounterpartyXCP/counterparty-lib/pull/848

code for CIP-0004: https://github.com/rubensayshi/cips/blob/cip-0004/cip-0004.md  
pending discussion on the CIP: https://counterpartytalk.org/t/cip-xcp-dividends-for-pos-voting/1895


rubensayshi:
closing in favor of using broadcasts

PR #851, opened 2016-03-17
placed broadcast != valid check on correct place in the `broadcast.parse` (PROTOCOL CHANGE)

rubensayshi:
this stuff smells, the `util.enabled('max_fee_fraction'):` check was only added recently, before that an invalid broadcast (invalid message or locked feed) would still trigger bet matches.  
after the `util.enabled('max_fee_fraction'):` it no longer does bet matches, but bet cancels are still possible ...

I think it would make sense a LOCK does properly lock the feed, or otherwise it should be made specific to the betting message (ie. BETLOCK). 

FYI; lock has never been used until I brought it up yday in the topic about the voting stuff and JP tested it out ...


counterpartylib/lib/messages/broadcast.py
```diff
@@ -186,16 +190,17 @@ def parse (db, tx, message):
         cursor.close()
         return
 
+    # stop processing if broadcast is invalid for any reason,
```
robby-dermody: this block doesn't even seem necessary (although it is better placed now than it was before)

maybe we should change the first `status != valid` block to just be:

`if (util.enabled('broadcast_invalid_check') or util.enabled('max_fee_fraction')) and status != 'valid':`  ??

robby-dermody:
utACK on this, once my input above is addressed. I'm in favor of keeping the semantics `lock` instead of `betlock` or something.

As this is a protocol change, we need to give some thought to the timing before merging. As this is non-urgent (IMO), for minimum disruption/upgrading required, I would time it to be the same as P2SH and kill two birds with one stone

counterpartylib/lib/messages/broadcast.py
rubensayshi: that's the old block that was already there (except it was in the `for`) so it needs to stay in place for parsing older stuff

rubensayshi: my consensus hashes are fine with only the `util.enabled('broadcast_invalid_check')` check and removing the `util.enabled('max_fee_fraction'))`

rubensayshi:
consensus hashes are the same no matter what, but I think it would be better to just remove the `max_fee_fraction` check once `broadcast_invalid_check` has been activated.

rubensayshi:
updated PR to be dependant on https://github.com/CounterpartyXCP/counterparty-lib/pull/853

so that I could add a test for this

PR #853, opened 2016-03-18
More quality of life testing improvements

rubensayshi:
all `insert_block` calls in the tests were not doing `blocks.parse_block`, I guess so that we could unittest `blocks.parse_block`.  
instead now do parse_block everywhere, except for the last block so we can still unittest `blocks.parse_block`, because it should be doing this and it was causing issues with adding more bets to the scenario.

also check `<= BLOCK_FIRST` for consensus hashes, because tests start 1 block before it (and it's a hell to change that to actually start on `BLOCK_FIRST`, so this is an easier change).

I'm trying to add a test to test the locked feed properly, but needed to do all this stuff to be able to add that test ...


rubensayshi:
running a `reparse` right now on mainnet to make sure consensus hashes are the same, so don't merge until I can confirm that ;-)

robby-dermody:
do you consider this ready for merge?

rubensayshi:
yes

PR #855, opened 2016-03-31
Problem with broadcast text exactly 52 chars!  (PROTOCOL CHANGE)

rubensayshi:
when the string is exactly 52 bytes it will encode it as `52p` which will be `<lengthbyte><string>` and the result is 53 bytes long ...  
so when parsing it will be parsed as `53s` and the result will contain the length byte!

also it never took into account that the text might contain utf-8 chars for length...

atm I changed it to always omit the length byte since it's unnecesary,  
but we could also go the other way and always use a length byte (using `bitcoin.core.VarIntSerializer`), which **I personally prefer** over no varint, so alternative pull request for that: https://github.com/CounterpartyXCP/counterparty-lib/pull/857

either way we're gonna BC break, but the other PR actually BC breaks for composing > 52 chars broadcast and this one BC breaks for composing <= 52 chars (which are way more common).  
also breaks consensus so that's why it contains a `protocol_change`.

also; anyone know why 52 was chosen when up to 255 bytes are possible for `p`?

depends on https://github.com/CounterpartyXCP/counterparty-lib/pull/856 because I needed a way to keep the old code tested!


unsystemizer:
I don't know why 52 bytes, but I just looked at the tests and the test for > 52 bytes tries to broadcast 59 bytes instead of 53.

rubensayshi:
closing in favor of #857

PR #856, opened 2016-03-31
Add a way to have vectors for code paths before protocol changes.

rubensayshi:
It's not the first time that I add a protocol change and regret that I can no longer test the old codepath, so here it is!

just add `'mock_protocol_changes': {"protocol_change": False}` to disable a particular protocol change (see: https://github.com/CounterpartyXCP/counterparty-lib/pull/855/commits/6ea2e8ce02c5ef2951042705378cb85858c29cb2#diff-fa3ae8bc7e155d00792f770690403cbbR402 for it in action).

we need this because when you reparse from 0 the older protocol changes shouldn't change with future code changes.

### DIFF
--- a/counterpartylib/test/conftest.py
+++ b/counterpartylib/test/conftest.py
@@ -27,6 +27,16 @@ from counterpartylib.lib import config, util, backend, transaction, database, ap
 import bitcoin as bitcoinlib
 import bitcoin.rpc as bitcoinlib_rpc
 
+# we swap out util.enabled with a custom one which has the option to mock the protocol changes
+MOCK_PROTOCOL_CHANGES = {}
+_enabled = util.enabled
+def enabled(change_name, block_index=None):
+    if change_name in MOCK_PROTOCOL_CHANGES:
+        return MOCK_PROTOCOL_CHANGES[change_name]
+
+    return _enabled(change_name, block_index)
+util.enabled = enabled
+
 def pytest_collection_modifyitems(session, config, items):
     """Run contracts_test.py last."""
     items[:] = list(reversed(items))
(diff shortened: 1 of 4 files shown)


robby-dermody:
I like this. Can you add some kind of safeguard/assertion to ensure that `MOCK_PROTOCOL_CHANGES` is not set/used during normal operations (i.e. not when running the test suite), as `util.enabled()` is used in both contexts.

rubensayshi:
I had the same idea but I can't think of anyway to do so?

I could set another global `ENABLE_MOCK_PROTOCOL_CHANGES` .. ?  
or use `inspect` to check that it's ran from pyunit?

robby-dermody:
Hmm I would have thought that we have a more generic global that is set when the test suite runs, but I can't find one. There's an approach like http://stackoverflow.com/a/25188424 but it's a bit of an antipattern. :) `util. ENABLE_MOCK_PROTOCOL_CHANGES` may be cleanest (hate adding additional globals) because it doesn't encourage use of the data (i.e. to see if pytest is running or not) outside of this one context. You could set it in that pytest hook as referenced in the stackoverflow link.

The inspect method is too magical and possibly fragile for me, I think.

rubensayshi:
maybe overwriting `util.enabled` from `conftest.py` is a decent solution?

robby-dermody:
magic is involved in that, right? (i.e. a runtime swapout of the function)? that might be the cleanest way though, as it keeps additional globals and magic out of the non-test code...

rubensayshi:
updated to overwrite the `util.enabled` from the `conftest.py`

robby-dermody:
please rebase and I will merge

PR #857, opened 2016-03-31
Problem with broadcast text exactly 52 chars!  (PROTOCOL CHANGE)

rubensayshi:
when the string is exactly 52 bytes it will encode it as `52p` which will be `<lengthbyte><string>` and the result is 53 bytes long ...  
so when parsing it will be parsed as `53s` and the result will contain the length byte!  
also it never took into account that the text might contain utf-8 chars for length...

I changed it to always use a length byte (using `bitcoin.core.VarIntSerializer`).

this is gonna be a BC break for composing <= 52 chars broadcast because they will require a varint after this change.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,7 @@
     * Only use first usable input for source. (protocol change: )
     * lock UTXOs used to construct a transaction for 3 seconds to avoid a user double spending against himself.
     * P2SH support for source / destination of addresses (protocol change: )
+    * Fixed issue with broadcasts of exactly 52 chars, by always adding a varint to specify the length. (protocol change: )
     * Many many improvements to test suite
 * v9.54.0 (2016-03-05)
     * Execute post install tasks when called via `pip`
--- a/counterpartylib/lib/messages/broadcast.py
+++ b/counterpartylib/lib/messages/broadcast.py
@@ -23,10 +23,13 @@ because it is stored as a four‐byte integer, it may not be greater than about
 
 import struct
 import decimal
+
 D = decimal.Decimal
 from fractions import Fraction
 import logging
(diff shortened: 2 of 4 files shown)


robby-dermody:
@adamkrellenstein do you have an opinion on this vs #855?

rubensayshi:
@robby-dermody @adamkrellenstein as you can see the test fails when parsing testnet because there's a message > 52 chars without a length byte.  
I think `util.enabled` shouldn't return TRUE for testnet but instead `protocol_changes.json` should have a height for testnet too? 

```
tx = {'block_hash': '0000000081e7903c94fef852ba45b9447b5a21a9c8e7ff9786ff2a8bc538d5a7', 'block_index': 317066, 'block_time': 1420588043, 'btc_amount': 0, ...}
message = b'T\xacX\x1b@E\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00Is this really going to be here forever? I better make it goo'
```

adamkrellenstein:
I don't have a strong opinion on which PR is better.

I'm okay with having `testnet` protocol changes as well, though there was a previous discussion where we decided not to. @ouziel-slama ?

rubensayshi:
@adamkrellenstein @ouziel-slama you recall the discussion about the testnet protocol changes?

adamkrellenstein:
What about it?

rubensayshi:
You said "I'm okay with having testnet protocol changes as well, though there was a previous discussion where we decided not to. @ouziel-slama ?"

I was curious if you recalled the reason why the choice was made not to

ouziel-slama:
Seems simpler to just update the `BLOCK_FIRST_TESTNET` than to have protocol changes in testnet and I'm not sure if it's worth to add complexity in the test suite for that.

PR #859, opened 2016-04-05
seperate the test for `blocks.parse_block` 

rubensayshi:
seperate the test for `blocks.parse_block` so that the other unittests don't have to use a scenario where we did not do `blocks.parse_block` for the latest block.

I ran into issues with this before, and now again, so why burden all the tests with an odd exception we make for just 1...

depends on #861


robby-dermody:
can you rebase this as I just merged in PR 860

PR #862, opened 2016-05-03
WIP: Let transactions crash in isolation 

rubensayshi:
using sqlite `with db:` will use the SAVEPOINT stuff that sqlite has, so we can nest `with db:` statements.  
wrapping `parse_tx` in a `with db:` and handling exceptions thrown will make it so that the TX is ignored and the node won't crash.

a while back @adamkrellenstein was looking into somehow doing this when we had every node on the network crash on 1 bug, I think he was in the understanding that `with db:` would use `BEGIN`/`COMMIT`/`ROLLBACK`, but because we're using sqlite we can actually leverage that it uses SAVEPOINT.

I'm still not 100% sure if this is the right thing to do, I'd rather not ignore errors at all ...  
Not only does this make it harder to notice errors if there are any, but if an error is thrown that isn't because the TX is funky (eg; maybe DB connection has a hickup or some other external / IO error) then the node will ignore that TX and it won't diverge it's consensus.

I personally don't have a machine available to run a node 24/7 so if any funky errors ever happen, though I've never had one while my node was up for development purposes which means my above concern isn't very relevant. (except for the 0conf TX not found err, but that's not in `parse_tx`). 

obviously this change should have a `protocol_changes.json` entry just incase.

also, can't seem to create a unittest for this, because there's nothing that is supposed to raise an exception...

### DIFF
--- a/counterpartylib/lib/blocks.py
+++ b/counterpartylib/lib/blocks.py
@@ -63,72 +63,82 @@ def parse_tx(db, tx):
     """Parse the transaction, return True for success."""
     cursor = db.cursor()
 
-    # Only one source and one destination allowed for now.
-    if len(tx['source'].split('-')) > 1:
-        return
-    if tx['destination']:
-        if len(tx['destination'].split('-')) > 1:
-            return
+    try:
+        with db:
+            # DEBUG STATEMENT TO TEST 'CRASHING' TX
+            # if tx['source'] == 'miDAc4uBw6X2cD41iRtfrPzQT2MezD8SNz':
+            #     print(tx)
+            #     raise Exception
 
-    # Burns.
-    if tx['destination'] == config.UNSPENDABLE:
-        burn.parse(db, tx, MAINNET_BURNS)
-        return
+            # Only one source and one destination allowed for now.
+            if len(tx['source'].split('-')) > 1:
+                return
+            if tx['destination']:
+                if len(tx['destination'].split('-')) > 1:
+                    return
 
-    if len(tx['data']) > 4:
-        try:
-            message_type_id = struct.unpack(config.TXTYPE_FORMAT, tx['data'][:4])[0]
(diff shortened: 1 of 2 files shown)


robby-dermody:
yes, I agree with many of your concerns, and If I had to choose between the two, I'd certainly want a simple network outage over a fork....

Given that, I would like to see if we can strike a balance between warning on something that is a specific parse error and will deterministically be encountered by all nodes on the network, vs something that is transient and node-specific (I/O errors, etc). The former class of errors, if it can be precisely outlined, could be warned on. Anything non-deterministic will need to be raised, as you noted, however.

In looking at the message `parse()` code, and the functions it calls (such as `validate()`), the two sources I see of potential non-determinism are any/all DB cursor operations, and any/all logger operations. (Do you agree? Maybe there are others?) One thing we could potentially do is wrap these and use that wrapped cursor and logger object instead. The wrapper could decorate the returned exception message somehow so that we could effectively detect any and all exceptions coming from a cursor or logger object and raise them. Anything else would be warned on, and the tx ignored.

The other alternative is to take an explicitly inclusive stance, and single out specific sections of the code and/or specific exceptions (e.g. `ValueError`) that we can ignore. However, I don't see the first as particularly good because it could bloat the code with exception handling logic, and you're never sure the exception was in the areas you've set up. The second isn't good either because a ValueError could be raised in a cursor object for something or another, for example...and just catching `ValueError` for example won't distinguish between the nature of the execution (deterministic or non-deterministic).

Thoughts?

rubensayshi:
we could catch `apsw.Error` specifically and crash if it is thrown, that's the only thing that I fear could happen that wouldn't be deterministic

robby-dermody:
I would say logging too, as it touches the file system (i.e. what if I have logs being logged to a remote mounted filesystem, and the network goes down)

chiguireitor:
Closing this to merge the conflict manually.

PR #863, opened 2016-05-04
 correct block hashes for unittests

rubensayshi:
use `sha256(sha256(block_index))` instead of `sha512(block_index)` for unittest blocks so that the hash is actually same length as real bitcoin blocks.


robby-dermody:
is this mostly for cosmetic/uniformity purposes? any advantages beyond that?

seems that using `util.dhash_string` could be less redundant here?

(although perhaps `util.dhash_string` itself could use some refactoring, as it's a bit redundant with `util.dhash` and should be able to just encode the text to bytes and call `util.dhash`, then decode the result?)

rubensayshi:
ok refactored `dhash_string` to use `dhash` and using `dhash_string` in `util_test.py`

robby-dermody:
+1 on the improved readability of those functions

PR #864, opened 2016-05-11
WIP: EVM on Counterparty

rubensayshi:
# WORK IN PROGRESS

I'll update the description later with some explaining, also I think it would be nice for review to actually put https://github.com/CounterpartyXCP/counterparty-lib/commit/8a8e1591d152ed314bff2361fc22e4f797d468d3 somewhere as a branch so the PR can be made against that, since the diff after the copypaste from pyethereum is a lot more relevant.
## Running it

It relies on serpent@develop, atm using my fork cause it's taking ethereum months to merge 2 tiny py3 fixes xD  
`python setup.py install_serpent` actually installs the correct tarball (from my repo)


robby-dermody:
when you get a chance, please reopen against `develop-evm`

axic:
Not sure if you it is being used in this incarnation, but it was mentioned before that Counterparty adds a few new opcodes to support assets.  I think a more canonical way right now could be using special "precompiled contracts" instead. That could mean a higher level of compatibility between the two EVMs.

robby-dermody:
@axic yes, that was the path we were going to take. @rubensayshi can confirm

chiguireitor:
Closing this in favor of other implementations in the future.

PR #866, opened 2016-05-18
Add code to add checkpoints

rubensayshi:
Goes hand-in-hand with counterparty-cli PR: https://github.com/CounterpartyXCP/counterparty-cli/pull/75

I'm planning to disable the old testnet-only EVM code and I need to update the checkpoints when doing so, this code is neccesary to do so and this is seperation of the stuff that doesn't actually change the checkpoint hashes, only reformatted.


counterpartylib/lib/config.py
```diff
@@ -95,6 +95,8 @@
 # Custom exit codes
 EXITCODE_UPDATE_REQUIRED = 5
 
+VERIFY_CHECKPOINTS = True
+VERIFY_OLD_HASH = True
```
robby-dermody: how about `VERIFY_STORED_HASH` instead?

robby-dermody:
please issue against `develop-evm`

PR #867, opened 2016-05-19
Testnet protocol changes

rubensayshi:


robby-dermody:
how are you using this feature? When would `testnet_block_index` not be 0?

rubensayshi:
https://github.com/CounterpartyXCP/counterparty-lib/pull/857

if testnet_block_index would be 0 then the reparse fails

robby-dermody:
please rebase and I will merge

PR #869, opened 2016-05-19
[WIP] add config option to disable mempool syncing

rubensayshi:
this makes for a very light weight node to run (at least memory wise)! 

CLI PR: https://github.com/CounterpartyXCP/counterparty-cli/pull/76

### DIFF
--- a/counterpartylib/lib/backend/__init__.py
+++ b/counterpartylib/lib/backend/__init__.py
@@ -36,7 +36,12 @@ def sortkeypicker(keynames):
     return getit
 
 def BACKEND():
-    return sys.modules['counterpartylib.lib.backend.{}'.format(config.BACKEND_NAME)] 
+    return sys.modules['counterpartylib.lib.backend.{}'.format(config.BACKEND_NAME)]
+
+def disable_unconfirmed_transactions():
+    global MEMPOOL_CACHE_INITIALIZED
+    MEMPOOL_CACHE_INITIALIZED = True
+    return BACKEND().disable_unconfirmed_transactions()
 
 def getblockcount():
     return BACKEND().getblockcount()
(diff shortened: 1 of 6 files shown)


robby-dermody:
I think this will raise an Exception if `unconfirmed` is set to true with any API calls, see https://github.com/CounterpartyXCP/counterparty-lib/blob/develop/counterpartylib/lib/backend/addrindex.py#L203
https://github.com/CounterpartyXCP/counterparty-lib/blob/develop/counterpartylib/lib/backend/addrindex.py#L126

https://github.com/CounterpartyXCP/counterparty-lib/blob/7d97da448a75dba16232e757db28936ceec56839/counterpartylib/lib/transaction.py#L398

I think this is fine an expected though, as it's explicit (i.e. if a node is not getting unconfirmed tx, and you ask for an API call that takes into account unconfirmed tx, you should get an error back)

robby-dermody:
@rubensayshi will test this more for unexpected side effects.

rubensayshi:
@robby-dermody updated to make it functional, 

the problem is though that when you try to construct multiple transactions within 1 block it won't know that certain UTXOs are already spent so it will keep constructing with the same UTXO.  

not really sure how to get around this atm without tracking 0conf TXs :/

robby-dermody:
Perhaps just some lightweight list that tracks utxos used for any transactions composed, and clears that out when a new block comes in and is processed? (You could maybe enhance the UTXO lock stuff we're doing to make it work in a mode that tracks until a block comes in and then expires it all... Or you could use make some other very simple way that will do this)... It's not perfect as it can't account for utxos used by other nodes, but may be good enough

chiguireitor:
@rubensayshi is this PR ready or is there something left to do for it to work? I know it's ages since this PR was opened, but would like to know if you recall something about it

chiguireitor:
Closing this, it has become incompatible with the indexd backend due to the elimination of the `searchrawtransactions` call

Issue #875, opened 2016-05-21
ensure utxo reuse is prevented when constructing new transactions 


David-R-Allen:
Counterparty frequently constructs transactions that are double spends or are inexplicably rejected by Bitcoin Core's broadcast function. Poloniex track inputs manually and have two fallbacks for broadcasting, but even so, sometimes the transactions never propagate.

robby-dermody:
Note that it's possible that this happens when there are a relative lot of outbound send requests being made, and counterparty-lib may end up reusing utxos when constructing new transactions. I have a branch 'utxocache' that needs to be tested, that may address this issue. It's between who can get to it first, @rubensayshi or I (or anyone else that wants to help out :)

robby-dermody:
closing as this is in `develop`

PR #883, opened 2016-05-24
WIP: implement UTXOcache

robby-dermody:
also see: https://github.com/CounterpartyXCP/counterparty-cli/pull/83

this is an attempted solution for #875 based on my hypothesis of what I think the issue is (https://github.com/CounterpartyXCP/counterparty-lib/issues/875#issuecomment-220782121)

@rubensayshi, can you try this one out? (I left in debugging statements in the commit to make it easier to debug)

in my testing (via issuing numerous "send" calls to `counterparty-client` to get unsigned raw txhashes), it successfully builds a UTXO cache, adding UTXOs that are used in earlier construct calls...UTXOs expire and are added back (and will show up again in the UTXO set if that composed tx is not broadcast)

I'm using a DictCache for the containing dict (of source addresses) to limit size over long runtimes. each source address then uses a TTLCache, which will auto expire individual UTXO entries for that source over some configurable period of time.

If you like the method, we'd have to talk about the optimal value for `DEFAULT_UTXO_AGING_CACHE_MAX_AGE` especially. Also, this patch will require a corresponding patch to `counterparty-cli`

EDIT: it will need tests too. it's failing the tests now because it's probably expiring out the sources UTXO set as the test suite runs. maybe disable it for the current test suite, and then have separate unit tests that enable it and test that expiry functionality in a more isolated manner


rubensayshi:
needs expiringdict to version bump to include their py3 fixes; https://github.com/mailgun/expiringdict/pull/20

robby-dermody:
I actually moved off of expiringdict to cachetools due to the better python3 support. See https://github.com/CounterpartyXCP/counterparty-lib/commit/88b3ce4405c025514552b12c842235618931fba7

rubensayshi:
@robby-dermody you need to

```sh
py.test -s --savescenarios --verbose --verbosediff --skiptestbook=all counterpartylib/test/integration_test.py
./tools/copyscenarios.py
git add counterpartylib/test/fixtures/
git commit -m "updated fixtures"
```

robby-dermody:
closed in favor of #895

Issue #871, opened 2016-05-20
User newer Sqlite?


unsystemizer:
In setup.py we download and build sqlite-amalgamation 3.8.7.3-r1

Quite a lot of (performance and other) improvements and bug fixes happened in Sqlite since 3.8 (now 3.13, Ubuntu 16.04 ships with 3.11). 

Could we consider using a newer version? 
What kind of "CP DB stress test" would be appropriate?

unsystemizer:
I've been using 3.12.2 for days, no apparent issues.

rubensayshi:
closing, pending https://github.com/CounterpartyXCP/counterparty-lib/issues/890 merge

PR #890, opened 2016-05-30
[WIP] upgrade APSW/SQLite to 3.12.2 and enlarge page_cache

rubensayshi:
Fixes: https://github.com/CounterpartyXCP/counterparty-lib/issues/871

Will update APSW / SQLite, testing it now...

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,4 +1,9 @@
 ## Library Versions ##
+* Next Release
+    * Upgrade SQLite/APSW to v3.12.2
+        * Run `python3 setup.py install_apsw` to upgrade your installed version.
+        * Run `./tools/upgradesqlitepagesize.py <PATHTOYOURDB>` while your node is OFFLINE to upgrade.
+          Most likely the path to your DB is `~/.local/share/counterparty/counterparty.db`.
 * v9.55.0 (2016-07-11)
     * P2SH support for source / destination of addresses (protocol change: 423888)
     * Moved check for invalid broadcast to better place to prevent broadcasting a cancel on a locked feed (protocol change: 423888)
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Then, download and install `counterparty-lib`:
 ```
 $ git clone https://github.com/CounterpartyXCP/counterparty-lib.git
 $ cd counterparty-lib
-$ sudo pip3 install -r requirements.txt
+$ sudo pip3 install --upgrade -r requirements.txt
 $ sudo python3 setup.py install
 ```
 
@@ -47,7 +47,7 @@ Followed by `counterparty-cli`:
 ```
 $ git clone https://github.com/CounterpartyXCP/counterparty-cli.git
 $ cd counterparty-cli
-$ sudo pip3 install -r requirements.txt
+$ sudo pip3 install --upgrade -r requirements.txt
 $ sudo python3 setup.py install
 ```
 
--- a/counterpartylib/lib/database.py
+++ b/counterpartylib/lib/database.py
@@ -70,7 +70,7 @@ def get_connection(read_only=True, foreign_keys=True, integrity_check=True):
     logger.debug('Creating connection to `{}`.'.format(config.DATABASE))
 
     if read_only:
-        db = apsw.Connection(config.DATABASE, flags=0x00000001)
+        db = apsw.Connection(config.DATABASE, flags=apsw.SQLITE_OPEN_READONLY)
     else:
         db = apsw.Connection(config.DATABASE)
     cursor = db.cursor()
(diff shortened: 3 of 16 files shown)


rubensayshi:
Relevant changes in SQLite changelog since 3.8.7 from http://www.sqlite.org/changes.html:

#### 2016-03-29 (3.12.0)
The SQLITE_DEFAULT_PAGE_SIZE is increased from 1024 to 4096. The SQLITE_DEFAULT_CACHE_SIZE is changed from 2000 to -2000 so the same amount of cache memory is used by default. See the application note on the version 3.12.0 page size change for further information.

rubensayshi:
tests fail cuz the fixtures .sql files contain "PRAGMA page_size=1024;" and after upgrade will contain "PRAGMA page_size=4096;".  

so if we're gonna upgrade this we need devs to upgrade to make their tests pass ;)

rubensayshi:
added command line tool to do the upgrade:
 - `python ./tools/upgradesqlitepagesize.py ~/.local/share/counterparty/counterparty.testnet.db`
 - `python ./tools/upgradesqlitepagesize.py ~/.local/share/counterparty/counterparty.db`

untested on mac/windows

rubensayshi:
Another breaking change; can't do `backup` on a DB that's already been written to, but the first `initialise_db` that was being done before the `backup` was useless anyway, so removed it.

unsystemizer:
@robby-dermody on my Ubuntu 16 `apt-get upgrade` overwrote the manually installed Sqlite 3.12.2 with 3.13.x. I don't know if this could impact anything in Docker or otherwise - if yes maybe we should "hold" sqlite3 to prevent it from being upgraded?

robby-dermody:
@unsystemizer yes, the docker setup will avoid those kinds of situations

robby-dermody:
@rubensayshi if we merge this, notes on how to upgrade will need to be written in `Changelog.md` (or somewhere else and linked to `Changelog.md`)

on that topic, probably a good practice for us to commit additions to `Changelog.md` as part of the actual PR, instead of all at once in preparation for a release. Thoughts?

rubensayshi:
yea, I'll try making a habbit of adding thing to `Changelog.md`.

not sure how we can make it extra clear though that they should be doing something, like upgrading sqlite

unsystemizer:
With Ubuntu 16.04 (which we don't yet "officially" support although it works) they would need to downgrade. 
counterparty-lib's readme.md could contain a note on that (although we can leave Ubuntu 14.04 as the primary/default distro and version we support).

rubensayshi:
@unsystemizer what do you mean downgrade?

unsystemizer:
I installed apsw/Sqlite 3.12.2 on my system (as I was testing it for a while). Earlier this week I executed `apt-get upgrade` and I saw Sqlite v3.1.13 was installed over it. 
As I mentioned in https://github.com/CounterpartyXCP/counterparty-lib/issues/871, Ubuntu 16.04 shipped with v3.11 so it was all good as far as my upgrade was concerned until they released v3.13.

Now, I didn't notice any issues, but at least in theory apsw 3.12.2 should be paired with a different version of Sqlite3 and apsw 3.12.2 is the latest release, so obviously not holding sqlite3 from automatic OS updates creates a bad situation. It's not a problem on Ubuntu 14.04 because you won't see a newer version, but on Ubuntu 16.04 it's different, that's why I asked.

rubensayshi:
afaik when apsw is build (`python setup.py install_apsw`) it downloads the corresponding sqlite version to compile with.  
because the apsw versions correspond with the sqlite versions and can't build against a different one.

unsystemizer:
Yes, that's why a Fed Node-like script should "hold" sqlite3 at the pre-set version (3.12.2, at the moment). 
pip3 can't control OS-level stuff, but Fed Node script can. Regarding the former, I wonder what happens with pip, counterparty-lib and apsw if someone tries to install counterparty-lib on the latest Ubuntu 16.04 where sqlite3 is v3.13..

rubensayshi:
when apsw install it downloads a 'local' copy of the sqlite version it needs, it's not using the globally installed sqlite

robby-dermody:
@rubensayshi if you think this is ready feel free merge it. I suspect the fednode counterparty-server container on develop may need an tweak to have that specific version, and the bootstrap DB will remain at the 1024 kb page size etc until the next release of course. I can do any necessary fednode updates when I get back.

unsystemizer:
@robby-dermody check out https://github.com/CounterpartyXCP/counterparty-lib/issues/950, I think I hit the 2GB limit which (I'm still troubleshooting) could be the limit of pages with 1kb block size.

unsystemizer:
The maximum Sqlite3 DB size using 1KB pages is 2 TB (earlier today I mistakenly calculated 2 GB), so there's no danger of hitting that limit soon, it seems.

chiguireitor:
Closing this as it is being manually merged to solve conflicts.

PR #894, opened 2016-05-31
Freeze Dependencies

rubensayshi:
We've survived pretty long without issues, but v1.0 for `pysha3` was just released which 'fixes' the sha3 function we used for contract_id's:

> pysha3 1.0 is not compatible with pysha3 0.3!
> 
> pysha3 < 1.0 used the old Keccak implementation. During the finalization of SHA3, NIST changed the delimiter suffix from 0x01 to 0x06. The Keccak sponge function stayed the same. pysha3 1.0 provides the previous Keccak hash, too. 

this PR adds a test for it and freezes the deps at the latest versions which have proven to be working.


robby-dermody:
The versions should be specified in `requirements.txt` instead of `setup.py`:

See http://python-packaging-user-guide.readthedocs.io/en/latest/requirements/:
```
Whereas install_requires defines the dependencies for a single project, Requirements Files are often used to define the requirements for a complete python environment.

Whereas install_requires requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving repeatable installations of a complete environment.
```

https://caremad.io/2013/07/setup-vs-requirement/ is helpful, too

robby-dermody:
I read up more on the "abstract" vs "concrete" requirements distinction, and am fine merging this, as abstract requirements can have a version number associated with them -- they must just not be resolved to a specific package repository

PR #899, opened 2016-06-13
Fee per byte

rubensayshi:
Depends on: https://github.com/CounterpartyXCP/counterparty-lib/pull/909

No longer use the old bitcoind way of calculating fees (`0.0001 BTC/kb` with kb=`ceil(size / 1000)`).  
Instead actually calculated based on the size (`0.0001 BTC/kb` with kb=`size/1000`).

The old formula would give about ~0.0003 BTC/kb for small TXs (sends, DEX orders), for bigger TXs that are still under 1kb it would result in significantly lower fees and very shitty confirmation times.

0.0003 BTC/kb is actually already on the low side, but we've been managing ...  
see http://bitcoinfees.21.co/ and https://www.blocktrail.com/BTC for reference.  
I've (idk if you'd call it fortunately) had to deal with estimated fees for the Blocktrail wallet on numerous occasions and the last months it rarely drops below 0.0004 BTC/kb, most of the day can be considered 'peak hours' by now and needs > 0.0004 BTC/kb...

so this PR removes the ceiling of the size to KBs and actually use the proper size to calcuate the fee, it also bumps the `DEFAULT_FEE_PER_KB` to 0.00025 BTC/kb.

**unfortunately this touches a ton of fixtures/vectors because when the fee changes the hex of txs change and the txids change ...**

I'm doing another PR straight after (just split this for reviewing) to start using the `bitcoind estimatefee` for a dynamic fee, with e minimum of the `DEFAULT_FEE_PER_KB` (0.00025 BTC/kb) for sanity.  
I'll explain everything here and we can discuss here ;-)

`bitcoind estimatefee` does a slow moving average (by decaying the older values, some smart fix for not having to store a lot of data) and takes `nblocks` as arguments, it will give you a value for what it estimates the fee needs to be to get 75% chance to get into the next `nblocks`.

it has some flaws though; 
 - as of late it's unable to estimate any sane fee for `nblocks=1` (as you can see on statoshi there's no line for it anymore).
 - a fresh bitcoind install won't have any data and won't be able to report a fee until it has received a bunch of blocks after being synced up...  
  **also applies to someone who just starts his bitcoind when he needs it and leaves it offline the rest of the time... it will just have a 'gap' in it's data and it's estimate will be off for a while, though the nature of the slow moving average means it will never been off by a lot (see statoshi)** 
 - it could (correctly) estimate a very high fee when someone does another 'stresstest'.

things still to consider / discuss:
 - should be a maximum as sanity limit? would we update counterparty-lib every once in a while with new maximums if required then?
 - what value to use for nblocks?
 - I think we should display the current fee estimate in counterwallet and possibly an 'advanced' section in the dialogs to set your own fee?

see: https://statoshi.info/dashboard/db/fee-estimates for chart of `bitcoind estimatefee` historical data (you can change the chart to last 30d at the top right).


unsystemizer:
CW Advanced Settings: yes, that would be great.
Some users prize the cost, others the speed. Maybe before each Send (I'm not sure if this can easily be done for DEx and other non-Send transactions) we could provide an estimate which, when clicked on, would take the user to Advanced Settings where the slider or input box would be provided for them.
If they just proceeded with Send, the estimate would be used.
Maybe the slider could be provided in the Send dialog box to make it easier to set desired fee for each transaction without going through 2-3 clicks.

We should consider whether we want to use nominal (0.002, 0.003) or descriptive (Lo, Med, Hi) or maybe time estimates. Each approach has its disadvantages.

I would suggest to hard-code the maximum value in CW, to avoid catastrophic mistakes/typos.

robby-dermody:
If estimatefee is unable to estimate any sane fee for blocks=1, why does estimate_fee_per_kb_nblocks default to 1? Wouldn't some other value be a safer default? (I see your question above regarding this, so I guess that is just your initial choice...). What is the lowest value of blocks where the response is "sane"/valid?

PR #901, opened 2016-06-14
Counterparty micropayments

F483:
Please don't except just yet, this is still work in progress and currently meant for review purposes.


F483:
Will reopen with a clean commit history later today.

PR #902, opened 2016-06-15
improved APSW install to downgrade when newer version is installed.

rubensayshi:
preperation for the APSW upgrade, this will downgrade APSW if a newer version is installed.

neccesary because CircleCI is caching the venv and as a result tests are messed up (cause the `page_size` is in the intergration tests output.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,9 +1,11 @@
 # Keep this as non-docker for now, to give some variety to our test configurations, as travis builds with docker
 dependencies:
     override:
-        - rm -rf /home/ubuntu/virtualenvs/venv-3.4.1/bin/serpent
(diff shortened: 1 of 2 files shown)


rubensayshi:
unfortunately downgrading APSW doesn't work by just installing 3.8 again ... 3.12 will still be there and supersede it :/

`rm -rf /home/ubuntu/virtualenvs/venv-*/lib/python3.4/site-packages/apsw*` does the trick for removing it from the CircleCI cache so it actually installs again ...

`pip uninstall apsw` would also do the trick, but that would make `pip` a dependency ...

robby-dermody:
I don't think pip being a dependency is necessarily a bad thing, and pip uninstall is less crude than rm -rf, but we have been using this approach and I'm fine with the changes here

PR #903, opened 2016-06-15
add `disable_utxo_locks` 

rubensayshi:
add `disable_utxo_locks` to `transaction.construct` and `compose_transaction` to disable the locking of selected UTXOs for when the 'user' doesn't intend to broadcast the TX (straight away).

for payment channels you want to compose a TX but don't broadcast it (straight away) and if the UTXOs are locked you'll run into insufficient balance errors if you compose multiple times in quick succession. 

depends on: https://github.com/CounterpartyXCP/counterparty-lib/pull/902

### DIFF
--- a/counterpartylib/lib/api.py
+++ b/counterpartylib/lib/api.py
@@ -68,7 +68,7 @@ API_TRANSACTIONS = ['bet', 'broadcast', 'btcpay', 'burn', 'cancel',
 COMMONS_ARGS = ['encoding', 'fee_per_kb', 'regular_dust_size',
                 'multisig_dust_size', 'op_return_value', 'pubkey',
                 'allow_unconfirmed_inputs', 'fee', 'fee_provided',
-                'unspent_tx_hash','custom_inputs']
+                'unspent_tx_hash', 'custom_inputs', 'dust_return_pubkey', 'disable_utxo_locks']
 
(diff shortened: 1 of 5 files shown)


counterpartylib/lib/api.py
```diff
@@ -68,7 +68,7 @@
 COMMONS_ARGS = ['encoding', 'fee_per_kb', 'regular_dust_size',
                 'multisig_dust_size', 'op_return_value', 'pubkey',
                 'allow_unconfirmed_inputs', 'fee', 'fee_provided',
-                'unspent_tx_hash','custom_inputs']
+                'unspent_tx_hash', 'custom_inputs', 'dust_return_pubkey', 'disable_utxo_locks']
```
robby-dermody: Was adding dust return pubkey here something we just forgot in the past?

robby-dermody:
Merging, however please see my code comment above

counterpartylib/lib/api.py
```diff
@@ -294,13 +294,6 @@ def compose_transaction(db, name, params,
         if not script.is_fully_valid(binascii.unhexlify(pubkey)):
             raise script.AddressError('invalid public key: {}'.format(pubkey))
 
-    # copy and remove dust_return_pubkey from params into var
```
robby-dermody: Ah, ok, I see what you did

robby-dermody:
At some point we need to update the API docs for these additional params...

PR #904, opened 2016-06-19
Fix BTC order status bug

loon3:
Currently, orders with BTC remain in open status until expire_index is reached regardless of whether they're matched.  

The change to order.py will set a status of 'expired' for any BTC sell order that is part of an order match if a valid btcpay tx is not received within 20 blocks of the match.

The change to btcpay.py will set a status of 'filled' for any order that is part of an order match with BTC if a valid btcpay tx is received within 20 blocks of the match and either give_remaining or get_remaining is equal to zero.


counterpartylib/lib/messages/order.py
```diff
@@ -426,6 +426,14 @@ def parse (db, tx, message):
 
         problems = validate(db, tx['source'], give_asset, give_quantity, get_asset, get_quantity, expiration, fee_required, tx['block_index'])
         if problems: status = 'invalid: ' + '; '.join(problems)
+
```
robby-dermody: this should be in a different PR...seems like a separate issue

loon3:
cancelling this PR to open three separate PRs for clarity

PR #905, opened 2016-06-24
Proof of Concept: P2SH Encoding

rubensayshi:


counterpartylib/lib/transaction.py
```diff
@@ -523,12 +326,30 @@ def construct (db, tx_info, encoding='auto',
     else:
         change_output = None
 
+    unsigned_pretx_hex = None
+    unsigned_tx_hex = None
+
+    if encoding == 'p2sh':
+        unsigned_pretx = serializer.serialise_p2sh_pretx(inputs, data_output, change_output, pubkey=dust_return_pubkey)
+        unsigned_pretx_hex = binascii.hexlify(unsigned_pretx).decode('utf-8')
+        print('unsigned_pretx_hex', unsigned_pretx_hex)
```
robby-dermody: debug statements to remove (and several lines below)

PR #906, opened 2016-06-25
test suite refactoring [take 101]

rubensayshi:
yet another refactoring of the test suite...

it started out as creating a small 'complex unittest' template for @f483 to use for the payment channels, but I got sucked in and did a bunch more things ...

I changed the test suite to actually track UTXOs (still initial state setup from the same .json file), this makes it possible (for @f483) to actually test different states of a chain of transactions and test between confirmed and unconfirmed states etc.  
It's not perfect, but it will do.

Once I implemented that it resulted in a big diff in the test scenarios, because their inputs changed and as a result the txid, etc...  
This happens a lot ... and I was tired of it; so I changed a few things that should ensure that vectors change less when we make even the smallest change in a scenario.  

So arc4 is being mocked with a fixed seed to avoid that ever changing in the vectors, and when constructing the txid of transactions the inputs are blanked so that different inputs also no longer change the txids.

oh and I got rid of the `rawtransactions.db` file, now using `:memory:`, also fixes another frustrating diff that keeps coming back all the time.

See inline comments for more ;-)


.gitignore
```diff
@@ -9,6 +9,7 @@ profile.txt
 
 # files generated by py.test
 /counterpartylib/test/fixtures/scenarios/*.new.*
+/counterpartylib/test/fixtures/rawtransactions.db
```
rubensayshi: bye useless file

counterpartylib/lib/api.py
```diff
@@ -690,7 +690,7 @@ def search_raw_transactions(address, unconfirmed=True):
 
         @dispatcher.add_method
         def get_unspent_txouts(address, unconfirmed=False, unspent_tx_hash=None):
-            return backend.get_unspent_txouts(address, unconfirmed=unconfirmed, multisig_inputs=False, unspent_tx_hash=unspent_tx_hash)
+            return backend.get_unspent_txouts(address, unconfirmed=unconfirmed, unspent_tx_hash=unspent_tx_hash)
```
rubensayshi: this argument was unused throughout the code, removed it everywhere.

counterpartylib/lib/blocks.py
```diff
@@ -492,7 +492,7 @@ def get_address(scriptpubkey):
 
             if ctx.is_coinbase():
                 raise DecodeError('coinbase transaction')
-            obj1 = ARC4.new(ctx.vin[0].prevout.hash[::-1])
+            obj1 = arc4.init_arc4(ctx.vin[0].prevout.hash[::-1])
```
rubensayshi: arc4 moved into a wrapper so we can easily mock it.

counterpartylib/test/arc4_test.py
```diff
@@ -0,0 +1,142 @@
...
+    # b'\x01' * 32 encrypt
+    k = arc4.init_arc4(b'\x01' * 32)
+    assert k.encrypt(text) == b'\xaa}\xfa5\xcaY:'
+    
+    # b'\x01' * 32 decrypt
+    k = arc4.init_arc4(b'\x01' * 32)
+    assert k.decrypt(b'\xaa}\xfa5\xcaY:') == text
+
+
+@conftest.add_fn_property(DISABLE_ARC4_MOCKING=True)
```
rubensayshi: still a way to disable the mocking in the test suite and still test the actual results.

counterpartylib/test/conftest.py
```diff
@@ -126,20 +137,139 @@ def cp_server(request):
...
+    log.set_up(log.ROOT_LOGGER, verbose=True, console_logfilter=os.environ.get('COUNTERPARTY_LOGGING', None))
+
+
+@pytest.fixture(scope='function', autouse=True)
+def mock_utxos(rawtransactions_db):
+    with open(util_test.CURR_DIR + '/fixtures/unspent_outputs.json', 'r') as listunspent_test_file:
+        util_test.MOCK_UTXO_SET = MockUTXOSet(json.load(listunspent_test_file), rawtransactions_db)
+
+    util_test.UNIQUE_DUMMY_TX_HASH = {}
+    util_test.TX_INDEX = 1
```
rubensayshi: `scope=function, autouse=True` so this is where we reset stuff between each test

counterpartylib/test/complex_unit_test.py
```diff
@@ -0,0 +1,141 @@
+import pprint
```
rubensayshi: this is the template 'complex unittest' which demostrates the use of the mocked UTXO set and is a good base for any tests that aren't just 1 function call (which almost all our current vectors are).

counterpartylib/test/integration_test.py
```diff
@@ -5,16 +5,18 @@
 from counterpartylib.test import util_test
 
 
-def test_scenario(scenario_name, base_scenario_name, transactions, rawtransactions_db):
```
rubensayshi: `rawtransactions_db` is now updated by the `MockUtxoSet`

counterpartylib/test/unit_test.py
```diff
@@ -8,6 +8,7 @@
 FIXTURE_SQL_FILE = CURR_DIR + '/fixtures/scenarios/unittest_fixture.sql'
 FIXTURE_DB = tempfile.gettempdir() + '/fixtures.unittest_fixture.db'
 
+@conftest.add_fn_property(DISABLE_ARC4_MOCKING=True)
```
rubensayshi: arc4 mocking is disabled for the unit_test vectors so they don't all need to be update (which is a hell)

counterpartylib/test/util_test.py
```diff
@@ -146,35 +158,100 @@ def create_next_block(db, block_index=None, parse_block=True):
...
+
+def insert_raw_transaction(raw_transaction, db):
     """Add a raw transaction to the database."""
+    cursor = db.cursor()
+
     # one transaction per block
     block_index, block_hash, block_time = create_next_block(db, parse_block=False)
 
-    cursor = db.cursor()
+    tx_hash = dummy_tx_hash(raw_transaction)
```
rubensayshi: use blanked out tx

counterpartylib/test/util_test.py
```diff
@@ -146,35 +158,100 @@ def create_next_block(db, block_index=None, parse_block=True):
...
+        txin.script = b'\x00'
+
+    # check if we have a change output
+    if tx.txs_out[-1].coin_value not in [0, config.DEFAULT_REGULAR_DUST_SIZE, config.DEFAULT_MULTISIG_DUST_SIZE, config.DEFAULT_OP_RETURN_VALUE]:
+        tx.txs_out[-1].coin_value = 0  # set change to 0
+
+    tx_id = tx.id()
+
+    # check we haven't created this before (if we do 2 exactly the sends for example)
+    if tx_id in UNIQUE_DUMMY_TX_HASH:
```
rubensayshi: still ensure uniqueness post-blanked-out

tools/updatetxids.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python3
```
rubensayshi: simple tool I used to update the txids in the python files, leaving it here for later use.

rubensayshi:
this needs review and also before merging will need a rebase to update vectors (dont let github fool you)

counterpartylib/test/complex_unit_test.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # balance before send
+    alice_balance = util.get_balance(server_db, alice, 'XCP')
+    bob_balance = util.get_balance(server_db, bob, 'XCP')
+    assert alice_balance == 91950000000
+    assert bob_balance == 0
+
+    # create send
+    v = int(100 * 1e8)
+    send1hex = util.api('create_send', {'source': alice, 'destination': bob, 'asset': 'XCP', 'quantity': v})
```
robby-dermody: do you see any potential reliability issues using util.api vs adding a more direct method (i.e. direct to code vs JSON RPC request over localhost)?  I don't really, but it may be a bit quicker/cleaner.

counterpartylib/test/conftest.py
```diff
@@ -69,16 +78,17 @@ def pytest_addoption(parser):
     parser.addoption("--scenario", action="append", default=[], help="list of scenarios to test")
     parser.addoption("--gentxhex", action='store_true', default=False, help="generate and print unsigned hex for *.compose() tests")
     parser.addoption("--savescenarios", action='store_true', default=False, help="generate sql dump and log in .new files")
+    parser.addoption("--alternative", action='store_true', default=False)
```
robby-dermody: adding a help string here would be useful

counterpartylib/test/conftest.py
```diff
@@ -104,6 +114,7 @@ def api_server(request, cp_server):
     api_server.start()
     for attempt in range(5000):  # wait until server is ready.
         if api_server.is_ready:
+            time.sleep(0.5)  # extra time window
```
robby-dermody: was it failing out from time to time without this?

counterpartylib/test/conftest.py
```diff
@@ -126,20 +137,139 @@ def cp_server(request):
...
+
+    return decorator
+
+
 @pytest.fixture(scope='function', autouse=True)
-def init_mock_functions(monkeypatch, rawtransactions_db):
+def init_mock_functions(request, monkeypatch, mock_utxos, rawtransactions_db):
     """Test suit mock functions.
 
     Mock functions override default behaviour to allow test suit to work - for instance, date_passed is overwritten
```
robby-dermody: "test suite" instead of "test suit"

counterpartylib/test/unit_test.py
robby-dermody: will it be enabled at some point in the future?

counterpartylib/test/integration_test.py
```diff
@@ -5,16 +5,18 @@
...
+def test_scenario(scenario_name, base_scenario_name, transactions):
```
robby-dermody: maybe I'm missing something, but I don't see an updated call to this function anywhere in the diff set (was trying to trace through the differences with `--savescenarios`)

robby-dermody:
utACK. Looks good at first blush. made some comments. will wait for your rebase.

robby-dermody:
@rubensayshi plz merge if you think this is ready after your final review

rubensayshi:
depends on https://github.com/CounterpartyXCP/counterparty-lib/pull/931 cuz serpent

counterpartylib/lib/api.py
```diff
@@ -392,7 +393,7 @@ def stop(self):
 
     def run(self):
         logger.info('Starting API Server.')
-        db = database.get_connection(read_only=True, integrity_check=False)
+        self.db = self.db or database.get_connection(read_only=True, integrity_check=False)
```
rubensayshi: `APIServer` now has a bound DB instance so we can swap it out for tests

counterpartylib/test/complex_unit_test.py
rubensayshi: @F483 actually needs to test stuff that is using the api server, so we need this test to cover the basics of that

robby-dermody:
I see that the picopayments PR depends on the complex unit tests patchset (https://github.com/rubensayshi/counterparty-lib/pull/10 ? ). Seems like this PR is the same? When you get the chance, could you get this PR cleaned up and rebased, and then we can merge it in. At that point, Fabian can rebase his picopayments PR and we can begin review of his actual changes (right now his PR includes much of this code)

deweller:
@rubensayshi - can you please review this again and see if it all still looks good to you?  If so, I will review it as well and we can merge it.

PR #911, opened 2016-07-02
CORS headers tweak to allow auth from web clients

robby-dermody:
according to @jdogresorg:

> This will allow browsers to make XHR JSON-RPC requests to the counterparty API.
> Without this header change, the brower is not able to make XHR requests to a counterparty API which has a username/password set

@jdogresorg please test this and see if it addresses your issue. if so, I can merge


jdogresorg:
Looks good.

jdogresorg:
actually, it appears that if you enable Access-Control-Allow-Credentials, you need to specify exactly what origins are allowed to access the API via Access-Control-Allow-Origin and wildcards are not allowed.

Works fine without
response.headers['Access-Control-Allow-Credentials'] = 'true'

PR #912, opened 2016-07-09
Btc order filled

loon3:
Update give and get btc order status as filled if order_match is completed, checks for other pending orders first


loon3:

PR #913, opened 2016-07-09
Btc order minimum

loon3:
Mark btc buy or sell orders as invalid if below btc minimum (0.001 BTC)


loon3:

PR #914, opened 2016-07-09
Btc sell expire on match expire

loon3:
Expire btc sell order if order match expires, first checks for pending order matches involving sell order txid


loon3:

PR #915, opened 2016-07-09
Update btc give and get order status as filled if order_match is completed

loon3:


chiguireitor:
Merged to develop manually

PR #917, opened 2016-07-09
Btc sell expire on order_match expire

loon3:


loon3:
@robby-dermody I looked this one over again and it's only canceling btc orders involved in expired order matches (not all btc sell orders from a particular address) so its not really a penalty, just resolving the btc sell order status appropriately

PR #925, opened 2016-08-01
avoid UTXO lock being set and then bytespersigop fix kicking in

rubensayshi:
conflict we didn't detect when the bytespersigop fix kicks in and there's only limited utxos then the locks already set cause it to fail


robby-dermody:
can you rebase when you get the chance and then we can merge this?

PR #959, opened 2017-02-08
Implements Subassets cip-4

deweller:
https://github.com/CounterpartyXCP/cips/blob/master/cip-0004.md


deweller:
The messages hash don't match yet due to the addition of the asset_longname field in the issuances table.  I'll work on that.

PR #961, opened 2017-02-09
Implements Subassets cip-4

deweller:
https://github.com/CounterpartyXCP/cips/blob/master/cip-0004.md


robby-dermody:
do you have multiple commits for this, or just this one monster commit? (generally it's better to have multiple, incremental commits as the work is done...it's not absolutely essential for the initial work, but definitely as reviews come in and adjustments are made)

deweller:
I just have the one big commit.

Almost all of the work is in issuances.py and util.py.  There are small changes in api.py, blocks.py, config.py and database.py.  Everything else is related to testing.

rubensayshi:
reread the CIP again, feedback that should probably be added to both the CIP and implementation:
 - `;;` (`3b3b`) is used as seperator for the metadata to start, we should maybe consider disallowing that in the `description` field
 - if we ever want to use the same encoding for more fields, define how and what we'd be doing when there's multiple of the same field in there, eg; `;;lPIZZA.DOMINOS;;lPIZZA.NEWYORK` ?

counterpartylib/lib/config.py
```diff
@@ -7,7 +7,7 @@
 # Versions
 VERSION_MAJOR = 9
 VERSION_MINOR = 55
-VERSION_REVISION = 1
+VERSION_REVISION = 2
```
rubensayshi: should be `MINOR` bump I think? but we usually bump this in `master` before we do a release, but leave this for now so it triggers a reparse for people trying to run the PR I suppose

counterpartylib/lib/database.py
```diff
@@ -58,6 +58,11 @@ def exectracer(cursor, sql, bindings):
         log.message(db, bindings['block_index'], command, category, bindings)
     # Record alteration in computation of message feed hash for the block
     if category not in skip_tables_block_messages:
+        # don't include asset_longname as part of the messages hash 
+        #   until subassets are enabled
+        if category == 'issuances' and not util.enabled('subassets'):
+            if isinstance(bindings, dict) and 'asset_longname' in bindings: del bindings['asset_longname']
+
```
robby-dermody: honestly I'd rather just bump MINOR than resort to magic like this

counterpartylib/lib/config.py
robby-dermody: +1, then you don't need to dynamically delete the bindings in database.py

counterpartylib/lib/messages/issuance.py
```diff
@@ -54,7 +55,14 @@ def initialise(db):
                       source_idx ON issuances (source)
                    ''')
 
-def validate (db, source, destination, asset, quantity, divisible, callable_, call_date, call_price, description, block_index):
+    # Add asset_longname for sub-assets
+    #   SQLite can’t do `ALTER TABLE IF COLUMN NOT EXISTS`.
+    columns = [column['name'] for column in cursor.execute('''PRAGMA table_info(issuances)''')]
+    if 'asset_longname' not in columns:
+        cursor.execute('''ALTER TABLE issuances ADD COLUMN asset_longname TEXT''')
+        cursor.execute('''CREATE UNIQUE INDEX IF NOT EXISTS asset_longname_idx ON issuances (asset_longname)''')
```
robby-dermody: tab this line one back, for consistency with the other CREATE INDEX lines (plus it will add the index back if the index is removed for some reason but the column name persists)

counterpartylib/lib/config.py
deweller: A reparse is a significant interruption for those of us running counterparty in production.  It brings down a production counterparty server for hours.

counterpartylib/lib/database.py
deweller: I really don't want a reparse if we can avoid it.

What is your concern here?

deweller:
> reread the CIP again, feedback that should probably be added to both the CIP and implementation:
> 
> - ;; (3b3b) is used as seperator for the metadata to start, we should maybe consider disallowing that in the description field

`compose` does not check for `;;l` in the description field.  So if a user accidentally includes `;;l` in the description field, compose will allow it, but the issuance will be marked invalid when the transaction confirms.

Perhaps we should parse for it and throw an error.  Or maybe we think that `;;l` is too uncommon to worry about.

> - if we ever want to use the same encoding for more fields, define how and what we'd be doing when there's multiple of the same field in there, eg; ;;lPIZZA.DOMINOS;;lPIZZA.NEWYORK ?

Only the rightmost field is parsed. In this cause `;;lPIZZA.DOMINOS` remains part of the description and the subasset longname `PIZZA.NEWYORK` is acted on.

counterpartylib/lib/messages/issuance.py
deweller: Will do.

counterpartylib/lib/config.py
robby-dermody: yes. I think I could be swayed either way. let's see what ruben thinks

counterpartylib/lib/util.py
```diff
@@ -198,6 +202,79 @@ def get_asset_name (db, asset_id, block_index):
...
+#   if it has a suffix like ;;lPARENT.child, then parse the subasset information
+#   throws exceptions for invalid asset names
+def parse_subasset_from_description(description):
+    subasset_parent = None
+    subasset_child = None
+    subasset = None
+    chunks = description.rsplit(';;l', 1)
+    if len(chunks) == 2:
+        split_description, asset = chunks
+        subasset_parent, subasset = parse_subasset_from_asset_name(asset)
```
robby-dermody: nit: the naming is a bit confusing, "asset" in the line above, and "subasset" end up being the same string, right? Both are the full asset name, e.g. PIZZA.DOMINOS... ? if so, can you go through and clarify the use of "subasset" vs "asset" in variable names to be a bit more consistent? (I think my nit is more with "subasset = asset" in parse_subasset_from_asset_name() then...)

counterpartylib/lib/util.py
```diff
@@ -198,6 +202,79 @@ def get_asset_name (db, asset_id, block_index):
...
+        if subasset is not None:
+            # if we parsed a subasset, strip ;;l{subasset} from the description
+            description = split_description
+
+    return (subasset_parent, subasset, description)
+
+# throws exceptions for invalid subasset names
+def validate_subasset_name(subasset, subasset_child):
```
robby-dermody: similar nit here... intuitively I think of `subasset` as DOMINOS in PIZZA.DOMINOS, but I see you use `subasset_child` for that some places...which seems a bit redundant. But then again, `asset` technically is the numerical asset name given how we are doing things here, but `longname` and `subasset` still seem to be referring to the same thing in different ways. perhaps we can get more consistency in naming and settle on something like `longname` to mean "PIZZA.DOMINOS", `subasset` to mean "DOMINOS", and get rid of `subasset_child`?

counterpartylib/lib/util.py
```diff
@@ -28,6 +29,9 @@
 D = decimal.Decimal
 b26_digits = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
 
+# subasset contain only characters a-zA-Z0-9.-_@!
+subasset_digits = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.-_@!'
```
robby-dermody: please make upper case as it's a constant (I see that `b26_digits` is lowercase too, that could be made upper case as well)

counterpartylib/lib/util.py
```diff
@@ -198,6 +202,79 @@ def get_asset_name (db, asset_id, block_index):
...
+    return (subasset_parent, subasset, description)
+
+# throws exceptions for invalid subasset names
+def validate_subasset_name(subasset, subasset_child):
+    if len(subasset_child) < 1:
+        raise exceptions.AssetNameError('subasset name too short')
+    if len(subasset) > 250:
+        raise exceptions.AssetNameError('subasset name too long')
+    previous_digit = '.'
+    for c in subasset_child:
```
robby-dermody: alternatively you could do this as `''.join(filter(lambda x: x not in SUBASSET_DIGITS, subasset_child))`and then if that string is not blank, it has invalid characters

counterpartylib/lib/util.py
```diff
@@ -198,6 +202,79 @@ def get_asset_name (db, asset_id, block_index):
...
+        if c not in subasset_digits:
+            raise exceptions.AssetNameError('subasset name contains invalid character:', c)
+        if c == '.' and previous_digit == '.':
```
robby-dermody: likewise, this could be done with `if re.search('\.\.', subasset_child)`

counterpartylib/lib/util.py
deweller: The lambda function is not familiar since I'm a less experience Pythoner.  I kept it simple.  Any problem here?

counterpartylib/lib/util.py
deweller: We need to catch a `.` at the beginning of the at the beginning of the subasset child as well.  As is, it is easy for me to read and understand what is happening.  I did add a comment to explain it.

counterpartylib/lib/util.py
deweller: done

counterpartylib/lib/util.py
deweller: I agree the naming is inconsistent.  I changed everything in util.py to `subasset_longname`, `subasset_parent`, or `subasset_child`.

`subasset_longname` is set to the value of `asset` in this function but only if `asset` is a subasset style name.

counterpartylib/lib/util.py
deweller: `subasset_longname` = PIZZA.DOMINOS
`subasset_parent` = PIZZA
`subasset_child` = DOMINOS

`subasset_child` is only used in 2 validation functions.

counterpartylib/lib/messages/issuance.py
```diff
@@ -124,6 +137,36 @@ def validate (db, source, destination, asset, quantity, divisible, callable_, ca
         if destination:
             problems.append('cannot transfer a non‐existent asset')
 
+    # validate parent ownership for subasset
```
robby-dermody: can you add a block here that's like:
```
if reissued_asset_longname and reissued_asset_longname != subasset_longname:
    problems.append('previous longname does not match longname specified in description')
```

just a good sanity check, I think (unless I'm missing something)

counterpartylib/lib/messages/issuance.py
```diff
@@ -164,7 +210,8 @@ def validate (db, source, destination, asset, quantity, divisible, callable_, ca
     if util.enabled('integer_overflow_fix', block_index=block_index) and (fee > config.MAX_INT or quantity > config.MAX_INT):
         problems.append('integer overflow')
 
-    return call_date, call_price, problems, fee, description, divisible, reissuance
+    return call_date, call_price, problems, fee, description, divisible, reissuance, reissued_asset_longname
```
robby-dermody: i'm not convinced that you need to return `reissued_asset_longname`. if `reissuance == True`, then `subasset_longname == reissued_asset_longname` and you can just use the `subasset_longname` passed in. if `reissuance == False`, then `reissued_asset_longname is None` and you can just use `reissuance`

counterpartylib/lib/messages/issuance.py
```diff
@@ -273,10 +340,17 @@ def parse (db, tx, message):
                 'asset_id': str(asset_id),
                 'asset_name': str(asset),
                 'block_index': tx['block_index'],
+                'asset_longname': subasset_longname,
             }
-            sql='insert into assets values(:asset_id, :asset_name, :block_index)'
+            sql='insert into assets values(:asset_id, :asset_name, :block_index, :asset_longname)'
             issuance_parse_cursor.execute(sql, bindings)
 
+    if reissuance:
```
robby-dermody: ah, this is juicy :D ... two issues with this block:
1. it seems it would crash counterparty-server in the event where an `invalid` error condition was set before `validate()` was called (as `reissuance` wouldn't be defined)
2. I don't even think it's necessary, if `validate()` no longer returns `reissued_asset_longname`, you can just remove it totally and use `'asset_longname': subasset_longname,` in `bindings` below

counterpartylib/lib/blocks.py
```diff
@@ -339,6 +339,14 @@ def initialise(db):
         cursor.execute('''INSERT INTO assets VALUES (?,?,?)''', ('0', 'BTC', None))
         cursor.execute('''INSERT INTO assets VALUES (?,?,?)''', ('1', 'XCP', None))
 
+    # Add asset_longname for sub-assets
```
robby-dermody: following the same logic used for `blocks` (https://github.com/CounterpartyXCP/counterparty-lib/blob/master/counterpartylib/lib/blocks.py#L202), also add `asset_longname` to the `CREATE TABLE IF NOT EXISTS assets(` statement at the end

counterpartylib/lib/blocks.py
```diff
@@ -339,6 +339,14 @@ def initialise(db):
...
+    #   SQLite can’t do `ALTER TABLE IF COLUMN NOT EXISTS`.
+    columns = [column['name'] for column in cursor.execute('''PRAGMA table_info(assets)''')]
+    if 'asset_longname' not in columns:
+        cursor.execute('''ALTER TABLE assets ADD COLUMN asset_longname TEXT''')
+        cursor.execute('''CREATE UNIQUE INDEX IF NOT EXISTS asset_longname_idx ON assets(asset_longname)''')
```
robby-dermody: also, tab this back so that the index is created independently of the column being added (if the index doesn't exist, given `IF NOT EXISTS` in the statement)

counterpartylib/lib/messages/issuance.py
```diff
@@ -186,7 +233,17 @@ def compose (db, source, transfer_destination, asset, quantity, divisible, descr
...
-    call_date, call_price, problems, fee, description, divisible, reissuance = validate(db, source, transfer_destination, asset, quantity, divisible, callable_, call_date, call_price, description, util.CURRENT_BLOCK_INDEX)
+    # check subasset
+    subasset_parent = None
+    subasset_longname = None
+    if util.enabled('subassets'): # Protocol change.
+        subasset_parent, subasset_longname = util.parse_subasset_from_asset_name(asset)
+        if subasset_longname is not None:
+            # generate a random numeric asset id which maps to this subasset
+            asset = util.generate_random_asset()
+            description = description + ';;l' + subasset_longname
```
robby-dermody: I'm on the bench here with making ';;l' as a constant (e.g. `PREFIX_LONGNAME`), and using the constant here instead

counterpartylib/lib/messages/issuance.py
robby-dermody: this would also catch the (infinitesimally small) chance that `util.generate_random_asset()` in `compose` generates a duplicate asset ID already used by the same source address for another asset

counterpartylib/lib/messages/issuance.py
```diff
@@ -124,6 +137,36 @@ def validate (db, source, destination, asset, quantity, divisible, callable_, ca
         if destination:
             problems.append('cannot transfer a non‐existent asset')
 
+    # validate parent ownership for subasset
+    if subasset_longname is not None:
+        cursor = db.cursor()
+        cursor.execute('''SELECT * FROM issuances \
+                          WHERE (status = ? AND asset = ?)
+                          ORDER BY tx_index ASC''', ('valid', subasset_parent))
```
robby-dermody: you are assuming `subasset_parent` is always defined if `subasset_longname` is, but I think that's fine given how you call this function

counterpartylib/lib/messages/issuance.py
```diff
@@ -124,6 +137,36 @@ def validate (db, source, destination, asset, quantity, divisible, callable_, ca
...
+        else:
+            problems.append('parent asset not found')
+
+    # validate subasset issuance is not a duplicate
+    if subasset_longname is not None and not reissuance:
+        cursor = db.cursor()
+        cursor.execute('''SELECT * FROM assets \
+                          WHERE (asset_longname = ?)''', (subasset_longname,))
+        assets = cursor.fetchall()
+        if len(assets) > 0:
```
robby-dermody: you can get rid of the `> 0`

counterpartylib/lib/messages/issuance.py
```diff
@@ -133,7 +176,10 @@ def validate (db, source, destination, asset, quantity, divisible, callable_, ca
             balances = cursor.fetchall()
             cursor.close()
             if util.enabled('numeric_asset_names'):  # Protocol change.
-                if len(asset) >= 13:
+                if subasset_longname is not None and util.enabled('subassets'): # Protocol change.
```
robby-dermody: should be able to get rid of `is not None` as we're not differentiating between `''` and `None`

counterpartylib/lib/messages/issuance.py
robby-dermody: actually, probably keep it as-is. we use magic numbers/letters enough in this file...when they're this short it's a matter of taste, and I'd rather code for readability rather than purism

counterpartylib/lib/util.py
```diff
@@ -198,6 +202,82 @@ def get_asset_name (db, asset_id, block_index):
...
+
+    return (subasset_parent, subasset_longname)
+
+# takes an asset description in 
+#   if it has a suffix like ;;lPARENT.child, then parse the subasset information
+#   throws exceptions for invalid asset names
+def parse_subasset_from_description(description):
+    subasset_parent = None
+    subasset = None
+    chunks = description.rsplit(';;l', 1)
```
robby-dermody: might want to add a comment that this function will get the right-most ';;l' and that additional ';;l' characters are ignored

counterpartylib/lib/util.py
robby-dermody: +1, can you also change `subasset` here to be `subasset_longname`, to be more consistent with `parse_subasset_from_asset_name`?

counterpartylib/lib/util.py
```diff
@@ -198,6 +202,82 @@ def get_asset_name (db, asset_id, block_index):
...
+        if c == '.' and previous_digit == '.':
+            raise exceptions.AssetNameError('subasset name contains consecutive periods')
+        previous_digit = c
+    if previous_digit == '.':
+        raise exceptions.AssetNameError('subasset name ends with a period')
+
+    return True
+
+# throws exceptions for invalid subasset names
+def validate_subasset_parent_name(asset_name):
```
robby-dermody: can you change `asset_name` to `subasset_parent`, for consistency?

counterpartylib/lib/util.py
```diff
@@ -198,6 +202,82 @@ def get_asset_name (db, asset_id, block_index):
...
+        raise exceptions.AssetNameError('parent asset name too long')
+    if asset_name[0] == 'A':
+        raise exceptions.AssetNameError('parent asset name starts with ‘A’')
+    for c in asset_name:
+        if c not in B26_DIGITS:
+            raise exceptions.AssetNameError('parent asset name contains invalid character:', c)
+    return True
+
+def generate_random_asset ():
+    return 'A'+str(random.randint(26**12 + 1, 2**64 - 1))
```
robby-dermody: space between '+'

counterpartylib/protocol_changes.json
```diff
@@ -110,5 +110,12 @@
         "minimum_version_revision": 0,
         "block_index": 441563,
         "testnet_block_index": 0
+    },
+    "subassets": {
+        "minimum_version_major": 9,
+        "minimum_version_minor": 55,
+        "minimum_version_revision": 2,
+        "block_index": 455000,
+        "testnet_block_index": 1088080
```
robby-dermody: any reason you are using `1088080` for `testnet_block_index`?

counterpartylib/test/conftest.py
```diff
@@ -42,6 +55,15 @@ def enabled(change_name, block_index=None):
         return _enabled(change_name, block_index)
 util.enabled = enabled
 
+RANDOM_ASSET_INT = None
+_generate_random_asset = util.generate_random_asset
+def generate_random_asset ():
+    if RANDOM_ASSET_INT is None:
+        return _generate_random_asset()
+    else:
+        return "A"+str(RANDOM_ASSET_INT)
```
robby-dermody: space around `+`

counterpartylib/protocol_changes.json
deweller: This was completely arbitrary for testing purposes.  It was the approximate current block at the time I was running this code on testnet.

counterpartylib/test/fixtures/vectors.py
```diff
@@ -1441,75 +1441,95 @@
...
+        }, {
+            'comment': 'A subasset name must be unique',
+            'in': (ADDR[0], None, 'A{}'.format(26**12 + 1), 100000000, True, False, None, None, 'description', 'PARENT', 'PARENT.already.issued', DP['default_block_index']),
+            'out': (0, 0.0, ['subasset already exists'], 25000000, 'description', True, False, None)
+        }, {
+            'comment': 'cannot change subasset name through a reissuance description modification',
+            'in': (ADDR[0], None, 'A{}'.format(26**12 + 101), 200000000, True, False, None, None, 'description', 'PARENT', 'PARENT.changed.name', DP['default_block_index']),
+            'out': (0, 0.0, [], 0, 'description', True, True, 'PARENT.already.issued')
+        }, {
+            'in': (ADDR[0], None, 'UNRELATED', 1000, True, False, None, None, 'description', 'PARENT', 'PARENT.child1', DP['default_block_index']),
```
robby-dermody: I think this all is a good first take, but @rubensayshi may have some additional tests he can recommend here

counterpartylib/protocol_changes.json
robby-dermody: fair enough... so set it to '0' now then for this PR?

counterpartylib/lib/messages/issuance.py
deweller: If a reissuance specifies a different subasset in the description, then the original subasset wins - that is the purpose of the reissued_asset_longname variable.  I wrote a test to cover this scenario (https://github.com/deweller/counterparty-lib/blob/cip4-wip-2/counterpartylib/test/fixtures/vectors.py#L1527-L1529).

I can add a sanity check here in case someone accidentally appends the wrong subasset into the description of a reissuance.  But that would make the reissuance transaction invalid.

So - the question is should a reissuance with a valid numeric asset and invalid subasset in the description be valid or invalid?

Scenario:
- I own PARENTASSET.mychild (as A100001)
- I issue 100 additional A100001, but I include `;;lPARENTASSET.oopsie` in the description

Should the reissuance of 100 A100001 be a valid or invalid issuance?  

It is ok to let it be valid.  The code will correctly insert PARENTASSET.mychild into the issuances table as the asset_longname.

I lean toward allowing the transaction and ignoring the bad description.  But I'm fine either way.  I'm not worried about the numeric asset collision.

counterpartylib/lib/messages/issuance.py
deweller: If we change the functionality in the above, then we can get rid of reissued_asset_longname.

counterpartylib/lib/messages/issuance.py
deweller: 1. Good find.  Changed to `if status == 'valid' and reissuance:` 
2. This is also related to the above discussion.  We can remove it if we also get rid of `reissued_asset_longname`.

counterpartylib/lib/blocks.py
deweller: Will do.

counterpartylib/lib/blocks.py
deweller: will do

deweller: I also had to move and modify the BTC and XCP asset initialization code after the additional column.

```python
    cursor.execute('''SELECT * FROM assets WHERE asset_name = ?''', ('BTC',))
    if not list(cursor):
        cursor.execute('''INSERT INTO assets VALUES (?,?,?,?)''', ('0', 'BTC', None, None))
        cursor.execute('''INSERT INTO assets VALUES (?,?,?,?)''', ('1', 'XCP', None, None))
```

counterpartylib/lib/messages/issuance.py
deweller: Was attempting to follow pep-8 here (not that my other code is perfectly pep-8 compiant...)

counterpartylib/lib/util.py
deweller: yep

counterpartylib/lib/util.py
deweller: ✔

counterpartylib/test/conftest.py
deweller: ✔

counterpartylib/lib/messages/issuance.py
robby-dermody: hmm, ok. When looking at PEP8, I see the following:

```Also, beware of writing if x when you really mean if x is not None -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!```

Perhaps that is what you are referring to? That's fine I think then, if there is absolutely no way that `subasset_longname` could be a blank string

File: counterpartylib/lib/messages/issuance.py (hunk 9 above)
Comment (robby-dermody) [inline counterpartylib/lib/messages/issuance.py]: I would prefer to reject it to remove the ambiguity (in both the code and the spec) as well as simplify the code. The only place where it might be ok is if the longname segment (e.g. `;;lPARENTASSET.mychild`) was totally missing from the description, and in that case I'd still probably want to reject it (but it *might* work to just have counterparty-lib deal with it in that case).

Really, the only place being more strict like this would affect would be with the 3rd party library writers (e.g the JS counterparty library, etc), as anyone else will be using the counterparty-lib API, and they will just specify "asset" as the full longname, and description to whatever, and not even know about `;;l`. I have no problem making library authors jump through an additional (minor) hoop if it means more clear cut logic in both the spec and implementation. :)

File: counterpartylib/lib/messages/issuance.py (hunk 17 above)
Comment (deweller) [inline counterpartylib/lib/messages/issuance.py]: I doublechecked.  subasset_longname is always either a valid subasset name or None - never an empty string.

File: counterpartylib/protocol_changes.json (hunk 21 above)
Comment (deweller) [inline counterpartylib/protocol_changes.json]: Should we pick a block_index for for livenet too?  Guess we'll need to wait a bit on that.

Comment (robby-dermody) [inline counterpartylib/protocol_changes.json]: yeah, I'd wait and do that right before release.

also, can you add to the changelog for this new version?

File: counterpartylib/lib/messages/issuance.py (hunk 17 above)
Comment (robby-dermody) [inline counterpartylib/lib/messages/issuance.py]: sounds good. can leave as-is then

File: counterpartylib/lib/messages/issuance.py (hunk 9 above)
Comment (deweller) [inline counterpartylib/lib/messages/issuance.py]: In my opinion, a reissuance description should **not** contain the `;;lPARENT.child`.  There is no need to waste the additional bytes in the blockchain.

Related: `compose` will fail if the child asset is transferred and the new owner tries to compose an issuance specifying `PARENT.child` as the asset name.  Hmmm...  I think this should be fixed.

Are you ok with leaving out `;;lPARENT.child` from reissuance descriptions?  If so, then I should modify `compose` to handle a reissuance of a transferred child asset when with an asset name like `PARENT.child`.

Comment (robby-dermody) [inline counterpartylib/lib/messages/issuance.py]: sure, I'm fine leaving it out on reissuance descriptions if we can be strict about it and say that ;;l exists on the initial issuance to tie that longname to the numerical asset, but reissuances must not have it (and fail if they do). That will also fail if ;;l is specified and does not match the initial longname as well, of course.

Can you also open up a PR to CIP004 that fills in that spec with these clarifications? (that is, given the decisions we are making here, the CIP is lacking necessary detail in certain areas)

File: counterpartylib/lib/util.py [hunk 24]
```diff
@@ -198,6 +202,82 @@ def get_asset_name (db, asset_id, block_index):
...
+            # if we parsed a subasset, strip ;;l{subasset} from the description
+            description = split_description
+
+    return (subasset_parent, subasset_longname, description)
+
+# throws exceptions for invalid subasset names
+def validate_subasset_longname(subasset_longname, subasset_child):
+    if len(subasset_child) < 1:
+        raise exceptions.AssetNameError('subasset name too short')
+    if len(subasset_longname) > 250:
```
robby-dermody: there was a debate recently on this, being 250 vs 40. what are your views?

I'm on the bench. I can't think of a usecase for 250 right now, beyond spam assets that cause a GUI headache.

deweller: I think a 40 character limit is the most practical, but since the CIP was discussed and a bounty raised, I think we should leave it.  My opinion is not strong.  I think the anti-spam fee is big enough to prevent harmful abuse and wallet makers can just ignore long assets.

counterpartylib/lib/config.py
rubensayshi: I think a protocol change should require a version bump but I never really thought about the auto-triggered reparse being unwanted before

counterpartylib/lib/messages/issuance.py
deweller: The CIP is updated and this is implemented in the latest commits on this PR.

counterpartylib/lib/messages/issuance.py
deweller: This is still necessary.  When a reissuance transaction is parsed, the subasset_longname is **not** passed to validate from `parse`.  Validate looks up the asset_longname and returns it.

It does mix concerns a little bit, but since validate is already looking up the asset, I returned it to avoid duplicate SQL queries.

It could be refactored to be cleaner, but I don't think it is worth it at this time.

counterpartylib/protocol_changes.json
deweller: Yep.  Added to the changelog.

counterpartylib/test/conftest.py
rubensayshi: and no space between `def generate_random_asset` and `()`

counterpartylib/lib/messages/issuance.py
```diff
@@ -306,6 +318,13 @@ def parse (db, tx, message):
     fee = 0
     if status == 'valid':
         call_date, call_price, problems, fee, description, divisible, reissuance, reissued_asset_longname = validate(db, tx['source'], tx['destination'], asset, quantity, divisible, callable_, call_date, call_price, description, subasset_parent, subasset_longname, block_index=tx['block_index'])
+
+        # extra validation for parsing only.  
+        # A reissuance description that contains the string ;;l in the description is invalid
+        if not problems and reissuance and ';;l' in unmodified_description and util.enabled('subassets', block_index=tx['block_index']):
```
robby-dermody: any reason for `not problems`? ...  `len(problems)` can be > 1 of course..

counterpartylib/lib/messages/issuance.py
```diff
@@ -306,6 +318,13 @@ def parse (db, tx, message):
...
+            problems.append('reissuance description cannot contain ;;l')
```
robby-dermody: `;` is used to separate multiple problems in the resultant status string (as a few lines below this one shows), of course. ...I'm concerned that some programs may be parsing out multiple problems by splitting at `;`... maybe instead of `'reissuance description cannot contain ;;l'` do something like `'reissuance description cannot re-specify longname key-value pair'` or something similar

counterpartylib/lib/messages/issuance.py
```diff
@@ -246,37 +249,65 @@ def compose (db, source, transfer_destination, asset, quantity, divisible, descr
...
     issuance_parse_cursor = db.cursor()
 
     # Unpack message.
     try:
-        if (tx['block_index'] > 283271 or config.TESTNET) and len(message) >= LENGTH_2: # Protocol change.
+        subasset_longname = None
+        if message_type_id == SUBASSET_ID and util.enabled('subassets', block_index=tx['block_index']):
+            # parse a subasset original issuance message
+            asset_id, quantity, divisible, compacted_subasset_length = struct.unpack(SUBASSET_FORMAT, message[0:SUBASSET_FORMAT_LENGTH])
+            description_length = len(message) - SUBASSET_FORMAT_LENGTH - compacted_subasset_length
```
robby-dermody: hmm maybe I'm missing it, but couldn't someone make a custom message that had 0 for `compacted_subasset_length`? In that case, the decode would be fine (`0s` would be specified in the struct unpack, which would give back `b''`), but `compacted_subasset_longname` would be `b''` and `subasset_longname` would be `''`, and line 341 below would essentially return None for both `subasset_parent` and `subasset_longname`... so in effect you may get a traditional numeric asset issuance with this new call maybe? Probably best to just put an explicit check that `compacted_subasset_length != 0`

counterpartylib/lib/messages/issuance.py
```diff
@@ -246,37 +249,65 @@ def compose (db, source, transfer_destination, asset, quantity, divisible, descr
...
+            if description_length < 0:
```
robby-dermody: good check 👍

robby-dermody:
looks good. minus:

* updated CIP4
* updated API docs
* maybe some additional unit tests
* Ruben checking it over

counterpartylib/lib/messages/issuance.py
deweller: Good catch.  A length of zero was getting through this check and issuing a numeric asset.

I added an explicit validation call in the parse function to prevent this.

I also added unit tests to cover this situation.

deweller:
Added 2 more unit tests for `issuance.parse`

counterpartylib/lib/messages/issuance.py
droplister: ```> 0``` is not necessary, but it's readable. ```if len(assets):``` would return 0/false if length was 0, but IMO good as-is.

counterpartylib/lib/blocks.py
```diff
@@ -93,8 +93,8 @@ def parse_tx(db, tx):
         order.parse(db, tx, message)
     elif message_type_id == btcpay.ID:
         btcpay.parse(db, tx, message)
-    elif message_type_id == issuance.ID:
-        issuance.parse(db, tx, message)
+    elif message_type_id == issuance.ID or message_type_id == issuance.SUBASSET_ID:
```
droplister: Looking at this, it's the only ```or``` logic in this group of else/if statements. It might be better as just another```elif```.

Really just a comment re: consistency.

counterpartylib/lib/blocks.py
```diff
@@ -326,18 +326,27 @@ def initialise(db):
...
     cursor.execute('''CREATE INDEX IF NOT EXISTS
                       id_idx ON assets (asset_id)
                    ''')
+
+    # Add asset_longname for sub-assets
+    #   SQLite can’t do `ALTER TABLE IF COLUMN NOT EXISTS`.
+    columns = [column['name'] for column in cursor.execute('''PRAGMA table_info(assets)''')]
+    if 'asset_longname' not in columns:
+        cursor.execute('''ALTER TABLE assets ADD COLUMN asset_longname TEXT''')
+    cursor.execute('''CREATE UNIQUE INDEX IF NOT EXISTS asset_longname_idx ON assets(asset_longname)''')
```
droplister: So, if this is the first installation. The table gets created with an asset_longname column, but if this is an upgrade and the table does not contain an asset_longname column, add it. Got it.

My comment would be around failures to add the column. I suppose at around 60-70k rows it's not that much of an issue, but I'm just going through and calling out what I see. Don't take my review too seriously.

counterpartylib/lib/messages/issuance.py
droplister: I agree with what you guys settled on here. It makes sense.

counterpartylib/lib/messages/issuance.py
droplister: In isolation, but knowing how the issuance of subassets works, it's safe to assume a subasset has a parent. Right?

counterpartylib/lib/messages/issuance.py
droplister: My comment here would be in regards to this check: ``` if asset[0] != 'A': ```. Just to point out that if, in the future the 'A' namespace is opened up, this logic would fail.

counterpartylib/lib/blocks.py
deweller: Yeah, I originally wrote it with 2 `elif`s.  But then I changed it to stand out and be obvious that both of these messages are routed to the same function.  Just a preference.

counterpartylib/lib/messages/issuance.py
deweller: Yes.  If `subasset_longname` is not None, then `subasset_parent` will not be None.

counterpartylib/lib/util.py
```diff
@@ -198,6 +207,89 @@ def get_asset_name (db, asset_id, block_index):
...
+    if integer == 0:
+        return ''
+    ret = ''
+    while integer != 0:
+        ret = SUBASSET_DIGITS[integer % 68 - 1] + ret
+        integer //= 68
+    return ret
+
+def generate_random_asset ():
+    return 'A' + str(random.randint(26**12 + 1, 2**64 - 1))
```
droplister: I'm just a simpleton, but is this basically assuming that the random string is always available? Is there any significant performance hit by checking? Will a collision result in a halting error across all nodes or is this something only a user would be inconvenienced by?

counterpartylib/lib/blocks.py
droplister: Right, I got why you did it and I think you justified it here fine.

counterpartylib/lib/util.py
deweller: It would be an inconvenience (the asset issuance would be invalid when it confirms).

The odds of a collision are roughly 1 in 100 million if there are 600k numeric assets in existence.  I chose not to worry about it.

droplister: Got it. I agree that it's not worth worrying about.

deweller:
@robby-dermody and @rubensayshi - I'm done with this.

robby-dermody:
@rubensayshi please review when you get a chance, then I will do a final review and we can pick a final block number, and merge after discussing what other items we'd want in this hardfork release (the main one is a fee on numeric assets)

rubensayshi:
nice idea to make the subasset name more compact to save a few bytes, we should keep that in mind for future things as well.  
should maybe be documented a bit in the CIP though ;-) and/or maybe a small docblock with the `util.compact_subasset_longname` to explain what is going on.

deweller:
It is (just barely) documented in the CIP where it says `encode the subasset longname in a base68 compressed format`.

counterpartylib/lib/messages/issuance.py
```diff
@@ -186,29 +236,78 @@ def compose (db, source, transfer_destination, asset, quantity, divisible, descr
...
 
-def parse (db, tx, message):
+def parse (db, tx, message, message_type_id):
     issuance_parse_cursor = db.cursor()
 
     # Unpack message.
     try:
-        if (tx['block_index'] > 283271 or config.TESTNET) and len(message) >= LENGTH_2: # Protocol change.
+        subasset_longname = None
+        if message_type_id == SUBASSET_ID and util.enabled('subassets', block_index=tx['block_index']):
```
rubensayshi: when `SUBASSET_ID` and subassets isn't enabled yet we shouldn't let it run through the `elif` and `else` that follow

deweller: Agreed.  I added this check.

rubensayshi:
ACK https://github.com/CounterpartyXCP/counterparty-lib/pull/961/commits/716594c471c80df8869489c80f51a9ecf16f7e9a

deweller:
I gave this a final sanity check.

ACK 716594c

robby-dermody:
ACK. Merging shortly. Beyond this:
* I will add Counterwallet support for subassets later this week.
* We will work to coordinate a release, as well as release announcements from press, etc.

Let's discuss what else we'd like in this release as well, on slack.

PR #964, opened 2017-04-25
add support for .1 XCP fee on numeric assets

robby-dermody:
I had to modify `CALLPRICEFLOAT` and `CALLDATEFLOAT` because their original names were >= 13 characters and raised an error on testsuite run (this was a latent bug the fee change uncovered)

### DIFF
--- a/counterpartylib/test/fixtures/vectors.py
+++ b/counterpartylib/test/fixtures/vectors.py
@@ -1472,8 +1472,8 @@ UNITTEST_VECTOR = {
             'in': (ADDR[0], None, 'NOSATOSHI', 1000.5, True, False, None, None, '', None, None, DP['default_block_index']),
             'out': (0, 0.0, ['quantity must be in satoshis'], 0, '', True, None, None)
         }, {
-            'in': (ADDR[0], None, 'CALLPRICEFLOAT', 1000, True, False, None, 100.0, '', None, None, DP['default_block_index']),
-            'out': (0, 0.0, [], 0, '', True, False, None)
+            'in': (ADDR[0], None, 'CALLPRICEFLT', 1000, True, False, None, 100.0, '', None, None, DP['default_block_index']), # CALLPRICE AS FLOAT
+            'out': (0, 0.0, [], 50000000, '', True, False, None)
         }, {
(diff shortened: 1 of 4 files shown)


counterpartylib/test/fixtures/vectors.py
```diff
@@ -1472,8 +1472,8 @@
             'in': (ADDR[0], None, 'NOSATOSHI', 1000.5, True, False, None, None, '', None, None, DP['default_block_index']),
             'out': (0, 0.0, ['quantity must be in satoshis'], 0, '', True, None, None)
         }, {
-            'in': (ADDR[0], None, 'CALLPRICEFLOAT', 1000, True, False, None, 100.0, '', None, None, DP['default_block_index']),
-            'out': (0, 0.0, [], 0, '', True, False, None)
+            'in': (ADDR[0], None, 'CALLPRICEFLT', 1000, True, False, None, 100.0, '', None, None, DP['default_block_index']), # CALLPRICE AS FLOAT
+            'out': (0, 0.0, [], 50000000, '', True, False, None)
```
deweller: I wonder how these didn't trigger errors in the test suite until now.  I guess there is no validation on the length in this part of the code.

deweller:
We should write a CIP for the protocol change and include some research to determine the appropriate fee levels here.

robby-dermody:
Will keep this item open but hold off for now.

jdogresorg:
We are now seeing some of the behavior I warned about... numerics being registered as throwaway receipts which is filling up the CP database with wasteful data. So far only a few hundred in the past week... but this will continue to increase as more users use the system.

I am still aware of 3-4 projects that plan on spamming numeric asset registrations in the next few weeks/months.

I would like to bring this conversation up for discussion once again... The majority of the community seems to be for it, the last CP foundation voted to institute this fee, and I believe this should be activated before we have an issue.

deweller:
This would require a consensus change to the protocol. The way to begin this process is to propose a CIP and open a discussion in the forums.

jdogresorg:
Heh... yep, this is a consensus change, hence why we wanted to push it out with the last consensus change. :P

If you require a CIP to move this forward, a CIP will be forthcoming :)

chiguireitor:
As there's no consensus change regarding this, i'm gonna close for now.

Let's keep the discussion at counterpartytalk and revisit it later

PR #965, opened 2017-04-28
9.55.2

robby-dermody:
@rubensayshi @deweller please check ASAP.


setup.py
```diff
@@ -213,7 +213,7 @@ def run(self):
     'pysha3==0.3',
     'pytest==2.9.2',
     'pytest-cov==2.2.1',
-    'python-bitcoinlib==0.7.0',
+    'python-bitcoinlib', #==0.7.0
```
deweller: By removing a specific version constraint, does this mean that the python-bitcoinlib version for different installs of counterparty based on the time that the installation was performed?

If so, we should lock this to 0.7.0 until we can test a newer version.

robby-dermody: with new installs it will use the newest version on pypi (0.7.0). with existing installs it won't enforce a particular version. I tried a few other version strings like `'python-bitcoinlib==0.7.0, ==0.7.1-SNAPSHOT',` and `'python-bitcoinlib>=0.7.0',` with no luck, but let me try again with this new code as I agree this needs to change

counterpartylib/protocol_changes.json
```diff
@@ -115,7 +115,7 @@
         "minimum_version_major": 9,
         "minimum_version_minor": 55,
         "minimum_version_revision": 2,
-        "block_index": 999999,
+        "block_index": 467417,
```
deweller: This is May 22nd, around 11 AM ET.  Give or take a few hours of course.

robby-dermody: yep. that's the goal -- 3 weeks from monday may 1 at 11am. sound good?

setup.py
```diff
@@ -208,12 +208,12 @@ def run(self):
...
     'json-rpc==1.10.3',
-    'pycoin==0.62',
+    'pycoin==0.77',
     'pycrypto==2.6.1',
     'pysha3==0.3',
-    'pytest==2.9.1',
+    'pytest==2.9.2',
     'pytest-cov==2.2.1',
-    'python-bitcoinlib==0.5.1',
+    'python-bitcoinlib', #==0.7.0
```
deweller: I think we should lock this to 0.7.0 so no one accidentally builds counterparty server with an unknown future version of python-bitcoinlib.

deweller:
ACK - pending the resolution of the python-bitcoinlib version.

counterpartylib/protocol_changes.json
deweller: Yep.

setup.py
rubensayshi: but when 0.7.1 is released it will install the latest version

robby-dermody: Yeah, I would have preferred "bitcoinlib==0.7.0, ==0.7.1-SNAPSHOT", but pip wasn't liking that. :/ any ideas are welcome.

I don't entirely see it as a bad thing, as I doubt there will be parsing differences between the teo, and TBH, by the time Peter gets around to releasing 0.7.1, we will have an upcoming release of CIP9/10 can can update it to ==0.7.1

robby-dermody: Actually, let me try `python-bitcoinlib>=0.7.0,<0.7.1` ...

robby-dermody: ok that worked 👍

deweller: Ahh - good.  We'll still need to doublecheck if a minor version update like 0.7.1 is released.  But that should be only a bugfix release anyway.

PR #966, opened 2017-05-01
necessary 9.55.2 hotfixes

robby-dermody:
* Fixes for the whole python-bitcoinlib issue (fixed by cloning the repo, and setting the version to 0.7.0)
* Fixes setuptools missing package issue (python-markdown) that happened on my boxes
* Fixes reparse issue on testnet

Everything seems to be smooth now with these fixes.


deweller:
ACK

PR #980, opened 2017-07-01
Move reparse working database out of memory onto disk

deweller:
This should work around the CircleCI memory exhaustion errors


cryptcoin-junkey:
I'm also trying to resolve CI failures.
And I agree the patch in this PR is effective for avoiding memory exhaust.
But we'll get test failures by timeout.

It may be hard to run reparse tests on public (free) CI services.

deweller:
Closing this as it didn't work.  It lowered memory usage, but the tests took too long to run.

PR #985, opened 2017-07-09
Implements CIP 9 (WIP)

deweller:
See https://github.com/CounterpartyXCP/cips/blob/master/cip-0009.md

Based on CIP 11.


counterpartylib/lib/api.py
```diff
@@ -189,8 +190,15 @@ def value_to_marker(value):
...
             raise APIError("case_sensitive must be a boolean")
 
+    # special case for memo and memo_hex field searches
+    if table == 'sends':
+        adjust_get_sends_memo_filters(filters)
+
     # SELECT
-    statement = '''SELECT * FROM {}'''.format(table)
+    if table == 'sends':
+        statement = '''SELECT *, CASE WHEN memo IS NULL THEN NULL ELSE hex(memo) END AS memo_hex FROM {}'''.format(table)
```
rubensayshi: could this be moved to `adjust_get_sends_results` instead?

counterpartylib/lib/messages/send.py
```diff
@@ -30,13 +32,29 @@ def initialise (db):
...
 def validate (db, source, destination, asset, quantity, block_index):
     return send1.validate(db, source, destination, asset, quantity, block_index)
 
-def compose (db, source, destination, asset, quantity):
+def compose (db, source, destination, asset, quantity, memo=None, memo_is_hex=False, use_enhanced_send=None):
+    # special case - enhanced_send replaces send by default when it is enabled
+    #   but it can be explicitly disabled with an API parameter
+    if util.enabled('enhanced_sends'):
+        if use_enhanced_send is None or use_enhanced_send == True:
+            return enhanced_send.compose(db, source, destination, asset, quantity, memo, memo_is_hex)
```
rubensayshi: I think it should raise an exception when `use_enhanced_send=True` before `enhanced_sends` is enabled?

counterpartylib/lib/messages/versions/enhanced_send.py
```diff
@@ -0,0 +1,202 @@
...
+MAX_MEMO_LENGTH = 34
+ID = 2 # 0x02
+
+def unpack(db, message, block_index):
+    try:
+        # account for memo bytes
+        memo_bytes_length = len(message) - LENGTH
+        if memo_bytes_length < 0:
+            raise exceptions.UnpackError('invalid message length')
+        if memo_bytes_length > MAX_MEMO_LENGTH:
```
rubensayshi: this check is also in `validate`, do we also need it here?

counterpartylib/lib/messages/versions/enhanced_send.py
```diff
@@ -0,0 +1,202 @@
...
+
+    if not isinstance(quantity, int):
+        problems.append('quantity must be in satoshis')
+        return problems
+
+    if quantity < 0:
+        problems.append('negative quantity')
+
+    # if quantity == 0:
+    #     problems.append('zero quantity')
```
rubensayshi: I guess the old `send` allowed to send 0 of an asset... weird ...  
I think it makes sense to make it require > 0 so uncomment this (and add a test to match the failure case)

counterpartylib/lib/messages/versions/enhanced_send.py
```diff
@@ -0,0 +1,202 @@
...
+def parse (db, tx, message):
+    cursor = db.cursor()
+
+    # Unpack message.
+    try:
+        unpacked = unpack(db, message, tx['block_index'])
+        asset       = unpacked['asset']
+        quantity    = unpacked['quantity']
+        destination = unpacked['address']
+        memo_bytes  = unpacked['memo']
```
rubensayshi: I dont think this is pythonesque nor do we ever do this aligning of `=` signs.  
could just do `asset, quantity, destination, memo_bytes = unpacked['asset'], unpacked['quantity'], unpacked['address'], unpacked['memo']`

counterpartylib/lib/messages/versions/enhanced_send.py
```diff
@@ -0,0 +1,202 @@
...
+                                     WHERE (address = ? AND asset = ?)''', (tx['source'], asset))
+        balances = cursor.fetchall()
+        if not balances:
+            status = 'invalid: insufficient funds'
+        elif balances[0]['quantity'] < quantity:
+            quantity = min(balances[0]['quantity'], quantity)
+
+    # For SQLite3
+    if quantity:
+        quantity = min(quantity, config.MAX_INT)
```
rubensayshi: tbh I wouldn't be opposed to taking the chance to breaking this "auto correcting" behaviour for enh sends ... (both for when quantity > MAX-INT and when quantity < balance), lets discuss on slack a bit maybe?

counterpartylib/lib/messages/versions/enhanced_send.py
```diff
@@ -0,0 +1,202 @@
...
+            raise exceptions.UnpackError('memo too long')
+
+        struct_format = FORMAT + ('{}s'.format(memo_bytes_length))
+        asset_id, quantity, short_address_bytes, memo_bytes = struct.unpack(struct_format, message)
+        if len(memo_bytes) == 0:
+            memo_bytes = None
+        
+        # unpack address
+        try:
+          full_address = address.unpack(short_address_bytes)
```
rubensayshi: needs a failure test case, also 4 spaces instead of 2

counterpartylib/lib/messages/versions/enhanced_send.py
```diff
@@ -0,0 +1,202 @@
...
+        except:
+          raise exceptions.UnpackError('address invalid')
+
+        # asset id to name
+        asset = util.generate_asset_name(asset_id, block_index)
+        if asset == config.BTC:
+          raise exceptions.AssetNameError('{} not allowed'.format(config.BTC))
```
rubensayshi: 4 spaces

counterpartylib/lib/messages/versions/enhanced_send.py
```diff
@@ -0,0 +1,202 @@
...
+    if status == 'valid':
+        util.debit(db, tx['source'], asset, quantity, action='send', event=tx['tx_hash'])
+        util.credit(db, destination, asset, quantity, action='send', event=tx['tx_hash'])
+
+    # log invalid transactions
+    if status != 'valid':
+        if quantity is None:
+            logger.warn("Invalid send from %s with status %s. (%s)" % (tx['source'], status, tx['tx_hash']))
+        else:
+            logger.warn("Invalid send of %s %s from %s to %s. status is %s. (%s)" % (quantity, asset, tx['source'], destination, status, tx['tx_hash']))
```
rubensayshi: needs test case to hit this

counterpartylib/lib/messages/versions/enhanced_send.py
deweller: Got it.  Still getting used to some of the python idioms.

counterpartylib/lib/messages/versions/enhanced_send.py
deweller: I think it is possible this method will get called without validate.  We should leave it.

counterpartylib/lib/messages/versions/enhanced_send.py
deweller: Turns out there is no possible failure case.  Any 21 bytes are valid and `short_address_bytes` must be 21 bytes long if we got here.  I removed this check.

counterpartylib/lib/api.py
deweller: Yes.  Good idea.

counterpartylib/lib/messages/send.py
deweller: I added this.  It also is triggered if memo is not none.

counterpartylib/lib/messages/versions/enhanced_send.py
deweller: I went ahead and added this.  There is no reason I can think of to send zero quantity with an enhanced send.

counterpartylib/lib/messages/versions/enhanced_send.py
deweller: Send quantities are no longer auto-corrected.
Send quantities greater than MAX_INT are now invalid.
Send quantities greater than the amount the user holds are now invalid.

counterpartylib/lib/messages/versions/enhanced_send.py
deweller: Both cases are covered now with the changes above.

rubensayshi:
need to fix circle, but ACK otherwise

deweller:
Rebased off of development which now includes CIP-11

PR #989, opened 2017-08-25
Fixes CIP-9 tests

deweller:


deweller:
Merging this so CIP-12 can rebase.

PR #991, opened 2017-08-30
Support SegWit

cryptcoin-junkey:
As SegWit was activated.


deweller:
Please see https://github.com/CounterpartyXCP/counterparty-lib/pull/1001

PR #996, opened 2017-09-15
Fix locked issuance bug

deweller:
If the description was changed after the asset was locked, further issuances were not prevented.


deweller:
I need to rebase after CIP-12 is merged to develop.  I also need to put a protocol change in for this.

deweller:
replaced by https://github.com/CounterpartyXCP/counterparty-lib/pull/999.

PR #998, opened 2017-09-16
Remove rpcthreads=100

unsystemizer:
This was never really proven to help, and by now it's almost 100% unnecessary and may even be harmful.

I added some other clarifications to the file.


deweller:
ack f983769

rubensayshi:
@deweller @unsystemizer has anyone been running without this sucesfully?

deweller:
Looks like federated node uses `rpcthreads=50`.  Perhaps we should just change the readme to match that.

deweller:
I've been running with the rcpthreads setting removed for a couple days now with no issues.

rubensayshi:
needs rebase but ACK for this commit: https://github.com/CounterpartyXCP/counterparty-lib/pull/998/commits/f983769f1a7affa7b3895c068cb3160e097bf6ae

PR #999, opened 2017-09-18
Fix locked issuance

deweller:
If the description was changed after the asset was locked, further quantity issuances were not prevented.


deweller:
ack 285bf6ae3829e587989983f2f7615e369abd2be3

rubensayshi:
ACK https://github.com/CounterpartyXCP/counterparty-lib/commit/285bf6ae3829e587989983f2f7615e369abd2be3

PR #1000, opened 2017-09-20
use python-bitcoinlib 0.8.0 for segwit support

rubensayshi:


deweller:
ack 5d1fd2c

cryptcoin-junkey:
Have you tested it? AFAIK, python-bitcoinlib 0.8.0 doesn't have SegWit support. See the fork tree on python-bitcoinlib.

screenshot 2017-09-21 at 16

At least 0.8.0 doesn't work on my machine. So I sent PR #991.

deweller:
@cryptcoin-junkey - You said 0.8.0 doesn't work on your machine.  Can you elaborate?  What happened?

deweller:
You are correct that this doesn't actually fix the segwit serialization problem.

Please see https://github.com/CounterpartyXCP/counterparty-lib/pull/1001

PR #1001, opened 2017-09-22
Use python-bitcoinlib snapshot for segwit support

deweller:
Since segwit support is not merged into the 0.8.0 release, we need to use a snapshot from the master branch.

After much trial and error, I found this way to be the cleanest way to handle this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ MAINTAINER Counterparty Developers <dev@counterparty.io>
 # Install counterparty-lib
 COPY . /counterparty-lib
 WORKDIR /counterparty-lib
-RUN pip3 install -r requirements.txt
+RUN pip3 install -r requirements.txt --process-dependency-links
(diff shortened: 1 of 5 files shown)


cryptcoin-junkey:
It looks better than my PR #991 even though I've not tested yet.

deweller:
We'd prefer to use a tagged release of https://github.com/petertodd/python-bitcoinlib.

But if there is not one available soon, we will go with this.

unsystemizer:
For reference, here's why the SW fix wasn't included in v0.8.0:
https://github.com/petertodd/python-bitcoinlib/issues/146#issuecomment-326463592

rubensayshi:
needs rebase, ACK with the 2 line code change to use `+git+https://github.com/petertodd/python-bitcoinlib.git@98676f981bf14a6a3a8313e762161cc289043b58#egg=python-bitcoinlib-0.8.1`

PR #1005, opened 2017-10-10
API extension to support Counterwallet

deweller:
Adds an `extended_tx_info` parameter to the `create_` methods

If this parameter is present, instead of returning a bare unsigned transaction hex, the API will return a JSON object with details about btc spent.  The output looks like this:

```json
{
  "btc_fee": 28621,
  "tx_hex": "0100000001688469...088ac00000000",
  "btc_in": 7683700,
  "btc_out": 0,
  "btc_change": 7655079
}
```

Also removes forced pubkeyhash downgrade and fixes OP_RETURN size estimation.

All of these changes are backward-compatible API changes.  There are no changes to consensus code.


deweller:
Fixes https://github.com/CounterpartyXCP/counterwallet/issues/810

Issue #992, opened 2017-09-03
update get_asset_names to support asset_longname


jdogresorg:
Currently get_asset_names() returns subassets as the numeric asset name.

It would be nice to see an option to return both the asset and asset_longname in the response.

droplister:
I started a PR with a possible way to approach this. Probably needs some work still, but might get it going.

PR #1007, opened 2017-10-25
Add longnames flag to get_asset_names

droplister:
I believe a change like this would address: https://github.com/CounterpartyXCP/counterparty-lib/issues/992

### DIFF
--- a/counterpartylib/lib/api.py
+++ b/counterpartylib/lib/api.py
@@ -721,7 +721,7 @@ class APIServer(threading.Thread):
             cursor = db.cursor()
             if longnames:
                 names = {}
-                for row in cursor.execute("SELECT asset, asset_longname FROM issuances WHERE status = 'valid' GROUP BY asset ORDER BY asset ASC")
+                for row in cursor.execute("SELECT asset, asset_longname FROM issuances WHERE status = 'valid' GROUP BY asset ORDER BY asset ASC"):
                     names[row['asset']] = row['asset_longname']
             else:
                 names = [row['asset'] for row in cursor.execute("SELECT DISTINCT asset FROM issuances WHERE status = 'valid' ORDER BY asset ASC")]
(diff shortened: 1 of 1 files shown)


droplister:
If this does what I expect it to, if you set the longnames parameter to True, you would get a dictionary where the asset was the key and the asset_longname was the value.

It could also be achieved by just creating a separate get_asset_longnames method.

deweller:
I think I would prefer a list here instead of a map.  Do you have a strong preference for a dictionary?

```json
[
  {
    "asset": "A18446744073709551615",
    "asset_longname": "PIZZA.anchovy"
  },
  {
    "asset": "A18446744073709551616",
    "asset_longname": "PIZZA.cheese"
  }
]
```

This will need unit tests in vectors.py.

droplister:
I have tested this code on a federated node that I spun up and it works as expected. It's possible there are more optimal ways to SELECT from the db and/or build the list. Open to suggestions!

Now, you can call: get_asset_names(), get_asset_names(Longnames=True), or get_asset_longnames(). The longnames method/parameter will return a list that looks like @deweller's comment.

get_asset_names has no existing tests, so I assume that this PR is okay without them.

@jdogresorg let me know if this covers your open feature request.

droplister:
It occur to me that this code is not DRY, so which is better, the flag or the new method?

jdogresorg:
This looks good to me and addresses issue #992 :) Nice job.

jdogresorg:
If you want DRY, prolly want to just have `get_asset_longnames()` call `get_asset_names(true)`;

droplister:
Possible documentation...
```
###get_asset_names

**get_asset_names(longnames=false)**

**Parameters:**

  * **longnames** (*boolean*): Set to `true` to include asset_longnames in list (i.e. subassets)

**Return:**

  A list of all Counterparty asset names, ordered alphabetically. 

  Set longnames to `true` to return a list of objects, one for each Counterparty asset.
  Each object has the following properties:

  - **asset** (*string*): The [asset](#assets) name.
  - **asset_longname** (*string*): The [subasset](#subassets) name.

###get_asset_longnames

**get_asset_longnames()**

**Parameters:**

  None

**Return:**

  An alias of [get_asset_names(longnames=true)](#get_asset_names). See above.
```

chiguireitor:
Merged to develop manually, but interface not marking it as merged, closing this one here

Issue #997, opened 2017-09-16
[Bitcoin 0.15] accommodate changes to fee estimate RPCs


unsystemizer:
Details on what's changed:
https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.15.0.md#changes-to-fee-estimate-rpcs

droplister:
I submitted a PR which I think accomplishes this change. My code does not let you choose between estimatefee or estimatesmartfee, but I think that's okay because estimatefee is deprecated.

PR #1008, opened 2017-10-25
Use estimatesmartfee instead of deprecated estimatefee

droplister:
Taking a shot at Issue #997.

Changes config value ESTIMATE_FEE_NBLOCKS to ESTIMATE_FEE_CONF_TARGET and add a ESTIMATE_FEE_MODE config value, cascades these changes through the code to use estimatesmartfee.

### DIFF
--- a/counterpartylib/lib/backend/addrindex.py
+++ b/counterpartylib/lib/backend/addrindex.py
@@ -244,7 +244,7 @@ def fee_per_kb(conf_target, mode):
     feeperkb = rpc('estimatesmartfee', [conf_target], mode)
 
     if feeperkb == 'Insufficient data or no feerate found':
-            return None
+        return None
 
     return int(feeperkb * config.UNIT)
 
--- a/counterpartylib/lib/backend/addrindex.py
+++ b/counterpartylib/lib/backend/addrindex.py
@@ -241,15 +241,9 @@ def fee_per_kb(conf_target, mode):
     :return: fee_per_kb in satoshis, or None when unable to determine
     """
 
-    # we need to loop because sometimes bitcoind can't estimate a certain conf_target
-    retry = 0
-    feeperkb = -1
-    while feeperkb == -1:
-        feeperkb = rpc('estimatesmartfee', [conf_target], mode)
-        conf_target += 1
-        retry += 1
+    feeperkb = rpc('estimatesmartfee', [conf_target], mode)
 
-        if retry > 10:
+    if feeperkb == 'Insufficient data or no feerate found':
             return None
 
     return int(feeperkb * config.UNIT)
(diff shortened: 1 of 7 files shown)


counterpartylib/lib/backend/addrindex.py
```diff
@@ -234,18 +234,19 @@ def getrawtransaction(tx_hash, verbose=False, skip_missing=False):
...
+    :param conf_target:
+    :param mode:
     :return: fee_per_kb in satoshis, or None when unable to determine
     """
 
-    # we need to loop because sometimes bitcoind can't estimate a certain nblocks
+    # we need to loop because sometimes bitcoind can't estimate a certain conf_target
     retry = 0
     feeperkb = -1
     while feeperkb == -1:
```
droplister: estimatesmartfee reports errors, as I understand it, whereas estimatefee would return -1.

droplister: If someone reviewing this could suggest a solution that would be great. I'm not that familiar with python.

Error found here: https://github.com/bitcoin/bitcoin/pull/10707/files#diff-9651347c8e00bed3ddc7631de569406dR857

Here is one possible way that just trades error messages for error message.

```
def fee_per_kb(conf_target, mode):
    """
    :param conf_target:
    :param mode:
    :return: fee_per_kb in satoshis, or None when unable to determine
    """

    # we need to loop because sometimes bitcoind can't estimate a certain conf_target
    retry = 0
    feeperkb = 'Insufficient data or no feerate found'
    while feeperkb == 'Insufficient data or no feerate found':
        feeperkb = rpc('estimatesmartfee', [conf_target], mode)
        conf_target += 1
        retry += 1

        if retry > 10:
            return None

    return int(feeperkb * config.UNIT)
```

Here is a less optimistic approach that does not make 10 attempts.

```
def fee_per_kb(conf_target, mode):
    """
    :param conf_target:
    :param mode:
    :return: fee_per_kb in satoshis, or None when unable to determine
    """

    feeperkb = rpc('estimatesmartfee', [conf_target], mode)

    if feeperkb == 'Insufficient data or no feerate found':
            return None

    return int(feeperkb * config.UNIT)
```

droplister: I committed the second option, in the meantime.

unsystemizer:
@droplister does this work the same with 0.15 (because IIRC some things lately changed between 0.14 and 0.15)?

droplister:
I followed the 0.15 release note link you provided. But what specifically are you asking about?

unsystemizer:
I mean are the changes okay for Bitcoin Core 0.15? Or 0.14?

chiguireitor:
Merged manually, github not updating status, closing this

PR #1010, opened 2017-11-03
Indexd backend

deweller:
Replaces addrindex backend with indexd backend

removes in-memory unconfirmed transactions cache.


unsystemizer:
- `get_unspent_txouts`: what is the reason it changed? is it mandatory to change?

- index behind: we now assume that `txindex` is always up to date. Not sure if that's risky, but assuming it too moves to indexd sometime soon and that it's much smaller than addrindex, it is unlikely to be behind with 0.15 which is much faster compared to 0.13 with txindex & addrindex. Besides I've never ran into this error (indexer behind > 1 block).

deweller:
get_unspent_tx_outs has to change because indexd doesn't index the scriptPubKey.  We could include the scriptPubKey by including additional calls to bitcoind, but that would be less efficient. We only need it at the time of transaction formation, which is what counterparty-lib does now.

As for the index being behind.  The API should not respond until indexd is always up to date.  There is no danger here as indexd is not used to maintain consensus.  It is only used by the API for forming bitcoin transactions.

ChangeLog.md
```diff
@@ -1,7 +1,16 @@
 ## Library Versions ##
-* master (unreleased)
+* development (unreleased)
+    * Use Bitcoin 0.15.0 with indexd as a transaction index
```
rubensayshi: should also state we removed btcd

rubensayshi:
<3 awesome work, so much lines of code removed that were a some of the more painful things we had to deal with

should update the README a bit I think to at least state it requires indexd running, with fednode very few people will need any instructions but for consistency and the odd case of someone running without fednode it should be updated.

utACK otherwise

PR #1015, opened 2017-11-11
CIP-6 p2sh encoding

deweller:
Needs rebase after complex unit tests are merged

[UPDATE: rebase complete]


counterpartylib/lib/transaction.py
```diff
@@ -381,17 +378,7 @@ def construct (db, tx_info, encoding='auto',
...
-            if len(data) + len(config.PREFIX) <= config.OP_RETURN_MAX_SIZE:
-                encoding = 'opreturn'
-            else:
-                encoding = 'multisig'
-
-        elif encoding not in ('pubkeyhash', 'multisig', 'opreturn'):
-            raise exceptions.TransactionError('Unknown encoding‐scheme.')
-
-
+        # @TODO: p2sh encoding require signable dust key
```
rubensayshi: what's this TODO?

counterpartylib/lib/transaction_helper/p2sh_encoding.py
```diff
@@ -0,0 +1,153 @@
...
+            source = None
+            redeem_script_is_valid = True
+    else:
+        pubkey = None
+        source = None
+        redeem_script_is_valid = False
+
+    return pubkey, source, redeem_script_is_valid
+
+def make_p2sh_encoding_redeemscript(datachunk, n, pubKey=None, multisig_pubkeys=None, multisig_pubkeys_required=None):
```
rubensayshi: there's some camelCase vars in this function

counterpartylib/lib/transaction.py
```diff
@@ -340,39 +313,63 @@ def construct (db, tx_info, encoding='auto',
...
+    '''Determine encoding method'''
+
+    if data:
+        desired_encoding = encoding
+        # Data encoding methods (choose and validate).
+        if desired_encoding == 'auto':
+            if len(data) + len(config.PREFIX) <= config.OP_RETURN_MAX_SIZE:
+                encoding = 'opreturn'
+            else:
+                encoding = 'p2sh' if not old_style_api and util.enabled('p2sh_encoding') else 'multisig'  # p2sh is not possible with old_style_api
```
rubensayshi: should we also check `and not (source_is_p2sh and dust_return_pubkey is None)`? because if that's not true then we'd fail later

counterpartylib/lib/transaction.py
```diff
@@ -451,111 +444,109 @@ def construct (db, tx_info, encoding='auto',
...
+                                                             data_output=data_output,
+                                                             change_output=change_output,
+                                                             pubkey=dust_return_pubkey,
+                                                             multisig_pubkeys=p2sh_source_multisig_pubkeys,
+                                                             multisig_pubkeys_required=p2sh_source_multisig_pubkeys_required)
+            unsigned_pretx_hex = binascii.hexlify(unsigned_pretx).decode('utf-8')
+
+        # with segwit we already know the txid and can return both
+        if segwit:
+            pretx_txid = hashlib.sha256(unsigned_pretx).digest()  # this should be segwit txid
```
rubensayshi: this isn't correct cuz we stuff the scriptPubKey into the scriptSig so the signing party knows what to sign with, I think we should just remove the segwit stuff for now until counterwallet actually supports it maybe?

jdogresorg:
We should close this pull request since p2sh encoding was recently added with commit e8f0673c72696d3f151ba6f8e574d8858a925f9f

PR #1024, opened 2018-01-25
Fix the minimum fee calculation.

cryptcoin-junkey:
We must consider the size of the change output.

### DIFF
--- a/counterpartylib/lib/transaction.py
+++ b/counterpartylib/lib/transaction.py
@@ -436,7 +436,7 @@ def construct (db, tx_info, encoding='auto',
 
     data_btc_out = sum([data_value for data_chunk in data_array])
 
-    '''Inputs'''
+    '''Size of outputs (Total)'''
 
     # Calculate collective size of outputs, for fee calculation.
     p2pkhsize = 25 + 9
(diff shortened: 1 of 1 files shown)


chiguireitor:
this probably affects the test vectors on your side, please run the unit_tests.py test suite on your side (check circle.yml for details on how to run it). Also, a full integration test should be done.... prolly this fee estimation will need a block activation and minor version for activation.

mandelmonkey:
does this take into account transactions that don't have change?
seems like it adds an extra p2pkhsize but if the transaction doesnt have a change output then this would lead to an incorrect estimation

counterpartylib/lib/transaction.py
```diff
@@ -449,7 +449,13 @@ def construct (db, tx_info, encoding='auto',
             data_output_size = data_output_size + len(data)
     else:
         data_output_size = p2pkhsize   # Pay‐to‐PubKeyHash (25 for the data?)
-    outputs_size = (p2pkhsize * len(destination_outputs)) + (len(data_array) * data_output_size)
+
+    # destination + data + 'change'
+    # Strictly we can't determine if the change output is contained in this transaction.
+    # But in almost all cases, it will be contained.
+    outputs_size = (p2pkhsize * len(destination_outputs)) + (len(data_array) * data_output_size) + p2pkhsize
```
chiguireitor: This will lead to overestimation for non-change sends.... these could happen a lot on really fragmented wallets that have a lot of small outputs.... although this could lead to better estimation on a common 280 bytes op (i.e. a standard enhanced send) small outputs would become unusable.

Gotta weigh which case is the most common and approve or reject based on that.

PR #1041, opened 2018-08-10
Make Address Options Bitwise

droplister:
The design of [CIP 12](https://github.com/CounterpartyXCP/cips/blob/master/cip-0012.md) and the SQLite database are setup to support bitwise address options, but the current checks in place are not sufficient to support multiple address options. This PR is designed to change that.

This pull request does three things:
1. Treats options as a bitwise flag. Current code is more or less boolean.
2. Refactors parts of the code to better support additional address options.
3. Creates reusable utility functions for parsing, validating, and confirming options.

I don't work with bitwise flags often, but my understanding is that the current code only evaluates whether options are within a given range which leaves room for non-existent options to be set and the current checks in place assume the only option present will be ```config.ADDRESS_OPTION_REQUIRE_MEMO```.

For example, given ```config.ADDRESS_OPTION_MAX_VALUE=(1 | 2 | 4 | 16)```. If you were to set your address to have ```options=8``` or ```options=9```, the code would incorrectly permit these non-existent configurations.

Or, if your address showed ```options=3```, the current code would think your address wasn't requiring memos, when it was, because ```1 != 3```, but that's not how you evaluate a bitwise flag.

There is nothing necessarily wrong with how the code currently works, but the goal of this PR is to give us the option of multiple address options AND some of these utility functions could be reused, if we were to have asset options set through issuances, similar to the way locking works.

A good resource for understanding bitwise vs booleans:
https://stovepipe.systems/post/using-bitwise-instead-of-booleans

### DIFF
--- a/counterpartylib/lib/exceptions.py
+++ b/counterpartylib/lib/exceptions.py
@@ -42,4 +42,7 @@ class BalanceError(Exception):
 class EncodingError(Exception):
     pass
 
+class OptionsError(Exception):
+    pass
+
 # vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4
--- a/counterpartylib/lib/messages/broadcast.py
+++ b/counterpartylib/lib/messages/broadcast.py
@@ -210,7 +210,7 @@ def parse (db, tx, message):
     # Options? Should not fail to parse due to above checks.
     if util.enabled('options_require_memo') and text and text.lower().startswith('options'):
         options = util.parse_options_from_string(text)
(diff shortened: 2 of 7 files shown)


droplister:
Also, I should note, the design of this assumes that you are overwriting the current options, so say you had two or three options already active, at the GUI-level, you would fetch the current options, add or subtract what you wanted, and then define the new configuration. The code, as written, also assumes that it is okay for any options flag to be turned on or off.

Issue #1044, opened 2018-10-08
Get_holders is returning holders with amount 0, which is affecting the calculation for dividend anti spam fee


chiguireitor:
When getting the amount of holders, the return contains addresses which held the target at for dividend distribution even if they currently hold 0 of the asset. This affects the antispam fee for distributions as multiple holders which won't receive the distributed asset are included into the calculation.

droplister:
I think this behavior may be preferable. The addresses that show up with zero quantity previously owned the asset in question, so reporting zero quantities sort of makes it explicit and definitive that they were not owned and did not receive any dividend.

chiguireitor:
It is interesting indeed, but shouldn't affect the dividend calculation, as it punishes and could eventually be used to attack a dividend distributing endeavor by creating lots of fake 0 holder addresses by moving a distribution-Target asset between multiple holders.

chiguireitor:
Indeed, and having someone with balance 0 is an error, as they aren't holders.

Attacks don't need to be economical on the face value if they imply stopping a working mechanism for a competing product. Making each distribution cost 1000x more could cost about $10 on a low fee scenario, that's economically viable for an interested party.

We could add an additional call on the api to query historical holders, in fact i believe a get_balances would serve the exact same purpose as balance ==  0 rows are returned.

jdogresorg:
Also, worth mentioning that we have over 190K records in the `credits` table where quantity=0.
IMO we should do everything we can to cut down on the creation of unnecessary database records like this.
![image](https://user-images.githubusercontent.com/376028/46592542-aa7cec00-ca78-11e8-97e3-94e493726245.png)

droplister:
I find it useful to know that 0 was credited. It can be meaningful information to a user.

chiguireitor:
A credit of 0 could be useful, yes, but calculation of an antispam fee including bogus data isn't.

jdogresorg:
>  in fact i believe a get_balances would serve the exact same purpose as balance == 0 rows are returned.

Can confirm 0 balance rows are returned in `get_balances` api calls. 

![image](https://user-images.githubusercontent.com/376028/46593181-ab188100-ca7e-11e8-8d6d-0e037e2136a4.png)

chiguireitor:
Fixed in develop

PR #1045, opened 2018-10-08
holders fix

jdogresorg:
- fix to only return holders with quantity > 0


jdogresorg:
Should fix issue #1044

Issue #1104, opened 2021-01-20
zero dividend error - create_dividend


jdogresorg:
Occasionally users will complain about getting 'Zero dividend' errors when trying to pay out dividends.

Here is an instance of that behavior... 

Problem Request
---
```
{
	"method": "create_dividend",
	"params": {
		"source": "1BH9sBVQ645gx2D3pfkc95SSvvEXpFhw2N",
		"asset": "PEPEYES",
                "dividend_asset": "STPEPESUNC",
		"quantity_per_unit": 3,
		"fee": 30277,
		"allow_unconfirmed_inputs": true
	},
	"jsonrpc": "2.0",
	"id": 0
}

```
Problem Details
---
There are 9 holders of PEPEYES tokens and there are 600 STPEPESUNC tokens issued... so, the dividend SHOULD work and pay out 3 STPEPESUNC to every holder of 1 token of PEPEYES.... instead we get a 'zero dividend' error returned

https://github.com/CounterpartyXCP/counterparty-lib/blob/765c00b76857d4181bf69a4ddd5b592790073689/counterpartylib/lib/messages/dividend.py#L110

The repo code seems to indicate that 'zero dividend` is thrown when there are no holders, yet clearly there are 9 PEPEYES holders... so what is going on?!?!

![image](https://user-images.githubusercontent.com/376028/105231711-22716200-5b1c-11eb-803b-815dc3e1ddc0.png)

jdogresorg:
After some investigation, it appears that the issue is that for non-divisible assets config.UNIT is set to 100000000 instead of 1

So, when we hit this line

https://github.com/CounterpartyXCP/counterparty-lib/blob/765c00b76857d4181bf69a4ddd5b592790073689/counterpartylib/lib/messages/dividend.py#L102

instead of calculating 
`    dividend_quantity /= 1`
we are doing
`        dividend_quantity /= 100000000`

this is why the `dividend_quantity` is being returned as 0 for all holders, resulting in the 'zero dividends' error

jdogresorg:
after some further investigation, I think my fix is not good as we have

chiguireitor:
Yes, the fix seems to affect adversely divisible asset dividends

jdogresorg:
The fix appears to be working :)
![image](https://user-images.githubusercontent.com/376028/107824545-827fb080-6d36-11eb-88cb-650424adc64a.png)

longhoangwkm:
I still got this error on testnet

[...]

PR #1105, opened 2021-01-21
stopped dividing dividend_quantity for non-divisible dividend assets

jdogresorg:
- should address issue #1104


chiguireitor:
This probably needs a protocol change flag

Issue #1108, opened 2021-02-08
Incorrect utxo amount due to floating point error


longhoangwkm:
https://github.com/CounterpartyXCP/counterparty-lib/blob/e3ea6f574a1fc6f4d275924bac1a9221298dff73/counterpartylib/lib/backend/addrindexrs.py#L400

Get utxos list of address `mgeZMQw3SuCLiMoWba49Ga78vjVBpt1tyW` return incorrect amount value.

```
 unpack_vout: {'scriptPubKey': {'reqSigs': 1, 'hex': '76a9140c684d69c0929221e5918b81a9709b784dc6fd2f88ac', 'addresses': ['mgeZMQw3SuCLiMoWba49Ga78vjVBpt1tyW'], 'asm': 'OP_DUP OP_HASH160 0c684d69c0929221e5918b81a9709b784dc6fd2f OP_EQUALVERIFY OP_CHECKSIG', 'type': 'pubkeyhash'}, 'n': 1, 'value': 0.00015743}
```

```
int(vout["value"] * config.UNIT)
# python >
int(0.00015743 * 100000000) # Return 15742 (Incorrect value)
```

We should use bigdecimal to handle.

**Solution**: 
```python
from decimal import *
int(Decimal(str(0.00015743)) * 100000000) # 15743
```

longhoangwkm:
I fix: https://github.com/CounterpartyXCP/counterparty-lib/pull/1109 cc @chiguireitor

cryptcoin-junkey:
@longhoangwkm I think this PR is reasonable.
And IMO, does it need to add a protocol change enabling this ?

longhoangwkm:
@cryptcoin-junkey  I think we do not need a new protocol change. That change doesn't change any API or change DB value

cryptcoin-junkey:
Ah, I should post my first comment to your PR wakumo/counterparty-lib#2 .
I agree that there is no concern to fix addrindexrs.py. :+1:

PR #1109, opened 2021-02-08
fix: floating point error

longhoangwkm:
Fix: https://github.com/CounterpartyXCP/counterparty-lib/issues/1108

### DIFF
--- a/counterpartylib/lib/backend/addrindexrs.py
+++ b/counterpartylib/lib/backend/addrindexrs.py
@@ -14,6 +14,8 @@ import binascii
 import hashlib
 import signal
 import bitcoin.wallet
+import decimal
+D = decimal.Decimal
 
 from counterpartylib.lib import config, util, address
 
@@ -390,7 +392,7 @@ def unpack_vout(outpoint, tx, block_count):
         "txId": tx["txid"],
         "vout": outpoint[1],
         "height": height,
-        "value": int(vout["value"] * config.UNIT),
+        "value": int(D(str(vout["value"])) * config.UNIT),
         "confirmations": tx["confirmations"]
     }
 
--- a/counterpartylib/lib/messages/broadcast.py
+++ b/counterpartylib/lib/messages/broadcast.py
@@ -118,7 +118,7 @@ def validate (db, source, timestamp, value, fee_fraction_int, text, block_index)
 def compose (db, source, timestamp, value, fee_fraction, text):
 
     # Store the fee fraction as an integer.
-    fee_fraction_int = int(fee_fraction * 1e8)
+    fee_fraction_int = int(D(str(fee_fraction)) * 1e8)
 
     problems = validate(db, source, timestamp, value, fee_fraction_int, text, util.CURRENT_BLOCK_INDEX)
     if problems: raise exceptions.ComposeError(problems)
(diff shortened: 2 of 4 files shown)


counterpartylib/lib/messages/issuance.py
```diff
@@ -226,7 +226,7 @@ def validate (db, source, destination, asset, quantity, divisible, callable_, ca
             if util.enabled('numeric_asset_names'):  # Protocol change.
                 if subasset_longname is not None and util.enabled('subassets'): # Protocol change.
                     # subasset issuance is 0.25
-                    fee = int(0.25 * config.UNIT)
+                    fee = int(D('0.25') * config.UNIT)
```
cryptcoin-junkey, line 229: Really it's required to fix in this line? 

```
Python 3.6.9 (default, Jan 26 2021, 15:33:00)
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 0.25 * 100000000
25000000.0
```

counterpartylib/lib/backend/addrindexrs.py
```diff
@@ -390,7 +392,7 @@ def unpack_vout(outpoint, tx, block_count):
         "txId": tx["txid"],
         "vout": outpoint[1],
         "height": height,
-        "value": int(vout["value"] * config.UNIT),
+        "value": int(D(str(vout["value"])) * config.UNIT),
```
cryptcoin-junkey, line 395: Why wrapping `str()` ?
It can be `D(vout["value"])`, IMO.

counterpartylib/lib/messages/broadcast.py
```diff
@@ -118,7 +118,7 @@ def validate (db, source, timestamp, value, fee_fraction_int, text, block_index)
 def compose (db, source, timestamp, value, fee_fraction, text):
 
     # Store the fee fraction as an integer.
-    fee_fraction_int = int(fee_fraction * 1e8)
+    fee_fraction_int = int(D(str(fee_fraction)) * 1e8)
```
cryptcoin-junkey, line 122: Why wrapping str() ?

cryptcoin-junkey: Ah, I see. Wrapping is reasonable.

counterpartylib/lib/backend/addrindexrs.py
cryptcoin-junkey: My bad. Wrapping is reasonable.

cryptcoin-junkey:
I reviewed whole code including this PR.
I suggest that the better alternative way is to reduce floating point values by using `int()` cast.
Almost all values in assets can be integer. 

In Python3, max value of integer becomes unlimited. Integer is easier to handle than the `decimal` package.

Indeed we may be not able to avoid using `decimal` in some cases (like addrindexrs.py in this PR).
Even though it should be a special case, IMO.

cryptcoin-junkey:
@longhoangwkm My suggestion is in https://github.com/monaparty/counterparty-lib/pull/49.
Note that there have some Monaparty dependent patches.

jdogresorg:
This issue has been resolved in commit https://github.com/CounterpartyXCP/counterparty-lib/commit/8395d6e1bcbc32b5b4dd97b91354bfacbadf4b27

PR #1115, opened 2021-03-27
Store memos in "MPMA send" as BLOB.

cryptcoin-junkey:
This patch is based on Monaparty#25.
It should be worked on Counterparty but I've tested on Monaparty only.
Please be careful before applying this PR.

The `memo` field in the `sends` table should be BLOB (`bytes`).
But the MPMA decoder sets as `str`.
So a type mismatch error will be occurred in `api.py` when a user call `get_sends` API.

As the type of `memo` is BLOB, decoding `memo` to UTF-8 may be failed.
It's better to decode with `replace`, IMO.

### DIFF
--- a/counterpartylib/lib/api.py
+++ b/counterpartylib/lib/api.py
@@ -306,7 +306,7 @@ def adjust_get_sends_results(query_result):
                 send_row['memo'] = None
             else:
                 send_row['memo_hex'] = binascii.hexlify(send_row['memo']).decode('utf8')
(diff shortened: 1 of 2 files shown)


cryptcoin-junkey:
This patches doesn't effect to any consensus.
But values in `sends` table must be updated.
So fednode operators must reparse their database.

cryptcoin-junkey:
Please have a look monaparty#43 also. Totally MPMA memo have complex issues.

jdogresorg:

PR #1117, opened 2021-04-07
Use `STATUS_OPEN` instead of `dispenser_status`.

cryptcoin-junkey:
Since `STATUS_OPEN_EMPTY_ADDRESS` was added.
It can be closed dispensers that have `STATUS_OPEN` as their status.

This patch may be required protocol changes.


jdogresorg:
This issue has been resolved in the oracled dispensers PR #1167 which has been merged into master

PR #1121, opened 2021-07-11
- Remove "FROM counterparty/base" to upgrade to Ubuntu 18.04 in order…

pataegrillo:
… to install the required version of setuptools-markdown


jdogresorg:
Did test builds on Ubuntu 18.04 and on Ubuntu 20.10 and both rebuilt fine using this file... merging

PR #1128, opened 2021-08-26
- Fixing utxos not getting updated on addrindexrs.py when confirmations are 0

pataegrillo:


jdogresorg:
These updates :
- Fix utxos not getting updated on addrindexrs.py when confirmations are 0
- Fix sanity error issue with letting different sources refill a dispenser

jdogresorg:
rejecting changes since the utxos issue probably does not happen on Counterparty, and hotfix code will be forthcoming in cleaner PR

PR #1142, opened 2021-11-12
- Fixing MPMA segwit change address issue, resulting in loss of btc

pataegrillo:


jdogresorg:
fix does not resolve core issue

PR #1145, opened 2022-01-12
- Fixed an error that happens when trying to reissue an existing toke…

pataegrillo:
…n with a name longer than 12 characters (old tokens)


jdogresorg:
tested fix on public.coindaddy.io and was able to parse past the problem transaction.... merging fix

PR #1146, opened 2022-01-12
- Fixed bitcoins loss due to bad change address when sending mpma usi…

pataegrillo:
…ng a segwit address

- Fixed bad source parsing when reading a mpma with a segwit source
- Adjusted the returning format when creating transactions and using the old_style_api=false. Now returns one item when there is only one to return, but an Array when there are more than one


jdogresorg:
closing

Issue #1132, opened 2021-09-14
Dispensers do not work on addresses that start with 3


jdogresorg:
Here is a dispenser address (3FNMQamxkzfBD5NKbjRa2qRd8vY7XxeDzy) and which SHOULD have been triggered when BTC was sent
https://xchain.io/tx/4a015f732c7d1c56e2cfa9e8fd697c9d36f58e79e596ec8cfcce02cc69777508

Here is the payment transaction to the 3FNMQamxkzfBD5NKbjRa2qRd8vY7XxeDzy address and yet a dispense was not triggered
https://blockstream.info/tx/70ed3f789a818c30859795df9fb052d943fcbd83acb1591ea3d5f1257bae1baf

AryanJ-NYC:
On top of buyers not being able to buy assets via 3 address dispensers, sellers' assets are also stuck on 3 address dispensers.

For example, I (stupidly) opened dispensers at https://host3.xchain.io/tx/d641fdbbc039b88335f8222c345743fbc96151865cf34ecb5819a737328234c7 and https://host3.xchain.io/tx/4a015f732c7d1c56e2cfa9e8fd697c9d36f58e79e596ec8cfcce02cc69777508. Since dispensers don't dispense on 3 addresses, my digital assets are unfortunately stuck.

I'm not complaining or anything; just providing a data point. If I can be of any service (more information on my case or even alpha testing), I'm very happy to help.

Thanks all!

jdogresorg:
Issue should be fixed with commit 127d5c6

PR #1155, opened 2022-02-22
Fixed '3' type address dispensers not triggering

pataegrillo:


jdogresorg:
Should fix issue #1132

PR #1156, opened 2022-02-23
- Multiple output dispenses from the same tx are now allowed

pataegrillo:


jdogresorg:
code does not parse multiple outputs, just the last one

PR #1172, opened 2022-07-15
- Fixed second mpma tx fee lower than exact_fee

pataegrillo:


jdogresorg (approved):
looks good in code review and in testing

PR #1177, opened 2022-07-20
Fix ARC4 tests to use static dust size

windsok:
It looks like when these ARC4 tests were written, they made the assumption that the default dust size set in `config.py` would always stay the same. Since this value was modified the tests break as the transaction hex returned contains a different output value.

Since these tests are not intended to test any part of the dust / output value logic, it seems safe to just pass a static known dust size here, which corresponds to what the output value in the expected hex is.

### DIFF
--- a/counterpartylib/test/arc4_test.py
+++ b/counterpartylib/test/arc4_test.py
@@ -122,7 +122,7 @@ def test_transaction_arc4_mocked(server_db):
     """
     v = int(100 * 1e8)
     tx_info = send.compose(server_db, ADDR[0], ADDR[1], 'XCP', v)
-    send1hex = transaction.construct(server_db, tx_info)
+    send1hex = transaction.construct(db=server_db, tx_info=tx_info, regular_dust_size=5430)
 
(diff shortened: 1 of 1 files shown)


jdogresorg:
Dust for p2sh outputs is 546 sats (as of a few days ago), except the last p2sh output could be larger, in order to make sure that the second MPMA tx has enough BTC in the outputs to cover the cost of the specified BTC miners fee.

J-Dog
***@***.***
GPG : 0C85 9EE8 BC7D 0A56 055A  6D31 2411 DE07 1B00 02BF

windsok:
Right so the issue with these tests is that they have hardcoded an expected output raw transaction hex which is assuming that the dust value will be `5430` when creating the transaction.

There is a couple of ways we can fix this test, either we can update the expected raw transaction hex to use the different value, or we can just pass the `transaction.construct` function the dust value which should be used to produce the expected transaction hex.

I figured the latter is a better approach, as it means this test won't break again if the dust value is updated to a different value, and this test is for testing that the ARC4 encryption is working as expected, and not actually intended to test anything related to the dust value / output values, so the value we use does not really matter, it only matters that it matches what our expected transaction raw hex is using.

jdogresorg:
Sounds reasonable. Merging now.

jdogresorg (approved):

PR #1179, opened 2022-07-20
Update test scenario db dumps to fix expected sqlite default page_size

windsok:
In some of the integration tests we run through a scenario then dump the database from sqlite to compare what was created there to a known expected output.

Back in 2016 the default `page_size` for sqlite was updated from `1024` to `4096` - https://www.sqlite.org/pgszchng2016.html

This is being included in the checks, and some tests are failing due to our expected output's having the old `1024` value, instead of `4096`.

A better fix might be to modify the test logic to not compare on these values, but since this value is unlikely to change again anytime soon, seems OK for now to just update the expected value to `4096`


jdogresorg:
looks good. merging

PR #1181, opened 2022-07-22
- Now mpma will have the last output increased to reach the exact fee

pataegrillo:


jdogresorg:
This looks good in testing on testnet. All outputs are at 546 sats (DUST) except for the last output which is increased to add any sats to make sure that we use `exact_fee` amount in the transaction to cover the miners fees.

PR #1197, opened 2022-08-13
- Added "asset" param to get_asset_info

pataegrillo:


jdogresorg:
looks good in testing... API returns results when specifying `asset` or `assets`

Issue #1129, opened 2021-08-30
add tx_index to close dispenser actions


jdogresorg:
We should include the `tx_index` in messages table when closing dispensers to make it explicit which dispenser we are closing.

Currently we only note the `source` and `asset` in the close update, which is not specific enough as an address could have had dispensers setup previously on the address... listing the specific `tx_index` of the dispenser that is being closed would make updates using the messages table much easier.

Currently have to do some hackery in counterparty2mysql to only update status if going from 0 (open) to 10 (closed) rather than just being able to reference the specific `tx_index` of the dispenser transaction

https://github.com/jdogresorg/counterparty2mysql/blob/master/counterparty2mysql.php#L330-L337

```
6238208|698195|insert|dispensers|{"asset": "PEPECASH", "block_index": 698195, "escrow_quantity": 8200000000000, "give_quantity": 100000000, "give_remaining": 8200000000000, "satoshirate": 166, "source": "1MX2kfSP5qspDcWiDg15M95PQrYityYG52", "status": 0, "tx_hash": "b8a7de441f2f82a1937f755259e23842aa6084eca02cbc699806572de1f261a1", "tx_index": 1625548}|1630284400
6239795|698220|update|dispensers|{"asset": "PEPECASH", "block_index": 698220, "escrow_quantity": 8200000000000, "give_quantity": 100000000, "give_remaining": 8000000000000, "prev_status": 0, "satoshirate": 166, "source": "1MX2kfSP5qspDcWiDg15M95PQrYityYG52", "status": 0, "tx_hash": "b8a7de441f2f82a1937f755259e23842aa6084eca02cbc699806572de1f261a1", "tx_index": 1625548}|1630296745
6239906|698223|update|dispensers|{"asset": "PEPECASH", "block_index": 698223, "escrow_quantity": 8200000000000, "give_quantity": 100000000, "give_remaining": 7100000000000, "prev_status": 0, "satoshirate": 166, "source": "1MX2kfSP5qspDcWiDg15M95PQrYityYG52", "status": 0, "tx_hash": "b8a7de441f2f82a1937f755259e23842aa6084eca02cbc699806572de1f261a1", "tx_index": 1625548}|1630299170
[...]

PR #1200, opened 2022-08-13
start passing tx_index in bindings on dispenser close

jdogresorg:
addresses issue #1129


jdogresorg:
looks good in testing
```
sqlite> select * from messages where bindings like '%"status": 10%' order by block_index desc limit 1;
112394|2316287|update|dispensers|{"asset": "BACONTITS", "block_index": 2316287, "source": "tb1q50kxp76j9l0k9jgwasvcz4mcz0v03fv2wj67a4", "status": 10, "tx_index": 29294}|1660427173

```

PR #1201, opened 2022-08-17
More efficient asset supply queries

jotapea:
The asset supply function was making queries to the complete set of issuances and destructions, and then filtering from these results for the asset of interest. This PR adds 2 new functions that provide the exact same result, but now querying the asset directly.

### DIFF
--- a/counterpartylib/lib/util.py
+++ b/counterpartylib/lib/util.py
@@ -691,13 +691,27 @@ def destructions (db):
     cursor.close()
     return destructions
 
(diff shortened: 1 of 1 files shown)


pataegrillo:
Great!, definitely more efficient! Thanks for contributing!

jdogresorg (approved):

jdogresorg:
This looks good to me as well. Nice work @jotapea 👍

PR #1207, opened 2022-08-26
update `get_balances()` API call with `divisible` field

jdogresorg:


jdogresorg:
looks good in testing

image

PR #1208, opened 2022-08-26
9.60.0 release

jdogresorg:
We should update `protocol_changes.json` with more appropriate activation blocks before we merge this PR and do a new `counterparty-lib` release.


jdogresorg:
bumped activation blocks to be 2 weeks to give users more time to update their fednodes before activation.

PR #1211, opened 2022-09-08
Fixed destructions logs quantity for divisible assets

pataegrillo:


jdogresorg:
looks good in testing.

PR #1219, opened 2022-12-23
Callability verification removal

pataegrillo:


pataegrillo:
Is really hard to tell, the whole block_hash is calculated using three components:

1. transactions 
2. credits and debits
3. messages (data parsed by cp)

If just one of them differ, even by a little bit, then the whole hash will be different. So, if txs, credits and debits are the same for a block, you have to see messages tables too, if for example there is one more field, then the hash will be also different

counterpartylib/protocol_changes.json
```diff
@@ -412,5 +412,12 @@
       "minimum_version_revision": 6,
       "block_index": 718318,
       "testnet_block_index": 2135609
+  },
+  "issuance_callability_parameters_removal": {
+      "minimum_version_major": 9,
```
jdogresorg, line 417: Please update to:
```
"minimum_version_major": 9
"minimum_version_minor": 60,
"minimum_version_revision": 2,
"block_index": 825000
```

jdogresorg:
looks good in testing.

Issue #1148, opened 2022-01-18
Just 1st dispenser dispenses when batch-sending sats to multiple dispensers


legalizemath:
This is the transaction (generated with a simple blue wallet)

https://blockstream.info/tx/11eb2b730e2e383a657c51d7b04bd55271d1e86ac9a2c74d3e3f6c78e88e23ff

where correct exact sats were sent to 5 different dispensers that all, even after tx confirmation, still had available assets to dispense

can see here just 1st issued https://xchain.io/tx/11eb2b730e2e383a657c51d7b04bd55271d1e86ac9a2c74d3e3f6c78e88e23ff

here are each of 5 dispensers in order of utxo output numbers (blockheight of tx = 719244):

* n0 https://xchain.io/tx/7ad422d7f74fa4fd8552ce50a295c48f4bf259bc3351d3a6482b24b7300829a7 (worked)
* n1 https://xchain.io/tx/b09ca2bd1e48a09b03bde25e91acbb250f4cdc43daae3942f8fc033b8b8c21fe (fail)
* n2 https://xchain.io/tx/8177deaaeb68ade1a099913a4d130f87cde02fdc6accba65dbf44660d0980793 (fail)
* n3 https://xchain.io/tx/ce04585fa2f377567564b21878bc1e132f39db37f8c84472f2349b62a9d79a62 (fail)
* n4 https://xchain.io/tx/4119116c715c1adebb4ae475515cda106e0a275d2f5447bb96dd1e80fb1515a3 (fail)

n5 is change

suggestion: as this might be current consensus rule if not just me misundersatnding, might suggest devs warnings not to batch as batching is now part of bluewallet and samourai wallet for starters and saves a lot on fees. it would seem obvious from address is same for each spend to dispenser that each would credit that address but it doesn't appear yet to be the case.

jdogresorg:
bump... will get this fixed in the next release.

jdogresorg:
This issue should be fixed with PR #1222

Jpja:
Forum thread to discuss benefits, costs and risks of implementing this.

https://forums.counterparty.io/t/batch-sending-to-multiple-dispensers/6499

jdogresorg:
closing.. update will be in upcoming release 9.60.3

jdogresorg:
[...]

PR #1222, opened 2023-02-22
Multiple dispenses: now a single tx can trigger one or many dispensers

pataegrillo:


jdogresorg:
Does this ignore the last output which is typically the `change` output?

We don't want to accidentally trigger a "dispense" on `change` returning to an address where a dispenser is setup.

IMO an output should be considered a `change` output, and ignored, if the `output` address matches any `input` address.

Thoughts?

jdogresorg (approved):

Jpja:
I have several concerns, some still unanswered,
https://github.com/CounterpartyXCP/counterparty-lib/issues/1148

The main issue being the cost of dispensing converging to zero. Everything else in Counterparty have implied BTC fees or explicit XCP fees. This is a significant tradeoff and should go through the CIP process.

NACK

PR #1226, opened 2023-03-03
Issuance backwards compatibility

pataegrillo:
- New message type for new issuance format
- Old issuance format (with callability parameters) will be parsed as well as the new one


jdogresorg:
pls reopen in different pull request which depreciates parsing of OLD format (no one issues in that format anymore)

PR #1228, opened 2023-03-17
Send change smaller than DUST to miners fee instead of error

pataegrillo:
This fix adds a new parameter dust_size to the backend utxo sort function in order to change it from DEFAULT_MULTISIG_DUST_SIZE to DEFAULT_REGULAR_DUST_SIZE depending if the tx uses a "multisig" encoding or not

### DIFF
--- a/counterpartylib/lib/transaction.py
+++ b/counterpartylib/lib/transaction.py
@@ -158,7 +158,12 @@ def construct_coin_selection(encoding, data_array, source, allow_unconfirmed_inp
                 filtered_unspent.append(output)
         unspent = filtered_unspent
 
-        unspent = backend.sort_unspent_txouts(unspent)
+        if encoding == 'multisig':
+            dust = config.DEFAULT_MULTISIG_DUST_SIZE
+        else:
+            dust = config.DEFAULT_REGULAR_DUST_SIZE     
+            
+        unspent = backend.sort_unspent_txouts(unspent, dust_size=dust)
         logger.debug('Sorted candidate UTXOs: {}'.format([print_coin(coin) for coin in unspent]))
         use_inputs = unspent
 
@@ -252,7 +257,7 @@ def select_any_coin_from_source(source, allow_unconfirmed_inputs=True, disable_u
     unspent = filtered_unspent
 
     # sort
-    unspent = backend.sort_unspent_txouts(unspent)
+    unspent = backend.sort_unspent_txouts(unspent, dust_size=config.DEFAULT_REGULAR_DUST_SIZE)
 
     # use the first input
     input = unspent[0]
--- a/counterpartylib/lib/transaction.py
+++ b/counterpartylib/lib/transaction.py
@@ -187,7 +187,8 @@ def construct_coin_selection(encoding, data_array, source, allow_unconfirmed_inp
 
 
     # pop inputs until we can pay for the fee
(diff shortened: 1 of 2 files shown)


jdogresorg:
To clarify what this PR is for (since I failed to create a github issue before Javier and I started tackling this issue).

CP has had a persistent annoying issue for years where users are unable to spend all the BTC in their wallet addresses, and CP throws an error indicating users need more BTC, even when they in fact have enough BTC.

This is why quite often we are telling users to use electrum to gather their dust and send any remaining BTC out of their addresses 

image

I was finally able to track this issue down and repeat it, and Javier discovered that the issue was related to 2 issues :

    A. filtering utxos that were less than `DEFAULT_MULTISIG_DUST_SIZE` (7800)
    B. BTC change greater than zero but less than `DEFAULT_REGULAR_DUST_SIZE` (546) 

The solution for A is simple, we adjusted the utxo filtering to use `DEFAULT_REGULAR_DUST_SIZE` (546) so that all spendable utxos are usable by CP.

I believe the most simple solution for B is to simply tack any remaining BTC (greater than zero but less than 546 sats) on to the tx miners fee. This would allow CP to generate the transaction without generating a false BTC balance error.

However, before we move forward with this solution, I would like to get some feedback from the community.... Since this change slightly modifies the behavior from what the user is expecting (user specifies one BTC miners fee, but the actual BTC miners fee on the TX is a bit higher as it contains the extra sats)

Thoughts?

counterpartylib/lib/transaction.py
```diff
@@ -158,7 +158,12 @@ def construct_coin_selection(encoding, data_array, source, allow_unconfirmed_inp
                 filtered_unspent.append(output)
         unspent = filtered_unspent
 
-        unspent = backend.sort_unspent_txouts(unspent)
+        if encoding == 'multisig':
+            dust = config.DEFAULT_MULTISIG_DUST_SIZE
+        else:
+            dust = config.DEFAULT_REGULAR_DUST_SIZE     
+            
+        unspent = backend.sort_unspent_txouts(unspent, dust_size=dust)
```
jdogresorg, line 166: I don't think we need to change utxo filtering based on if tx is multisig or not... what we care about is getting spendable utxos to use as inputs, and using DEFAULT_REGULAR_DUST_SIZE (546) will return all usable utxos to use as inputs.

I know for multisig the outputs need to be at least `DEFAULT_MULTISIG_DUST_SIZE` (7800), but Is there a reason we only want to use inputs greater than 7800 with multisig?

jdogresorg:
Counterparty does not keep track of address balances, addrindexrs does that... and the issue here is NOT one of balances being out of sync.

Issue is if CP tries to generate a transaction and the amount of change left after the transaction is less than DUST, then it throws an error, as it does not know how to handle the remaining change. The solution which you feel is "hacky" is to simply pass this change (which you can't return to yourself in a spendable output) on to the miner and avoid throwing an error entirely.

jdogresorg:
Looks good in testing.

SuzukiMatt:
Can someone explain how to do fix this to a beginner? Im new to the space and trying to figure out how to use my left over btc on CP. But I dont understand the lingo and whatnot and not the most computer savvy but trying to learn. Can anyone explain what I do exactly in Lehman's terms, please? Thank you for any advice.

PR #1233, opened 2023-05-10
Now mempool parsing gets interrupted when a new block is mined

pataegrillo:


jdogresorg:
Looks good in testing. 

Been running this update on XChain.io for the past 24 hours with no unexpected behavior.

need to remove `MEMPOOL_TXCOUNT_UPDATE_LIMIT` from config.py since it is no longer used

PR #1244, opened 2023-05-28
Multisig dust fix

jdogresorg:
It has recently been discovered, that due to a misleading summary, Counterparty transactions have been paying 7800 sats for years instead of 780 in multisig transaction outputs. 

The original post in question indicated a multisig dust cost of 0.00000780, However in the summary, the value was changed to 0.0000780

0.00000780 = 780 sats = Correct
0.00007800 = 7800 sats = Incorrect

https://bitcointalk.org/index.php?topic=528023.msg7469941#msg7469941

🤦 I dont want to even think about how much BTC this has burned over the years... thank goodness those multisig outputs are spendable, so ppl can reclaim all these 7800 outputs which should have been 780.

This PR adjust the multisig dust level down to 1000 sats, which is reasonable, but not the minimum. We can probably find that minimum and verify it with additional testing, however I feel it is important to get this PR merged ASAP to reduce multsig dust encoding fees by a factor of almost 10x


Jpja:
Setting the dust this low will effectively burn outputs. It has always been the intention that multisig outputs should be redeemed. A high default dust (like 7800 sats) gives an incentive to do this. Redeeming UTXOs reduces the UTXO set, which is good for Bitcoin nodes – including Counterparty of course. The default should reflect this. 

If stamps or freewallet want to override the defaults, let it be on them.

NACK

https://github.com/CounterpartyXCP/cips/discussions/93
https://github.com/CounterpartyXCP/counterparty-lib/commit/71ff5f0b953237cd72012c111ec102954aa170f0#commitcomment-119462529

Issue #1243, opened 2023-05-22
Dividends issue with `reset` flag


jdogresorg:
It seems that when trying to do a dividend on an asset which has had its divisibility changed via the `reset` flag, an error is thrown due to being unable to accurately determine divisibility (code looks up first issuance, not latest `reset` issuance).

Need to get fix in place so that `dividends` work for `assets` with `reset` supply/divisibility.

This will require a protocol change and activation block in `protocol_changes.json`

Screenshot of CP API throwing `zero dividend` error 
![image](https://github.com/CounterpartyXCP/counterparty-lib/assets/376028/8dc3488c-ee8c-46d8-8b36-a166954fe8b2)

Screenshot of holders of COLLECTOR token
![telegram-cloud-photo-size-1-5656623885231171517-y](https://github.com/CounterpartyXCP/counterparty-lib/assets/376028/ec3c637c-25cd-4a0c-8790-659d01eac17b)

jdogresorg:
closing... fix will be in 9.60.3 release

PR #1257, opened 2023-09-24
Fixed an bug when trying to create a dividend for an asset whose divisibility was changed with a reset

pataegrillo:


jdogresorg:
This PR fixes issue #1243 where dividends are unable to be issued for assets that have had divisibility changed via a reset.

jdogresorg:
looks good in testing.

PR #1260, opened 2023-10-04
- Fixed numeric asset leading zero issuances not throwing an error when token exists

pataegrillo:


jdogresorg:
Looks good in testing. Now an error is thrown instead of a transaction which will fail.

image

jdogresorg (approved):

PR #1261, opened 2023-10-09
- Added a protocol change to btc balance checking for dispensers opened from another address

pataegrillo:


jdogresorg:
doing reparse after protocol change to see if this resolves the ledger checksum mismatch.

PR #1265, opened 2023-10-26
Fixed can't change description using lock flag

pataegrillo:


jdogresorg:
Looks good in testing

```
sqlite> select block_index, locked, description from issuances where asset='A18120456150883946182';
2536389|0|This is a test
2536390|1|changed description
```

PR #1276, opened 2023-11-13
CIP31 - Enhanced File Encoding Support

pataegrillo:


jdogresorg:
This PR adds support for [CIP31 - Enhanced File Encoding](https://github.com/CounterpartyXCP/cips/blob/master/cip-0031.md) to counterparty-lib.

The primary enhancements in CIP31 are :
- Support for storing `content-type` in issuance and broadcast transactions (default = `text/plain`)
- Limit amount of data stored in database to 1024 characters
- Adds a `db_data` API flag  (default = `true`)
    - `true` = Retrieve data from Counterparty database
    - `false` = Retrieve data from Bitcoin transaction

CIP31 is `optional` and node operators who wish to continue storing ALL transaction data in the counterparty database can do so via a fednode flag.

CIP31 is enabled by default, in order to limit database bloat and keep the database small / fast.

jotapea:
https://github.com/CounterpartyXCP/counterparty-lib/blob/4dffc2309f539b6b6da3ebed6c1ade21cd48b4c9/counterpartylib/lib/messages/issuance.py#L481

I see this line by itself, and is shouting how broken the whole concept is.

Why does transaction parsing have to be dependent on anything other than what is written in Bitcoin? What are the `store_only_text` and `return_info` params doing in this function?

Adding a `data_type` to `issuances` is unnecessary. Keeping everything as standard text/plain UTF8 text is a positive in my view. Anyone wishing to add a data type, should just do it in a base64 [data url](https://developer.mozilla.org/en-US/docs/web/http/basics_of_http/data_urls) format.

No protocol change.

NACK

keyuno:
It fails to respond to the concerns raised in the [discussion.
](https://github.com/CounterpartyXCP/cips/discussions/109)

> CIP31 is optional and node operators who wish to continue storing ALL transaction data in the counterparty database can do so via a fednode flag.

Optional simply implies that it is not necessary.

NACK

Jpja:
From the [CIP31 discussion](https://github.com/CounterpartyXCP/cips/discussions/109) I am convinced that DB optimization is a better solution.  

NACK

pataegrillo:
> https://github.com/CounterpartyXCP/counterparty-lib/blob/4dffc2309f539b6b6da3ebed6c1ade21cd48b4c9/counterpartylib/lib/messages/issuance.py#L481
> 
> I see this line by itself, and is shouting how broken the whole concept is.
> 
> Why does transaction parsing have to be dependent on anything other than what is written in Bitcoin? What are the `store_only_text` and `return_info` params doing in this function?
> 
> Adding a `data_type` to `issuances` is unnecessary. Keeping everything standard, as text/plain UTF8 text, is a positive in my view. Anyone wishing to add a data type, should just do it in a base64 [data url](https://developer.mozilla.org/en-US/docs/web/http/basics_of_http/data_urls) format.
> 
> No protocol change.
> 
> NACK

As you know, the main value for counterparty are the asset identifiers that can be named or numerical, but also, these assets are, in many cases, linked to a multimedia resource such as an image, audio, video, etc. In CP, there is a desire to have that resource within the blockchain, so users began to include this data within the description. CIP31 was born to prevent that data from flooding the cp database and also to make things easier by indicating what type of data to expect from those descriptions.

This is totally optional, if you don't like it or don't need it, you don't have to use it. Those params are needed because without the data already parsed we need to parse it on demand, specifically they are used this way:

store_only_text: allows you to configure an option to avoid storing data types other than text-plain on a CP node, this parameter will change accordingly with a installation config that is not already included, 
return_info: CP architecture forces any parsing to make changes to the database, this parameter allows retrieving the parsed message without doing any of those changes and it's used to retrieve the data part directly from the bitcoin node

jotapea:
@pataegrillo Thanks for explaining.

Still, in my opinion, the whole concept is out of place.

I believe Counterparty, the consensus (protocol), should not have anything to do with files. [Steganography](https://en.wikipedia.org/wiki/Steganography) will always exist.

`issuances` are financial assets.
`broadcasts` are... broadcasts!

Maybe in broadcasts "files" could make a bit more sense, but still not necessary at all. Data urls serve the same purpose perfectly. And quite ironically, your implementation of this feature is using a similar strategy!

As explained [here](https://github.com/CounterpartyXCP/cips/discussions/109#discussioncomment-7245038), the main problem with DB bloating is with the `issuances` table being too denormalized. A deeper redesign (there are alternatives) is what will solve this for the longer term.

jdogresorg:

Issue #1275, opened 2023-11-09
CRITICAL ISSUE: master branch handling


jotapea:
The `master` branch is getting changes merged without proper review by multiple developers (specially, federated node network peers). So, I believe we need to FOCUS on this CRITICAL issue (following up from https://github.com/CounterpartyXCP/counterparty-lib/pull/1256).

From my perspective of over a year observing protocol development activity, it seems @pataegrillo works on "orders" by @jdogresorg because his PRs have no descriptions (maybe is not "orders" but more of an agreement, but definitely these discussions are happening outside of the repo). Then, @jdogresorg merges these changes, effectively making a single person being responsible of the protocol code! I believe it is unacceptable (and dangerous!) for a supposedly decentralized protocol to have protocol affecting decisions made outside of the repository! Please let me know if I am incorrect in anything in this assessment. And more importantly, why the call to include more developers is being ignored.

And in a separate but VERY related note, the `master` branch is used by the [fednode](https://github.com/CounterpartyXCP/federatednode) software to [download the latest release](https://github.com/CounterpartyXCP/federatednode/blob/master/fednode.py#L337). Thus, this branch should not be an unstable non-consensus work in progress, instead it should be kept at the point of the latest stable "consensus"-agreed (we are not there yet, but assuming like it was) release (v.9.60.2) which is commit [09b20b3e](https://github.com/CounterpartyXCP/counterparty-lib/commit/09b20b3e1848481bed2fec3dba10c4cc11e5095e). 

Even if the current `master` branch is not being worked on in the way "I would [approach](https://github.com/CounterpartyXCP/counterparty-lib/pull/1270) it", please at least rename it. The new name could be something like `v9.60.3_rc_jj` (Release Candidate Jeremy Johnson).

Please don't ignore this critical issue. Or take it the wrong way, because my intention is to improve processes to have a better DECENTRALIZED network protocol.

adamkrellenstein:
Yep. `master` branch is now locked.

PR #1256, opened 2023-09-20
9.60.3 release

jdogresorg:
- Bumped bitcoin core version to 24.1
- Adjusted DEFAULT_MULTISIG_DUST_SIZE (7800->1000)
- Fixed issue with oracle function returning bad values
- Fixed issue with change smaller than DUST throwing error
- Fixed issue with issuances and callability error
- Fixed issue with issuance backwards compatability
- Fixed issue with dividends not working on reset assets
- Redefined EMPTY address to mean no XCP or BTC history
- Fixed issue with skipping owner validation on numeric issuances
- Fixed issue with sweeps not transferring asset ownership if no supply issued
- Added origin field to dispensers (allows refill/close/reopen from `origin`)
- Added support for triggering dispenses on all tx outputs
- Added max dispense limit of 1000

## Notes
Mainnet activation block is set to 825,000 for all protocol_changes in this release. Need to update release to set activation blocks 2 weeks out from release date, once closer to release.

PR #1280, opened 2023-11-18
v9.61.0 Release Candidate

jotapea:
CIP-Discussion: https://github.com/CounterpartyXCP/cips/discussions/124

An almost identical representation of work up to commit https://github.com/CounterpartyXCP/counterparty-lib/commit/fffdb01e966f43979e7f853f7bff224c663f0378. @jdogresorg Thanks for this work (and the [summary](https://github.com/CounterpartyXCP/counterparty-lib/commit/91c744e59fea3298ce265ea893ef738f04f5d714) with links!).

The only change in this PR focuses on one part of the [main issue](https://github.com/CounterpartyXCP/counterparty-lib/issues/1275) that needs URGENT action:

***CURRENT fednode installers will activate the changes at the VERY SOON block height. This is a BIG ISSUE, effectively ALREADY acting like a forced sudden protocol update!***

This PR handles part of the problem by setting the activation block way into the future as a placeholder. But what is left is not possible for me to change without repository contributor privileges, which is a simple rename of the current `master` branch. (And now, it should also include removal of the non-consensus [published release](https://github.com/CounterpartyXCP/counterparty-lib/releases/tag/v9.61.0).)

### DIFF
--- a/counterpartylib/protocol_changes.json
+++ b/counterpartylib/protocol_changes.json
@@ -417,42 +417,42 @@
       "minimum_version_major": 9,
       "minimum_version_minor": 60,
       "minimum_version_revision": 2,
-      "block_index": 819300,
+      "block_index": 850000,
       "testnet_block_index": 2412394
   },
   "issuance_backwards_compatibility": {
       "minimum_version_major": 9,
       "minimum_version_minor": 60,
       "minimum_version_revision": 2,
-      "block_index": 819300,
+      "block_index": 850000,
       "testnet_block_index": 2422000
   },
   "zero_balance_ownership_sweep_fix": {
       "minimum_version_major": 9,
       "minimum_version_minor": 60,
       "minimum_version_revision": 2,
-      "block_index": 819300,
+      "block_index": 850000,
       "testnet_block_index": 2505725
   },
   "dispenser_origin_permission_extended": {
       "minimum_version_major": 9,
       "minimum_version_minor": 59,
       "minimum_version_revision": 6,
-      "block_index": 819300,
+      "block_index": 850000,
       "testnet_block_index": 2499850
   },
   "multiple_dispenses": {
       "minimum_version_major": 9,
       "minimum_version_minor": 59,
       "minimum_version_revision": 6,
-      "block_index": 819300,
+      "block_index": 850000,
       "testnet_block_index": 2420996
   },
   "issuance_description_special_null": {
       "minimum_version_major": 9,
       "minimum_version_minor": 60,
       "minimum_version_revision": 2,
-      "block_index": 819300,
+      "block_index": 850000,
       "testnet_block_index": 2505725
   },
   "addrindexrs_required_version": {
@@ -460,7 +460,7 @@
       "1":{
         "value":"0.4.0"
       },
-      "819300":{
+      "850000":{
         "value":"0.4.1"
       }   
     },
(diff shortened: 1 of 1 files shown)


jdogresorg:
As has been explained to you a few times in a few different places... I am simply following the same procedure for picking activation block that Counterparty has been doing for years. The original founders themselves set activation blocks 2 weeks out from release, that is what is happening in the 9.61.0 release.... nothing surprising here.

A full parse IS NOT NECESSARY (and has been typically been done by CP devs BEFORE a release). If you choose to run your CP node and always do a full parse on every release, that is your right..... but you can not expect other node operators to want to spend a week in downtime doing a full parse, or even 24 hours doing a reparse. 

Quick Example: Exchanges on Counterparty integration? Downtime is a no-no for them... They want a simple list of commands to run and get back up ASAP, forced full parse or reparse are NOT viable options for them, nor for MOST node operators. 

The bootstrap is validated via a reparse BEFORE every release, so that the end user does not HAVE to do the reparse (unless they want, which they can easily do so at any time by running a command... not by being FORCED to)

IMO setting an placeholder block WAY into the future at 999,999 does no good in a release, as you have to then do yet ANOTHER release to actually activate the feature. 

Also, you had previously mentioned that all development work should take place on the `develop` branch. I agreed with you and said that after this 9.61.0 release, we would switch to using the `develop` branch and only merge to `master` branch at release time or in case of hotfixes, etc.

Since we will be using the `develop` branch going forward, please close all your existing pull request on `master` and reopen them on `develop`

I will be closing all active pull requests on `master` after activation block 819,300

jotapea:
Thanks for your time in explaining @jdogresorg.

About the 2 weeks activation block, please see https://github.com/CounterpartyXCP/cips/discussions/124#discussioncomment-7608087 and lets follow up there as that explains why this "original founders practice" needs to evolve to become a truly decentralized network.

The parsing argument is also on that thread, so maybe is better to also continue talking about that there. But I can say that I agree that it does make sense that some counterparty-lib node operators will find use for the bootstrap (there must be a reason why it exists).

About the activation block, I clearly specified it as a placeholder. This PR does not represent my endorsement of all the code changes. THIS RELEASE IS NOT CONSENSUS AGREED! Please @ mention other node operators that have reviewed the protocol changing code. I have, and it seems I will just have to list these (mostly ignored) interactions...

But time is limited and what I did here is FOCUSING on the main issue at hand. If you are already publishing a release, this is the place the resistance to this release needs to focus. I still believe this is not an appropriate release, it would have been if it was a non-protocol changing revision release, which was its appearance until a [VERY SHORT time ago](https://github.com/CounterpartyXCP/counterparty-lib/pull/1256#issuecomment-1782898370)!

Your strategy works to keep you in control. Minimize the importance of repository activity, ignoring whatever goes against your views. So, this PR was my way of trying to minimize disruption of your "general practices", that I believe are not decentralization friendly.

You have a TOO MUCH POWER in this supposedly decentralized protocol.

You are in control of the repo and what most "non-technical average Joe" users consider Couterparty = XChain+FreeWallet. You are the establishment, and act like don't want any competition. So, newer participants in the network struggle and I have been voicing these in "different places", but you just continue doing things "like always".

Do you really want Counterparty to be decentralized? Your actions continuously prove you don't.

jdogresorg:

PR #1281, opened 2023-11-25
Fixed primary keys not being created because of duplicate names

pataegrillo:


jotapea:
This is awesome @pataegrillo ! Most likely you haven't seen it, but I already worked on fixing these indexes on the CNTRPRTY repo.

https://github.com/CNTRPRTY/counterparty-lib/commit/a5ecfbe581f3463b3a03621c12d6f8938fa88da1

Notice I went a bit farther, removing the indexes I think aren't necessary. And then added others I think are (see dispenser).

But in my first full parse test, that took a while because I did it on mainnet (*), resulted in different consensus hashes. I thought this was not supposed to affect the consensus hashes, so I "forgot" about it (... or was still too salty to follow up here lol).

What do you think of that work? If you like it, should I make another PR or should I join this one with the adjustments?

Thinking this can be the main improvement for a new v9.60.3, and then we can cherry pick the non-consensus-affecting features from the current, [imminent consensus-war triggering v9.61](https://github.com/CounterpartyXCP/cips/discussions/124), which I hope doesn't happen.

(*) How do you test for consensus changes?

jdogresorg:

PR #1292, opened 2023-12-15
Fixed get_blocks function to work with non-correlated block_index due to delay in closing dispensers.

pataegrillo:
There is a problem introduced with the dispensers close delay. Everything works corectly except for the block_index in the message bindings, instead of inserting the block_index of the actual closure, the block_index of the original tx that marked the dispenser to be close is used. This is generating a non-correlated block_index in the messages table, so now it cannot be assumed that greater the messsage_index equal or greater is the block_index.

Without this fix, some queries to the get_blocks function (counterblock depends on this function) will result in many blocks having empty messages and eventually making counterblock to crash.

Along with this PR I created another one for Counterblock as a work around for this problem. Also, i'll create another PR to fix the dispenser close delay block_index.

### DIFF
--- a/counterpartylib/lib/api.py
+++ b/counterpartylib/lib/api.py
@@ -745,7 +745,7 @@ class APIServer(threading.Thread):
             # Packages messages into their appropriate block in the data structure to be returned
             for block in blocks:
                 block['_messages'] = []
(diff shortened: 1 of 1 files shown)


adamkrellenstein:
See https://github.com/CounterpartyXCP/counterparty-lib/issues/1294

PR #1293, opened 2023-12-15
Correct `block_index` Used for Dispensers Closing Delay

pataegrillo:
THIS FIX IS FOR REFERENCE ONLY!

This fixs the wrong block_index used when closing a dispenser with delay. With this fix there is no need to use the recent counterparty and counterblock PRs to fix the non-correlated block_index but a full parse from the last update block_index is needed and that will change the message ledger (but it won't change any credit/debit nor transactions).

This fix could be apply from a specify block_index but a protocol change is needed and those non-correlated block_index will remain in the message table and the fixes for counterparty and counterblock would be needed.

### DIFF
--- a/counterpartylib/lib/messages/dispenser.py
+++ b/counterpartylib/lib/messages/dispenser.py
@@ -596,7 +596,7 @@ def close_pending(db, block_index):
                 'source': dispenser['tx_source'],
                 'asset': dispenser['asset'],
                 'status': STATUS_CLOSED,
-                'block_index': dispenser['tx_block_index'],
+                'block_index': block_index,
                 'tx_index': dispenser['tx_index']
             }
             sql = 'UPDATE dispensers SET give_remaining=0, status=:status WHERE source=:source AND asset=:asset'


adamkrellenstein:
> a full parse from the last update block_index is needed 

How far back do you mean? Do you mean on counterblock, or will counterpartyd have to reparse?

pataegrillo:
Counterpartyd has to make a full parse since the activation of the dispenser closing delay, block 819300 (aprox 6200 blocks before the current one right now). But as I said, it will change the message ledger

adamkrellenstein:
> THIS FIX IS FOR REFERENCE ONLY!

Is something missing? :)

adamkrellenstein:
Gotcha, okay right. This seems like the correct solution. We'll do a full reparse and double-check that debits/credits etc. don't change. 👌

ouziel-slama:
If I understand correctly the only consequence of this modification is the content of the message table.
The `block_index` field is now automatically updated by the `ledger.insert_update()` function and the `messages` table as well as the logs must be corrected and cleaned. see https://github.com/CounterpartyXCP/counterparty-lib/issues/1400 and https://github.com/CounterpartyXCP/counterparty-lib/issues/1445.

adamkrellenstein:
Okay, I made this a line-item in https://github.com/CounterpartyXCP/counterparty-lib/issues/1445

PR #1295, opened 2023-12-23
Invalidate Broadcasts with Malformed Text

pataegrillo:
I don't know why there wasn't a single broadcast with a malformed text before, but counterparty testnet crashed because the length marked in the first byte was not the same as the length of the whole text. The crash occurs due to an assertion. With this change, if that assertion is not valid, then the broadcast gets rejected as a "invalid: could not unpack text"


adamkrellenstein:
Wait, the network will crash if someone triggers this?

pataegrillo:
Yes, counterparty testnet crashed because of that, so it can be replicated on mainnet

adamkrellenstein:
Uhh, can we merge this and release it as an urgent hotfix??

jdogresorg:
> Wait, the network will crash if someone triggers this?

YES! I pointed this out to Joe when it crashed testnet... didn't want to draw TOO much attention to the issue, since I wasn't in a position to push forward a hotfix release myself without it being controversial.

I DID include this fix in my 9.62.0 fork version... its a critical fix which should be addressed immediately IMO... but, maybe wait a lil bit on release, cuz MAYBE community gonna want some other changes (like XCP fee on numerics).

adamkrellenstein:
IMO these kinds of issues should be dealt with ASAP. Like, point release in <24 hours with a mandatory upgrade for soon after (a week?). I would recommend we keep this as a hotfix and completely independent from any other patches to maximize speed and safety.

jdogresorg:
> IMO these kinds of issues should be dealt with ASAP.

I 100% agree, and if I was in the same position a year ago or more, when I discovered issues like this, Javier and I have put out hotfix ASAP.... unforunately, @jotapea has made pushing forward ANYTHING a PITA through his non-stop attacks on any issue, idea, suggestion, pull request, merge, etc... I breathe, or use the wrong word somewhere, there is a post from @jotapea about it on github.... not trying to cause more drama, just linking to a single example of what i've had to deal with non-stop while trying to work on, test, fix, and push out releases for CP.

So, yes, I agree this is an issue that should have been fixed ASAP... but, dev process was so broken at the time, that even critical hotfixes like this could not be pushed through without a fight... I ran out of energy to keep fighting... sorry.

jotapea:
The repo speaks for itself.

From 2023-09-20, less than 4 months ago, recent, this was the proposed REVISION release of @jdogresorg.

https://github.com/CounterpartyXCP/counterparty-lib/pull/1256

jotapea:
Anyway, originally came to say that https://github.com/CNTRPRTY/counterparty-lib has this hotfix applied already. So [xcp.dev](https://xcp.dev) is safe.

There is [another](https://github.com/CNTRPRTY/counterparty-lib/commit/7d93a2a2df04af49d14da00c336669972957c5ce) similar one with invalid destroys. Also already applied at CNTRPRTY/counterparty-lib, which I was going to make a PR but wanted to deal with this first.

https://github.com/CounterpartyXCP/counterparty-lib/issues/1296

adamkrellenstein:
Ah great! @jotapea would you mind making that PR here? Totally agreed we need to clean that branch mess up, though...

jotapea:
@adamkrellenstein Done. https://github.com/CounterpartyXCP/counterparty-lib/pull/1321

adamkrellenstein:
awesome, thanks!

pataegrillo:
I must warn something about the invalid destruction fix. Typically, people don't destroy an asset by creating the tx manually. I came across this error many times and it means that some ledger has an asset that the ledger with the error does not have, some issuance had an error in this last ledger. If this is true (and it always has been for me), then it's an invisible snowball because one mistake creates a cascade of mistakes.

adamkrellenstein:
@pataegrillo I'm sorry I don't understand... can you explain more?

pataegrillo:
Of course, there are two cases in which an invalid destruction could appear: 

1) someone wrote a wrong code to create a destroy tx with an invalid asset without the help of counterparty api (because counterparty api could detect the wrong asset), I didn't detect a case like this. 

2) some user, using an app, selected to destroy an existing asset (obviously, an app wouldn't ask you to put any token name you desire to destroy, it would be absurd). In that ledger, the one the app uses, the asset does exists, but on another ledger without the asset the error occurs. This would mean that the ledger without the asset failed to read the issuance that created the asset in first place for some reason. That reason could be that the user didn't have enough XCP because some dispenser didn't trigger in that ledger and on and on and on...

So, until now and if no one creates an invalid destruction transaction on purpose, if some invalid destruction appears it's a sign of something going wrong blocks ago.

jotapea:
I can understand the point @pataegrillo makes, this has been used to identify issues elsewhere. But is still an attack vector that will break nodes. The fix will have to be done eventually, more so if we want to allow users to (optionally) build transactions without validation (to reduce node load).

pataegrillo:
Just to clarify, I didn't say it shouldn't be fix it, I just wanted to warn about what it means to receive that error, because it usually means a disaster has been brewing for several blocks ago.

adamkrellenstein:
Tests forthcoming!

PR #1298, opened 2024-01-06
Add 0.10 XCP fee on numerics

jdogresorg:
This pull request puts a 0.10 XCP fee on numeric assets and activates on block 829,020.

# Activation Logic
```
 144 blocks/day
x 30 days (1 month)
---
4,320 blocks

824,700 current block
+ 4,320 blocks
---
829,020 = Activation Block (1 month)
```

### DIFF
--- a/counterpartylib/lib/messages/issuance.py
+++ b/counterpartylib/lib/messages/issuance.py
@@ -245,7 +245,10 @@ def validate (db, source, destination, asset, quantity, divisible, lock, reset, 
                     # subasset issuance is 0.25
                     fee = int(0.25 * config.UNIT)
                 elif len(asset) >= 13:
-                    fee = 0
+                    if util.enabled('numeric_asset_fee'):
+                        fee = int(0.10 * config.UNIT)
+                    else:
+                        fee = 0
                 else:
                     fee = int(0.5 * config.UNIT)
             elif block_index >= 291700 or config.TESTNET or config.REGTEST:     # Protocol change.
--- a/counterpartylib/protocol_changes.json
+++ b/counterpartylib/protocol_changes.json
@@ -459,7 +459,7 @@
       "minimum_version_major": 9,
       "minimum_version_minor": 61,
       "minimum_version_revision": 1,
-      "block_index": 829020,
+      "block_index": 824888,
       "testnet_block_index": 2571199
   },
   "addrindexrs_required_version": {
--- a/counterpartylib/protocol_changes.json
+++ b/counterpartylib/protocol_changes.json
@@ -455,6 +455,13 @@
       "block_index": 819300,
       "testnet_block_index": 2505725
   },
+  "numeric_asset_fee": {
+      "minimum_version_major": 9,
+      "minimum_version_minor": 61,
+      "minimum_version_revision": 1,
+      "block_index": 829020,
+      "testnet_block_index": 2571199
+  },
   "addrindexrs_required_version": {
     "mainnet":{
       "1":{


jotapea:
NACK

What is this? A PR for a protocol change and activation all at the same time?

I highly recommend you reconsider and withdraw this PR. If not, then be prepared for taking FULL responsibility of forking the protocol.

jdogresorg:
> NACK

no surprise, not sure i've ever seen an ACK from you to anything  I propose.. lol

> What is this? 

Again, would know if you were in the dev channel where all the devs talk about issues n not limiting yourself to just github interactions.

> ... be prepared for taking FULL responsibility ...

fully prepared bro. Here is the release that xchain.io is now running on 👍 

https://github.com/jdogresorg/counterparty-lib/releases/tag/v9.62.0

Jpja:
This is in line with CIP29 from May last year - https://github.com/CounterpartyXCP/cips/blob/master/cip-0029.md

However, having skimmed 1000+ messages over the weekend from the Telegram dev channels, this certainly is a heated topic. In the worst case we'll find Xchain on 9.62 while significant parts of the community remain on 9.61. I am not sure the benefits from a fee on numerics are worth the fork risk.

Also, two of the original founders recently joined the chat. From what I understand, they are leaning to DB optimization as a better solution. I am therefore **against the fee** at this point, given that a CIP and/or code to replace CIP29 will be written (that details how DB optimization can be implemented).

adamkrellenstein:
Hi, all! Yep, as I said on Telegram, it seems like the primary motivation for this proposed protocol change is actually just to workaround a simple technical problem in the backend of xchain.io which has made it challenging to support and maintain that site. IMO we should do what we can to make it easier for devs to build explorer-like functionality on top of `counterpartyd`. See #1299 and #1300 .

pappyg45:
Fully support adding this fee in fact I think it should be increased to .25XCP

jotapea:
The reasons for adding a fee are misled. And the approach taken by @jdogresorg goes against the social contract, and 100% against decentralization.

[20 blocks notice](https://github.com/CounterpartyXCP/counterparty-lib/pull/1298/commits/ab6cd5b6826ec2b3b0885160cedf53ac1a5eb69c) to activate a protocol change? WTF?

This is a full on attack to Counterparty as a brand, XCP as a token, and the rest of the assets. What is the truth when there are 2 ledgers? What is Counterparty then? See: https://github.com/CounterpartyXCP/cips/discussions/124#discussion-5865143.

dikafradana:
i  support adding this fee, I think it should be increased to 0.5 xcp. i dont want see on xchain explorer get spammed more.

jotapea:
@dikafradana xchain can hide any data they wish. No need to change the protocol.

Are you aware that this was activated without consensus? Meaning that right now, xchain IS NOT the counterparty-lib protocol?

Even if you agree with the fee, THE APPROACH TAKEN was hostile and completely unnecessary. It is an attack to Counterparty. Everyone gets affected by this situation.

Nobody wins, and everyone loses. And it won't be evident until more time passes, but your tokens' value will also lose.

pappyg45:
> The reasons for adding a fee are misled. And the approach taken by @jdogresorg goes against the social contract, and 100% against decentralization.
> 
> [20 blocks notice](https://github.com/CounterpartyXCP/counterparty-lib/pull/1298/commits/ab6cd5b6826ec2b3b0885160cedf53ac1a5eb69c) to activate a protocol change? WTF?
> 
> This is a full on attack to Counterparty as a brand, XCP as a token, and the rest of the assets. What is the truth when there are 2 ledgers? What is Counterparty then? See: [CounterpartyXCP/cips#124 (comment)](https://github.com/CounterpartyXCP/cips/discussions/124#discussion-5865143).

The reason for adding this fee are completely in compliance with natural law. JDog has put all the work into this protocol and his "proof of work" is all the reason he needs to stop the spammers and grifters from taking advantage of his time and energy.

heunland:
I'm in favor of adding the XCP fee to numerics, as well as addressing the DB optimization issues asap.

0cam:
How does the counterparty community show they are for or against this? Is it by the number of registered accounts saying yes or no? On the surface it looks like three or so devs need to hit a majority agreement for anything to happen regardless of the issue.

jotapea:
@0cam maybe reading the "[documented](https://github.com/CounterpartyXCP/cips/discussions/124)" process for the previous release will help you understand where we are. Is not good, but there was at least some CLEAR attempt to improve it.

For this one specifically, it can be considered a hostile attack because there was no wait for consensus.

First doing a PR with an activation block already established (and deployed), with a "I don't care if you accept it or not, is happening" attitude was wrong.

But then made 100x worse by suddenly deciding to activate with 20 blocks notice.

adamkrellenstein:
So if we do want to add an XCP fee to numerics, it seems like 0.1 XCP isn't going to be enough. A STAMP today already costs about US$10 in fees, and I think we'd have to increase the total price (in fiat equivalent) by a factor of at least 5 to have a significant effect on issuance volumes. At current market prices, that's around 10 XCP per issuance, or 100 times greater than the current proposal. Then, after only 100,000 more issuances (less than 1/3) the total we've had so far, we'd have already burned 1/2 of the XCP in existence. But maybe we want it to cost >US$50 to issue an asset, and maybe we're okay with extremely rapid deflation??

I suggest we table this PR for now while we collectively work on addressing performance and scalability at the implementation level. Once we've got some benchmarks in place, some initial optimizations and some quality-of-life improvements for block-explorer hosts, that might be a better time to revisit this issue.

Screenshot 2024-01-13 at 12 13 34 AM

keyuno:
> So if we do want to add an XCP fee to numerics, it seems like 0.1 XCP isn't going to be enough. A STAMP today already costs about US$10 in fees, and I think we'd have to increase the total price (in fiat equivalent) by a factor of at least 5 to have a significant effect on issuance volumes. At current market prices, that's around 10 XCP per issuance, or 100 times greater than the current proposal. 

There is no sound rationale for assuming that the value of XCP will remain constant in a deflationary environment where XCP is becoming more desirable. 

I also suggest considering the introduction of an XCP fee for all substantial actions, extending beyond numerics. This approach could serve to reduce the risk of potential [forks](https://github.com/CounterpartyXCP/cips/discussions/129#discussioncomment-8100434).

adamkrellenstein:
I'm fine with discussing implementing a much more general and dynamic gas system.

However I'm just saying that this PR is not the place for that conversation. (I know we all want to have it though! :))

keyuno:
> I'm fine with discussing implementing a much more general and dynamic gas system.
> 
> However I'm just saying that this PR is not the place for that conversation. (I know we all want to have it though! :))

> I'm fine with discussing implementing a much more general and dynamic gas system. However I'm just saying that this PR is not the place for that conversation. (I know we all want to have it though! :))

I appreciate your perspective. In my initial response, I shared a link to the 'Enhancing the XCP Ecosystem' thread, aiming to spark a discussion on this topic and ensure thorough documentation on GitHub.

adamkrellenstein:
Closing this PR per my above comments. Would love to re-open the discussion of an XCP fee on numerics as part of a larger conversation around cryptoeconomics, and then only after the codebase is in a more stable place.

Issue #1302, opened 2024-01-12
Update Docs to Recommend Anaconda with Python 3.9, No Root Install


adamkrellenstein:
See https://github.com/CounterpartyXCP/counterparty-lib/issues/1301

adamkrellenstein:
We're going to go with Hatch.

Issue #1301, opened 2024-01-12
Installation Breaks with Python 3.10 (Ubuntu 22.04 LTS)


adamkrellenstein:
This bug in upstream libraries, in particular: https://stackoverflow.com/questions/70943244/attributeerror-module-collections-has-no-attribute-mutablemapping

There are other issues though with unpinned dependencies.

* `Werkzeug==2.2.2` (https://stackoverflow.com/questions/77213053/why-did-flask-start-failing-with-importerror-cannot-import-name-url-quote-fr)
* `itsdangerous==2.0.1` (https://stackoverflow.com/questions/71189819/importerror-cannot-import-name-json-from-itsdangerous)
* `Flask-0.11.1` (https://stackoverflow.com/questions/71645272/importerror-cannot-import-name-markup-from-jinja2)
* `PyCrypto` per https://github.com/easybuilders/easybuild-easyconfigs/issues/17410

Need to update our pegged dependency versions and possibly throw a warning if Python 3.10 is used.

Also, need `setuptools<58` per https://github.com/pycrypto/pycrypto/issues/327

adamkrellenstein:
Wait a minute, is this a Python 2 codebase??

adamkrellenstein:
@ouziel-slama is moving straight to Python 3.11 ☺️

PR #1304, opened 2024-01-12
Update Python Dependencies; Revise Installation Instructions [WIP]

adamkrellenstein:
See https://github.com/CounterpartyXCP/counterparty-lib/issues/1302 and https://github.com/CounterpartyXCP/counterparty-lib/issues/1301

We shouldn't merge this until we get the tests fully passing... too many things could break.


adamkrellenstein:
Obviated by #1322

PR #1322, opened 2024-01-14
Python 3.11 + Hatch + Latest version for all dependencies + Fix tests

ouziel-slama:
- [x] Replace `setup.py` with `pyproject.toml`.
- [x] `counterparty-lib` requires now Python >= 3.11.
- [x] Using the latest versions for all dependencies.
- [x] Update `circle.yml` and `README.md`
- [x] Fix regular tests.
- [x] Replace `--skiptestbook` by `--testbook`
- [x] Add `--testbookblockindex` to test book only from the given block 
- [x] Temporarily deactivates the test book
- [x] Fix https://github.com/CounterpartyXCP/counterparty-lib/issues/1134

To install the library just use `pip install .` or `pip install -e .` in editable mode.

You can also use hatch to run test in a venv.
```
$ hatch run pytest counterpartylib/test/ --skiptestbook=all
```

PR #1323, opened 2024-01-14
Tweak README

adamkrellenstein:
This will need to be revised further based on https://github.com/CounterpartyXCP/counterparty-lib/pull/1322.


adamkrellenstein:
Obsolete with #1322

PR #1349, opened 2024-01-22
Rollback

ouziel-slama:
- [x] Remove undolog.
- [x] Refactor `reparse` functions using the `block_index` field.
- [x] Refactor `rollback` function using the `block_index` field.
- [x] DRY `SELECT`s and `UPDATE`s queries in all contracts.
- [x] put all these queries in `ledger.py` into own functions.
- [x] Add `block_index` field in `balances` table
- [x] Remove UPDATEs in balances.
- [x] Remove UPDATEs in orders.
- [x] Remove UPDATEs in bets.
- [x] Remove UPDATEs in rps.
- [x] Remove UPDATEs in dispenser.
- [x] Refactor `held()` and `holders()` functions.
- [x] Add test for `check.asset_conservation()`.
- [x] DRY and update API.
- [x] Add test for `holders()` function.
- [x] Lock UPDATE in all tables.
- [x] Optimize queries in contracts.
- [x] Clean indexes (remove useless, add missing)
- [x] Write SQL migrations
- [x] Debug and fix (need to pass all `testnet` checkpoints, test well new `reparse` and `rollback`).


adamkrellenstein:
Amazing to delete so much code! ❤️

ouziel-slama:
> Amazing to delete so much code! ❤️

good that you see the good side of things :))

ouziel-slama:
Sorry I spoke too quickly.. it's true that a full reparse no longer makes much sense, but from a certain index, assuming that there is no change of protocol in the creation of "dispenser", it remains interesting. I'm going to clean up and restore the reparse... and for the test book we can do a reparse on the last 3 checkpoints, for example, to stay within reasonable times.

adamkrellenstein:
Sounds good!

jotapea:
This changes the `balances` table from an updateable present-view to an append-only ledger. The holistic implications of this should be studied more thoroughly imo.

This [query](https://github.com/CounterpartyXCP/counterparty-lib/pull/1349/files#diff-57c5c4a17f8859a3ca8ff02337ee1844b9c91a059615dfee127568a94afe891aR746) in particular is concerning:

```sql
SELECT asset, SUM(quantity) AS total FROM (
    SELECT (address || asset) AS aa, address, asset, SUM(q) AS quantity FROM (
        SELECT (address || asset) AS aa, address, asset, SUM(quantity) AS q FROM credits GROUP BY aa
        UNION ALL
        SELECT (address || asset) AS aa, address, asset, -1 * SUM(quantity) AS q FROM debits GROUP BY aa
    ) GROUP BY aa
) GROUP BY asset
```

This would be called for *every* asset in *EVERY* block in the [asset_conservation](https://github.com/CounterpartyXCP/counterparty-lib/blob/e65172a6700a6a45207120ca082da2673d8b240c/counterpartylib/lib/check.py#L152) step. So, it seems like raw transaction parsing is going to be slowed down quite considerably by this.

And from a Counterparty application developer's perspective, this will require EVERYONE to update how they deal with balance queries... this is NOT a trivial change!

ouziel-slama:
> A query that would be called for _**every**_ asset in _**EVERY**_ block in the [asset_conservation](https://github.com/CounterpartyXCP/counterparty-lib/blob/e65172a6700a6a45207120ca082da2673d8b240c/counterpartylib/lib/check.py#L152) step. So, it seems like raw block parsing is going to be slowed down quite considerably.

The additional time this new query takes compared to the old one is really negligible compared to the total time to parse a block. Furthermore, this check is not really part of the block parsing and it is only executed when the node is up to date, that is to say when it is waiting for a new block.

> And from a Counterparty application developer's perspective, this will require EVERYONE to update how they deal with balance queries.

Yes, the game is worth it. And we can possibly avoid it by updating the API so that this modification is transparent.

jotapea:
I don't want to step over your shoes, but we do need to consider alternatives, and discuss them with other stakeholders. There must be other options with different tradeoffs...

> Yes, the game is worth it. And we can possibly avoid it by updating the API so that this modification is transparent.

This is true for people relying on the api. Some don't, as it is way too inefficient for some types of applications. E.g. xcp.dev would not be possible if I was using the api. I build my own custom api and queries to make it possible. And any other "power" application would do the same.

jotapea:
The [query](https://github.com/CounterpartyXCP/counterparty-lib/pull/1349/files#diff-57c5c4a17f8859a3ca8ff02337ee1844b9c91a059615dfee127568a94afe891aR746) is kind of rebuilding the `balances` for all assets, per block. Why not take the other route and rebuild the balances table once, only when necessary?

Maybe the only new column needed is a "last updated" block. And with this the balances table can continue acting as a view, which is clear that was its original intent and design. Forcing it into a ledger seems forced and over-engineered.

Lets just discuss it, and if I am wrong, I should be convinced that your approach is the best.

jotapea:
> No one should do this, and expect compatibility with future versions. The API is the only official way to query the database and Counterparty cannot guarantee backwards compatibility outside of the API. No software does this and it's normal. It makes no sense to guarantee backwards compatibility for the internal architecture of software.

This makes perfect sense.

But that is why I categorized mine as a power application, which has deeper integration into the protocol database to be able to perform. I'm definitely a "minority" case, but it doesn't have to be, and I would argue these kinds of integrations are what can elevate Counterparty based apps!

The beauty of Counterparty is that it IS just a database. If more developers knew this, it would take the protocol to another level imo. I see the "consume only through the api" approach as a negative honestly. I would explicitly show both: simpleApps=>api, powerApps=>sql.

Database schemas should rarely be updated for a mature protocol like this. Not that I am against change for its own sake, I can be convinced about changes if these are worth it.

And I do understand that not having updates in the Counterparty database makes sense. But that is precisely is why `balances` stood out. I don't know... I just think there can be another way to achieve the same result and maintain the current expectations.

Can you at least elaborate why a "last updated block" column wouldn't work?

fractastical:
You could create a diff with the existing structure in a format that
guarantees backward compatibility but also allows for new variables.

jotapea:
@fractastical Thanks for your suggestion. I can update my queries, that is not my main concern. I just want to bring in my perspective, and would also like to understand if there are alternatives...

ouziel-slama:
> I would explicitly show both: simpleApps=>api, powerApps=>sql.

sorry I can't agree with that. powerApps == api + cache in a consolidated and personalized database and possibly client/server. No one should do SQL queries directly in the database but use the API (in fact 10 years ago I added an endpoint to do SQL queries and I think today that it was a very bad idea)

> Database schemas should rarely be updated for a mature protocol like this.

it's true, that's why this table hasn't been touched for years and years ;-))

> And I do understand that not having updates in the Counterparty database makes sense. But that is precisely is why balances stood out. 

This is precisely why it must be modified as a priority ;) As a bonus, we keep the history of the balance, we have more information to debug and check the integrity.

>  I just think there can be another way to achieve the same result and maintain the current expectations.
> Can you at least elaborate why a "last updated block" column wouldn't work?

I didn't say it couldn't work.. just that it was cleaner to add a `block_index` field.

adamkrellenstein:
Agreed with @ouziel-slama . I think this is a great change, that will dramatically improve the architecture of the reference implementation. (Agreed about that API endpoint for SQL queries! I think that was my idea. 😬) Making the Counterparty database fully log-structure will have significant benefits for the correctness and performance of rollbacks + reparses. It is also critical that we have well-defined APIs, and that users use them *or* simply patch their software when the implementation changes, as it would here. That's what APIs are for. If the APIs aren't working, then we should change *that*.

jotapea:
We see things differently.

Counterparty being a Bitcoin in SQL platform is a PLUS! Powerful.

The ONLY drawback to using SQL directly is having to keep up with schema changes. And that is why precisely I am here discussing it. And this should not happen frequently, even less as it matures.

Treating Counterparty like any traditional application does a disservice to it. My opinion.

jotapea:
Looking forward to the new api queries. If this PR is going to represent a transparent replacement, these must be included.

loon3:
> (in fact 10 years ago I added an endpoint to do SQL queries and I think today that it was a very bad idea)

If you're referring to the get_{table} call its the only realistic way to build performant wallet UIs.  It is what I use for 90% of my API calls.  Empirical data suggests having it was in fact a very good idea.

adamkrellenstein:
I fully believe that it's been the only way to make performant UIs so far! but I'm also very confident that we'll be able to fix that 💪

loon3:
I would like nothing more than better, faster APIs but having the ability to create custom calls if nothing exists for my specific application is certainly not bad idea, far from it.  We can't predict how the platform will be used so always having that ability supported in the API is essential.

ouziel-slama:
@Io

> > (in fact 10 years ago I added an endpoint to do SQL queries and I think today that it was a very bad idea)
> 
> If you're referring to the get_{table} call its the only realistic way to build performant wallet UIs. It is what I use for 90% of my API calls. Empirical data suggests having it was in fact a very good idea.

nope about this one: https://github.com/CounterpartyXCP/counterparty-lib/blob/master/counterpartylib/lib/api.py#L557 (@adamkrellenstein hum.. sorry I think this one it's my bad idea lol)

yeah `get_{table}` was designed to be the main API endpoint but today I reread it and I don't find it elegant, it assumes that the user knows the database schema, which in theory he is not supposed to have to learn. I would be curious to know which queries you use most often. 
In any case, now that it has been around for so long, I think we have to do our best to keep it compatible with previous versions. And if we remove it one day, then with a long depreciation period.

ouziel-slama:
> Looking forward to the new api queries. If this PR is going to represent a transparent replacement, these must be included.

fair enough ;)

jotapea:
> it assumes that the user knows the database schema, which in theory he is not supposed to have to learn

I fundamentally disagree with this. So I don't have much more to say, this severely limits its potential.

My docs would show both, an api for some simple use cases, and the schema for serious developers.

jotapea:
> Add block_index field in balances table

This should be its own PR. There are too many changes here, and I think the `balances` table nature change warrants it being on its own. I would have made it a CIP actually... but missing that, then at least lets make it its own PR.

adamkrellenstein:
I don't think it's worth the time to tease out the balances changes into a separate PR (it's up to @ouziel-slama though). Really, there's no reason for all of the FUD around such a straightforward change to a simple database schema. Some things should be controversial—reducing tech debt can't be one of them, esp. given the enormous amount of tech debt there is to deal with.

jotapea:
Why not? Is the absolute least that can be done for a substantial schema change.

Characterizing my feedback as FUD is quite a leap. And not cool. I am just sharing my concerns as someone actually RUNNING Counterparty infrastructure.

The changes to the `balances` table are to make it consisntent with the rest of the tables to be able to rollback, reparse etc. And I understand the benefits of this.

But, in normal node running circumstances, this change makes ALL balance read queries slower. None will be faster and saying otherwise is straight up false.

adamkrellenstein:
Because I've responded in detail to each one of your concerns and the goalposts keep moving:

1. It's a "substantial schema change" -> It is, in fact, a single additional, integer column on an existing table.
2. It's going to make your life harder. -> You literally just need to add `ORDER BY block_index DESC LIMIT 1` to your queries.
3. It's going to affect the whole community. -> The community was notified and was extremely supportive of the change.
4. The database schema is sacrosanct. -> A DB schema is an implementation detail, yes, even for a blockchain. APIs exist for a reason, and even those can change when there's cause. If you want to bypass the API, see (2).
5. "it seems like raw block parsing is going to be slowed down quite considerably" -> Building the table is about 4x faster.
6. Why not just add a *different* column to the table? -> Because it would provide none of the architectural benefits...
7. It's super complicated. -> It makes it possible to reduce the size of the codebase by over 6,000 lines LOC.

If you have real concerns about a degradation read performance for your application, that's fine, but _please provide benchmarks that you would like us to optimize for_. (Empty open issue: https://github.com/CounterpartyXCP/counterparty-lib/issues/1359) Breaking this out into a separate PR isn't going to have an effect on performance. If you want to do some code review, please refer to [the commit in question](https://github.com/CounterpartyXCP/counterparty-lib/pull/1349/commits/acb1949d4660af5cd855b85eeb675f021e4cd801).

jotapea:
Thanks for the summary.

I just think is fair to ask to split PRs into their logical components. Having a separate PR that focuses only on the changes throughout the app required for the schema change will show the implications of the change clearly, from the positives to the negatives.

Is not the best precedent to treat huge PRs like a completely acceptable and normal procedure. Splitting the PR will be a bit more work, yes, but is the correct thing to do here.

jotapea:
Agree. You can see my concerns to many of the recent (last <2 years) changes throughout the repo ;)

ouziel-slama:
@jotapea I'll look into trying it but at this stage it seems complicated. I've already extracted from this PR everything that wasn't directly related to adding `block_index`. I understand your concern about the size of the PR and that is exactly why it will remain in WIP as long as:

1) the checkpoints updated on `develop` are definitively confirmed.
2) that this PR passes all testnet and mainnet checkpoints.

Moreover, it is true that there is a lot of code that is modified but nothing really complex (compared for example to the code to parse/compose the transactions).. you especially have to be meticulous.. I do my best and count on your meticulous review ;-).

jotapea:
@ouziel-slama Please. Separating the new `balances` schema into its own PR is all I ask now. While I still think my arguments are valid, I am not going to resist it, mainly because this is not a protocol change.

And it will help make this PR smaller and more manageable. It seems like this one is like your own internal "develop" with many interrelated changes lol. Which is fine considering the "end product", but in terms of repository organization (and history for future studies), we should split PRs as much as possible.

adamkrellenstein:
sorry @jotapea, we don't do development here by personal plea. please focus on technical matters and write those benchmarks that you're so worried about.

jotapea:
@adamkrellenstein Respectfully, you are not setting the best example here.

Did I include personal plea in my message, yes and maybe that was a mistake it seems. But I have also been presenting technical arguments from the first message (which was a query), specifying that my concern is specifically in READs, and finally focusing on the bigger picture of the repo code organization.

Btw, benchmarks are only relevant when comparing in isolation... which can ONLY be done if the PR is split!

If this change is really not such a big deal, I see no reason why it cannot be its own PR. Lets be productive and not add fuel to an argument unnecessarily. You know my position, I know yours. But is @ouziel-slama's decision at this point.

adamkrellenstein:
Incredible work!!! 👏

Of course this is a huge PR, but it's a logical unit of work and there was obviously a massive amount of architectural refactoring necessary to fix this critical functionality.

consensus_error.txt
```diff
@@ -0,0 +1,109 @@
+Incorrect ledger_hash hash for block 316000.  Calculated 50e6c784b35badf8ac41c5d933f5bff881f09569b09a99eb48947f7050808734 but expected 5c4506b2863161af45ef7f4afa0fe8730826a0e671cbdb3d9ecfde93d370c972
```
adamkrellenstein, line 1: Can we add some comments to indicate the context for these errors? (Also this should probably not be here.)

counterpartylib/lib/blocks.py
```diff
@@ -490,6 +444,13 @@ def initialise(db):
                       timestamp INTEGER)
                   ''')
 
+    # Lock UPDATE on all tables
```
adamkrellenstein, line 447: ❤️

counterpartylib/lib/blocks.py
```diff
@@ -1236,6 +1034,35 @@ def list_tx(db, block_hash, block_index, block_time, tx_hash, tx_index, tx_hex=N
     return tx_index
 
 
+def clean_table_from(cursor, table, block_index):
```
adamkrellenstein, line 1037: Super elegant!

counterpartylib/test/util_test.py
```diff
@@ -825,3 +734,56 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             self.mock_protocol_changes[k] = self._before[k]
         for k in self._before_empty:
             del self.mock_protocol_changes[k]
+
+def reparse(testnet=True, checkpoint_count=10):
```
adamkrellenstein, line 738: Slick :)

tools/benchmark.py
```diff
@@ -0,0 +1,327 @@
...
+import os
+import time
+import difflib
+
+import apsw
+
+
+MAX_INT = 2**63 - 1
+
+def remove_from_balanc_old(db, address, asset, quantity):
```
adamkrellenstein, line 13: Typo

adamkrellenstein (approved):

counterpartylib/lib/backend/__init__.py
```diff
@@ -110,7 +109,6 @@ def fee_per_kb(conf_target, mode, nblocks=None):
     return BACKEND().fee_per_kb(conf_target, mode, nblocks=nblocks)
 
 
-@functools.cache
```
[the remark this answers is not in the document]
ouziel-slama: yeah I tested the Python cache but it's not better if not worse.. I don't think the size is the problem but the serialization to store it in the cache.. `bitcoinlib` returns classes .. probably with `dict` it would be faster.. I would look at this more closely by cleaning `get_tx_info()`.

consensus_error.txt
ouziel-slama: oops! it's just a not relevant debug file! will remove it!

tools/benchmark.py
ouziel-slama: fixed!

JavierCervilla:
This is huge @Ouziel big thanks for your huge work really glad to have you all back guys!!

PR #1358, opened 2024-01-26
[WIP] Move `counterparty-cli` into `counterparty-lib`

ouziel-slama:
This is the very first step. We now need to create a `counterparty-lib` folder and move everything into it.. etc.
(merged with `--allow-unrelated-histories` to preserve commit history)

### DIFF
--- a/counterparty-cli/CONTRIBUTING.md
+++ b/counterparty-cli/CONTRIBUTING.md
@@ -1,0 +1,27 @@
+# Security Issues
+
+* If you’ve identified a potential **security issue**, please contact us
+  directly at <support@counterparty.io>.
+
+
+# Reporting an Issue
+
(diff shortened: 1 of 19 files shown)


jotapea:
> (merged with --allow-unrelated-histories to preserve commit history)

Very cool! This is NOT what I did for [CNTRPRTY/counterparty-lib](https://github.com/CNTRPRTY/counterparty-lib/commit/630bc4f251e5eeef921029816f9bd7c29acc0453) lol. Good to know it won't be a requirement for future versions of the repo fork (which is essentially a couple of additional commits I continuously rebase).

ouziel-slama:
I will open a new PR on the Counterparty repository, it will be simpler than a rebase etc.

PR #1407, opened 2024-02-13
Clean `gettxtinfo()`

ouziel-slama:


ouziel-slama:
OK there is still room for cleaning, and this version needs to pass all the checkpoints, but it is now clean enough to be able to concentrate on optimization.

adamkrellenstein (approved):
so so so much better! ❤️

PR #1412, opened 2024-02-15
Use 'pycoin_rs' for base58 encoding and decoding

ouziel-slama:
From the tests:
```
counterpartylib/test/pycoin_rs_test.py::test_pycoin_rs 
rust duration for 400K encodes and 400K decodes:  1.0097870826721191
python duration for 400K encodes and 400K decodes:  4.263357400894165
PASSED
```

profile16

on this profile of a mainnet kickstart we see, without real surprise, that base58 encoding and decoding consumes a lot of time..

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,8 +17,6 @@ jobs:
           command: |
             pip install --upgrade pip
             pip install hatch
-            curl https://sh.rustup.rs -sSf | sh -s -- -y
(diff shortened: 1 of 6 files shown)


adamkrellenstein (approved):
Nice! Looks like roughly 55% of the time is spent on this one set of operations, which is now >4x faster. :)

Issue #1429, opened 2024-02-19
Pickle Error when Connecting to LevelDB


adamkrellenstein:
```
[2024-02-19 10:27:05][WARNING] Warning:
- `/Users/adam/Library/Application Support/counterparty/counterparty.db` will be moved to `/Users/adam/Library/Application Support/counterparty/counterparty.db.old` and recreated from scratch.
- Ensure `addrindexrs` is running and up to date.
- Ensure that `bitcoind` is stopped.
- The initialization may take a while.
Proceed with the initialization? (y/N) : y
[2024-02-19 10:27:15][INFO] Last known block hash: None
[2024-02-19 10:27:15][INFO] Acquiring lock.
[2024-02-19 10:27:15][INFO] Connecting to database (SQLite 3.44.2.0).
[2024-02-19 10:27:15][ERROR] SQLITE_LOG: no such table: blocks in "SELECT * FROM blocks WHERE block_index = (SELECT MAX(block_index) from blocks)" (1) SQLITE_ERROR
[2024-02-19 10:28:39][INFO] Blocks indexed in: 83.690s
[2024-02-19 10:28:39][ERROR] Unhandled Exception
Traceback (most recent call last):
  File "/opt/homebrew/bin/counterparty-server", line 8, in <module>
    sys.exit(server_main())
             ^^^^^^^^^^^^^
  File "/Users/adam/Documents/work/Counterparty/Repos/counterparty-cli/counterpartycli/__init__.py", line 18, in server_main
    server.main()
  File "/Users/adam/Documents/work/Counterparty/Repos/counterparty-cli/counterpartycli/server.py", line 172, in main
    server.kickstart(
  File "/Users/adam/Documents/work/Counterparty/Repos/counterparty-lib/counterpartylib/server.py", line 506, in kickstart
    kickstarter.run(bitcoind_dir=bitcoind_dir, force=force, last_hash=last_hash, resume_from=resume_from)
  File "/Users/adam/Documents/work/Counterparty/Repos/counterparty-lib/counterpartylib/lib/kickstart/__init__.py", line 160, in run
    block_parser = BlockchainParser(bitcoind_dir, memory_db, last_parsed_block)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1432, opened 2024-02-19
Try fork method

ouziel-slama:


ouziel-slama:
fix https://github.com/CounterpartyXCP/counterparty-lib/issues/1429 on macos

PR #1437, opened 2024-02-20
Replace `bitcoinlib` by  `block_parser` to deserialize transaction

ouziel-slama:
Simpler, faster (uses dicts instead of classes) and well tested with kickstart. This will allow for cleaner code by using the same functions everywhere and improve the kickstart speed by 10% (this is approximately the CPU time that the small function which transforms a `dict` into a `class` now takes)

profile33


ouziel-slama:
```
Time to deserialize 1000000 transactions with block_parser: 17.764070749282837 seconds
Time to deserialize  1000000 transactions with bitcoinlib: 21.92827606201172 seconds
```
while waiting rust :) will be normally easier to convert with dict than classes also.

ouziel-slama:
profile35

looks better!

adamkrellenstein (approved):

PR #1444, opened 2024-02-22
Pass checkpoint 825000

ouziel-slama:


adamkrellenstein:
Should we implement a check to prevent this from happening again? Obv. would be a protocol change if it's on the backend.

We could also do it in the API as a temporary measure.

ouziel-slama:
This should no longer happen with the new `script_to_address()` function.. on the other hand we probably need to hardcode a similar fix somewhere to be sure we can use the assets on these addresses.

ouziel-slama:
Merging but:
- see https://github.com/CounterpartyXCP/counterparty-lib/issues/1452
- we must add checkpoints before the next release.

PR #1458, opened 2024-02-28
Tweak Kickstart CLI

ouziel-slama:


ouziel-slama:
- Remove `--resume-from` and `--last-hash` arguments
- Automatically retry connecting to `addrindexrs` instead of throwing an assertion error.
- Auto-resume if possible
- Create new database from scratch only if version < 10
- It is now possible to alternate `kickstart` and `start` on the same database.
- Don't show block parsing logs unless `--verbose`
- Clean keyboard interruption
- Spinners and colors

Capture d’écran 2024-02-29 à 12 34 38

counterpartylib/lib/database.py
```diff
@@ -128,6 +130,9 @@ def get_connection(read_only=True, foreign_keys=True, integrity_check=True):
             raise exceptions.DatabaseError('Could not perform integrity check.')
         logger.info('Integrity check completed.')
 
+    cursor.execute('PRAGMA auto_vacuum = 1')
+    cursor.execute('PRAGMA journal_size_limit = 0')
```
adamkrellenstein, line 134: this seems very low, esp. when setting it not just for kickstart but for everything. can we 1) set it higher and/or 2) set it to a different value juts for kickstart?

counterpartylib/lib/kickstart/__init__.py
```diff
@@ -111,74 +167,116 @@ def run(bitcoind_dir, force=False, last_hash=None, resume_from=None, max_queue_s
...
 
     # backup old database
-    if os.path.exists(config.DATABASE) and resume_from is None:
-        os.rename(config.DATABASE, config.DATABASE + '.old')
+    if not new_database and not resuming:
+        user_version = cursor.execute('PRAGMA user_version').fetchall()[0][0]
+        version_major = user_version // 1000
+        if version_major < 10:
+            if not force:
+                print(colored(f"Version lower than 10.0 detected. Kickstart must be done from the first block.", "yellow"))
```
adamkrellenstein, line 204: `v10.0.0`

adamkrellenstein (approved):

counterpartylib/lib/database.py
ouziel-slama: hum.. the default is -1 (no limit) and 0 means the "always truncate rollback journals and WAL files to their minimum size".. not sure what the best.. I don't see how to decide on an arbitrary value otherwise..

counterpartylib/lib/kickstart/__init__.py
ouziel-slama: fixed, thanks!

counterpartylib/lib/database.py
adamkrellenstein: okay, it seems we only need this when using `PERSIST` mode, which we're just doing for `kickstart`, I think.

I think in that case we should set the `journal_size_limit` to `0` only for `kickstart`

PR #1459, opened 2024-03-01
Pass checkpoint 825000

ouziel-slama:


counterpartylib/lib/backend/addrindexrs.py
```diff
@@ -548,3 +550,75 @@ def init():
 def stop():
     if 'Indexer_Thread' in globals():
         Indexer_Thread.stop()
+
+
+# Basic class to communicate with addrindexrs.
```
adamkrellenstein, line 555: What's the plan for using this? It looks like it doesn't replace the existing method of communicating...

adamkrellenstein (approved):

counterpartylib/lib/backend/addrindexrs.py
ouziel-slama: For now this class make a lot of assumptions rights for `get_oldest_tx()`.. it's a good base to replace the current client but definitely needs more work and testing.. especially on the side of the API.. and it also depends on when/how we replace `addrindexrs`.
Opened https://github.com/CounterpartyXCP/counterparty-lib/issues/1460

PR #1472, opened 2024-03-05
Fix Pylint and Bandit alerts

ouziel-slama:


ouziel-slama:
I greatly underestimated the work to pass all the Pylint checks.. I merge and I opened this issue so as not to delay the next release.

PR #1488, opened 2024-03-10
Replace warning by debug when a dispenser tx is not valid

ouziel-slama:


adamkrellenstein:
uh, what is this crazy logic? isn't the loop itself problematic??

PR #1503, opened 2024-03-12
Fixes

ouziel-slama:


ouziel-slama:
- [x] Update current block index after a rollback
- [x] Rebuild database on full rollback
- [x] Use `ripemd-hash` (to fix openssl problem with Python 3.10)
- [x] Remove generic except when parsing multisig tx

PR #1523, opened 2024-03-16
Restore Test Book

ouziel-slama:


ouziel-slama:
1. mock `addrindexrs` when running `book_test.py`.
2. reparse 5 checkpoints
3. In github workflow: bootstrap testnet database then run `book_test.py`

.github/workflows/test_book.yml
```diff
@@ -0,0 +1,40 @@
...
+        sudo apt-get update -y
+        sudo apt-get install -y libgirepository1.0-dev libleveldb-dev
+        python -m pip install --upgrade pip
+        pip install pylint maturin pytest
+        cd counterparty-rs && pip install -e . && cd ..
+        cd counterparty-lib && pip install -e . && cd ..
+        cd counterparty-cli && pip install -e . && cd ..
+    - name: Bootstrap testnet database
+      run: |
+        counterparty-server --backend-password=rpc --testnet bootstrap
```
adamkrellenstein, line 36: isn't `rpc` the default backend pass? (i.e. do we need to pass it here explicitly?)

.github/workflows/test_book.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Run tests
+      run: |
+        cd counterparty-lib
+        pytest counterpartylib/test/book_test.py --testbook -s
```
adamkrellenstein, line 40: roughly how long does this take? should we be doing this on every push or nightly or on PRs...

adamkrellenstein (approved):

.github/workflows/test_book.yml
ouziel-slama: yep! I restored the default password!

.github/workflows/test_book.yml
ouziel-slama: it's only 3m 46s.. around the same time as the "Build, Install and Run test".

PR #1524, opened 2024-03-16
Restore 'check.software version()' but only every 24 hours

ouziel-slama:


adamkrellenstein:
Hmmm this is a good thing, but I think we should make it a *random* block every 144 blocks.

PR #1530, opened 2024-03-18
Accept config args before and after the command

ouziel-slama:


adamkrellenstein (approved):
nice! let's just also include it in the Release Notes

PR #1535, opened 2024-03-19
Update Dockerfile and `compose.yml`

ouziel-slama:


ouziel-slama:
- Add entrypoint in the Dockerfile
- Use default value when possible

simplenode/compose.yml
```diff
@@ -11,7 +11,7 @@ services:
       # - "-chain=test"                       # testnet
       - "-rpcallowip=0.0.0.0/0"
       - "-rpcbind=0.0.0.0"
-      - "-rpcuser=rpc"
+      - "-rpcuser=bitcoinrpc"
```
adamkrellenstein, line 14: ahh I wanted to standardize on `rpc:rpc` _everywhere_... can we change it on the other end?

simplenode/compose.yml
```diff
@@ -55,19 +55,12 @@ services:
...
       # "--testnet"                           # testnet
       - "--backend-connect=bitcoind"
-      - "--backend-user=rpc"
-      - "--backend-password=rpc"
       - "--indexd-connect=addrindexrs"
-      - "--rpc-user=rpc"
-      - "--rpc-password=rpc"
       - "--rpc-host=0.0.0.0"
-      - "--log-file=1"
-      - "--api-log-file=1"
```
adamkrellenstein, line 69: these are on by default?

adamkrellenstein (approved):

simplenode/compose.yml
ouziel-slama: yep since the v10!

PR #1543, opened 2024-03-21
`SimpleNode` is just Docker Compose 🙄

adamkrellenstein:


ouziel-slama:
Do we really need the sub folders with config file ?

adamkrellenstein:
I think we need the `bitcoind` one? but the counterparty config files can probably just be examples in the docs

ouziel-slama:
> I think we need the `bitcoind` one? but the counterparty config files can probably just be examples in the docs

hum.. I think we need it only if we want to maintain a `counterparty/bitcoind` image with this default configuration else the parameters in `compose.yml` should be enough.

adamkrellenstein:
Ah yes, I forget why I did that! I think I just wanted people to be able run the bitcoind in docker without docker compose?? but there's probably no reason for that.

We can get rid of our bitcoind image.

PR #1548, opened 2024-03-22
Test and tweak `compose.yml`

ouziel-slama:


ouziel-slama:
Bitcoind, Addrindexrs and Counterparty data are stored in `~/counterparty-core-data` unless `$COUNTERPARTY_CORE_DATA` is set. So you can do:

````
$ COUNTERPARTY_CORE_DATA=/my/big/external/disk docker-compose up
````

File: README.md [hunk 1]
```diff
@@ -13,12 +13,26 @@ The simplest way to get your Counterparty node up and running is to use Docker C
...
+Then run node services in background with:
 
 ```bash
 git clone git@github.com:CounterpartyXCP/counterparty-core.git
 cd counterparty-core
-docker-compose up
+docker-compose up -d
+```
+
+**To run a node you must have at least 1.5Gb free.** By default all data is stored in the `~/counterparty-core-data` folder. You can modify this folder with the environment variable `$COUNTERPARTY_CORE_DATA`. For example:
```
adamkrellenstein, line 24: Tb?

README.md
```diff
@@ -13,12 +13,26 @@ The simplest way to get your Counterparty node up and running is to use Docker C
...
+
+```bash
+COUNTERPARTY_CORE_DATA=/var/data docker-compose up -d
```
adamkrellenstein, line 27: I think maybe call this COUNTERPARTY_DOCKER_DATA

adamkrellenstein (approved):
nice! much more elegant

README.md
ouziel-slama: oops! fixed!

README.md
ouziel-slama: yep!

ouziel-slama:
it's normally finished but I'm waiting to do a complete catch-up on my mac before merging..

ouziel-slama:
I also removed  `ADDRINDEXRS_JSONRPC_IMPORT=true` and shared `bitcoind` service volume with `addrindexrs` service.

compose.yml
```diff
@@ -60,10 +62,10 @@ services:
         source: ${COUNTERPARTY_DOCKER_DATA:-~/counterparty-docker-data}/
         target: /data
     ports:
-      - "4000:4000"                           # mainnet
-      - "14000:14000"                         # testnet
+      - "4000:4000"                                        # mainnet
+      - "14000:14000"                                      # testnet
     command:
-      - "--chain=${NETWORK:-main}"            # export NETWORK=test for testnet
+      - "--chain=${COUNTERPARTY_NETWORK:-main}"            # export COUNTERPARTY_NETWORK=test for testnet
```
adamkrellenstein, line 68: hm, I prefer the old style of `--testnet`/`--regtest` etc. that's also what Ordinals does, fwiw.

unfortunately I think that bitcoind uses tons of antipatterns in its CLI and API :/

ouziel-slama: The old style still works! you cas use --testnet/--regtest and that what we could document everywhere.
It's really just to have a really simple `compose.yml`.. It can't hurt to have an extra argument just for that imho..

ouziel-slama: ok found a solution without the flag `--chain` ;-) see https://github.com/CounterpartyXCP/counterparty-core/pull/1548/commits/2b0c2d13b004282119e3d5bd8dd412ce5fba3841

ouziel-slama:
1. Add 3 environments variables for Docker Compose:

   - `COUNTERPARTY_NETWORK` (`main` or `test`, by default `main`)
   - `COUNTERPARTY_DOCKER_DATA` (by default `~/counterparty-docker-data`)
   - `ADDRINDEXRS_JSONRPC_IMPORT` (`true` or `false`, by default `false`)

1. Update README.

1. Test `compose.yml` with Github Actions and GCP. A `testnet` node is continually running on GCP. At each push on the `develop` branch the following test is performed:

    - stopping the node with `docker compose stop`
    - download the new code to the GCP VM
    - building a new Docker image
    - deletion of the Counterparty database
    - restarting the node with `docker compose up`
    - waiting for the `/healthz` endpoint to return a code 200.

Notes:
- the duration of the test depends on the age of the bootstrap database
- If there is already a test in progress the test stops with an error and must be restarted later (manually or with another push).

compose.yml
```diff
@@ -1,14 +1,16 @@
...
+      - data:/bitcoin/.bitcoin
     ports: 
-      - "8332:8332"                           # mainnet
-      # - "18332:18332"                       # testnet
+      - "8332:8332"                                        # mainnet
+      - "18332:18332"                                      # testnet
     command: 
-      - "-chain=main"                         # mainnet
-      # - "-chain=test"                       # testnet
+      - "-chain=${COUNTERPARTY_NETWORK:-main}"             # export COUNTERPARTY_NETWORK=test for testnet
```
adamkrellenstein, line 13: can we have `COUNTERPARTY_NETWORK` be either `testnet` or `mainnet` and then do `${COUNTERPARTY_DOCKER_DATA:0:4}` for bitcoind CLI compatibility?

compose.yml
```diff
@@ -21,50 +23,56 @@ services:
...
       - bitcoind
       - addrindexrs
     volumes:
+      - data:/root/.bitcoin
       - data:/data
     ports:
-      - "4000:4000"                           # mainnet
-      # - "14000:14000"                       # testnet
+      - "4000:4000"                                        # mainnet
+      - "14000:14000"                                      # testnet
```
adamkrellenstein, line 59: is it possible to do this as an inline Bash conditional based on the value of `COUNTERPARTY_NETWORK`?!

adamkrellenstein (approved):

compose.yml
ouziel-slama: hum.. I don't think.. we don't have all the base funcitonnality here.. we could just have a default value and pass 1400 when testnet.. IMO is not a problem to expose a unused port here.. if really needed security should be done with a firewall on another level..

ouziel-slama:
Temporary using `get_running_info`, I will update with `/healthz` when https://github.com/CounterpartyXCP/counterparty-core/pull/1560 will be merged.

mattmarcello:
lgtm

Issue #1486, opened 2024-03-09
[WIP] Redesign Logs


adamkrellenstein:
## Human-Readable (Default for `STDOUT`)
```
2024-03-09-T13:08:10+0000 - INFO - Block 833899 - 2.00s [L:114c0 / TX:78a2b / M:d2e00]
2024-03-09-T13:08:10+0000 - INFO - Block 833899 - Broadcast (775b93) [valid]
2024-03-09-T13:08:10+0000 - INFO - Block 833899 - Send KEKUI (0e6394) [valid]
2024-03-09-T13:08:10+0000 - INFO - Block 833899 - Issuance 50 A344381935640106602 (d3654c) [valid]
```

## Semantic (Default for Log File)
`{"event": "broadcast", "block": 833897, "block_timestamp": "2024-03-09T12:24:00+00:00", "module": "counterpartylib.lib.blocks", "log_level": "debug", "asset": "A884346988877074888", "source": "bc1qpvmcqcx99cemaj5f9gls70q7tn29gr72qaf0zz", "fee": 0, "tx_hash": "8f68a9866ec7ecfe33dacfcd9f1baed7e10a7c0dfa024366d072eba16f955472", "status": "valid"}`

## Notes
* always log verbose to the log files

adamkrellenstein:
Also don't throw exceptions for routine events:
<img width="1289" alt="Screenshot 2024-05-22 at 9 32 28 PM" src="https://github.com/CounterpartyXCP/counterparty-core/assets/6189829/86937f0e-b60a-4d9f-b196-5bf7a0894fc3">

adamkrellenstein:
https://github.com/CounterpartyXCP/counterparty-core/pull/1827#pullrequestreview-2085976773

PR #1550, opened 2024-03-23
Clean Messages

ouziel-slama:


ouziel-slama:
Updates on this PR:

1) all INSERTs are done systematically from the `ledger.insert_record()` function (with the exception of the `messages` and `balances` tables and for the mempool).

2) all updates are done systematically from the `legder.insert_udpate()` function (with the exception of the `balances` and `transactions` tables).

3) `ledger.insert_record()` and `legder.insert_udpate()` take a required argument: `event`. This argument must be an uppercase character string. For example `OPEN_ORDER` or `SEND`. See `log.py` for full list of events.

4) Added an `event` field to the `messages` table.

5) From these two functions we call the function `ledger.add_to_journal()` (formerly `log.message()`) which adds a row to the `message` table.

6) From the `ledger.add_to_journal()` function we call the `log.log_event()` function which logs the event translated into English unless the `--json-log` flag is used.

7) removal of the spaghetti function `log.log()`.

The content of the `messages` table has most likely changed a little for certain events but remains essentially identical and compatible with old versions.

From the `messages` table it is possible to completely rebuild the Counterparty database. (BTW I think that to make an explorer like xcp.dev the best method is to only monitor this table with the API and rebuild a database optimized for search and display needs).

The logs now also correspond exactly to the contents of the `messages` table.

ouziel-slama:
Also removed the spaghetti function `database.exectracer()`

adamkrellenstein:
This is great! A major cleanup and refactor! ❤️


> 4. Added an `event` field to the `messages` table.

Let's mention this in the release notes.


> 6) From the `ledger.add_to_journal()` function we call the `log.log_event()` function which logs the event translated into English unless the `--json-log` flag is used.

So I was thinking that we'd *always* log both JSON and human-readable logs to files and then always log human-readable to the console. (If a dev wants JSON logs, he can just `tail` a file.) See https://github.com/CounterpartyXCP/counterparty-core/issues/1486


> The content of the `messages` table has most likely changed a little for certain events but remains essentially identical and compatible with old versions.

So really great if you could provide a few examples of the "before" and "after". Ideally we'd even provide dumps of the tables from old and new DBs up to a certain block height so that people can see exactly what changed (because they incorrectly depend on it :/)

Will the hashes change?


> From the `messages` table it is possible to completely rebuild the Counterparty database. (BTW I think that to make an explorer like xcp.dev the best method is to only monitor this table with the API and rebuild a database optimized for search and display needs).

Very cool. Should we plan to build a test for this?

This is the kind of thing we should highlight in our documentation, btw! (I'm not sure that downstream apps shouldn't use the API for building out their own DBs... I'm not sure if we want to commit to having to support the `messages` table schema. What do you think/)


> The logs now also correspond exactly to the contents of the `messages` table.

❤️

ouziel-slama:
> Let's mention this in the release notes.

done!


> So I was thinking that we'd _always_ log both JSON and human-readable logs to files and then always log human-readable to the console. (If a dev wants JSON logs, he can just `tail` a file.) See #1486

hm.. why would we store the exact contents of the `messages` table in a file? you can get it easily with the API..

 
> So really great if you could provide a few examples of the "before" and "after". Ideally we'd even provide dumps of the tables from old and new DBs up to a certain block height so that people can see exactly what changed.

hum... it's complicated... there are especially more lines in the message table (which was not systematically filled with each SQL query). Rows that were already present should not have any fields missing.

> Will the hashes change?

only the `message_hash` which is not and has never been verified..

> 
> Very cool. Should we plan to build a test for this?

yep! it is not a priority imho but yes it is better to have more tests than fewer..


> These architectural features are the kind of thing we should highlight in our documentation, btw! 

yep!

> (I'm not sure that downstream apps shouldn't use the API for building out their own DBs... I'm not sure if we want to commit to having to support the `messages` table schema. What do you think?)

I am sure that downstream apps should not directly use the counterparty API to display the pages of their sites. This is definitely bad practice. Downstream apps MUST have their own database optimized for their use case and the API must only populate their own database.
I am also sure that we must provide a simple and quick method so that downstream apps can easily maintain their database. This is the purpose of the `messages` table.

adamkrellenstein:
> > So I was thinking that we'd _always_ log both JSON and human-readable logs to files and then always log human-readable to the console. (If a dev wants JSON logs, he can just `tail` a file.) See #1486
> 
> hm.. why would we store the exact contents of the `messages` table in a file? you can get it easily with the API..

Ahh what about logs that don't have to do with consensus? Like for polling retries, or errors like "out of disk space", or... ?

I'd expect the logs to be a strict superset of the `messages` table, but not identical to it...


> > So really great if you could provide a few examples of the "before" and "after". Ideally we'd even provide dumps of the tables from old and new DBs up to a certain block height so that people can see exactly what changed.
> 
> hum... it's complicated... there are especially more lines in the message table (which was not systematically filled with each SQL query). Rows that were already present should not have any fields missing.

Okay so we can promise 100% backwards compatibility? If, so, we need to say that explicitly in the release notes: "New fields and additional rows only".


> > Will the hashes change?
> 
> only the `message_hash` which is not and has never been verified..

Okay yeah we should still document this.


> > Very cool. Should we plan to build a test for this?
> 
> yep! it is not a priority imho but yes it is better to have more tests than fewer..


> > (I'm not sure that downstream apps shouldn't use the API for building out their own DBs... I'm not sure if we want to commit to having to support the `messages` table schema. What do you think?)
> 
> I am sure that downstream apps should not directly use the counterparty API to display the pages of their sites. This is definitely bad practice. Downstream apps MUST have their own database optimized for their use case and the API must only populate their own database. I am also sure that we must provide a simple and quick method so that downstream apps can easily maintain their database. This is the purpose of the `messages` table.

Hm, I don't like the idea that a downstream app would use `messages` instead of the (future) REST API... that's the problem we have with the current API design—the implementation details are exposed.

If we recommend people read directly from `messages`, then we won't ever be able to change it (much).

ouziel-slama:
> Ahh what about logs that don't have to do with consensus? Like for polling retries, or errors like "out of disk space", or... ?
> 
> I'd expect the logs to be a strict superset of the `messages` table, but not identical to it...


For the moment all these unexpected errors are logged in the classic way and not in JSON. It's a deeper refactoring to do this and outside the scope of this PR whose main goal is to clean up several spaghetti functions that are no longer maintainable.


> Okay so we can promise 100% backwards compatibility? If, so, we need to say that explicitly in the release notes: "New fields and additional rows only".

yep!
 
> Hm, I don't like the idea that a downstream app would use `messages` instead of the (future) REST API... that's the problem we have with the current API design—the implementation details are exposed.
> 
> If we recommend people read directly from `messages`, then we won't ever be able to change it (much).

I'm not saying that you have to directly read the `messages` table. The ideal is of course to use the future REST endpoint `/journals/` or even better imo `/events`.
The idea is not to expose the structure of the base but all the values necessary to be able to build your own base perfectly adapted to your needs. `sqlite` is a database for a client application and optimized for parsing and maintaining consensus not for a web application. Directly using the Counterparty API to build a website is bad practice and makes no sense.

adamkrellenstein:
> > Ahh what about logs that don't have to do with consensus? Like for polling retries, or errors like "out of disk space", or... ?
> > I'd expect the logs to be a strict superset of the `messages` table, but not identical to it...
> 
> For the moment all these unexpected errors are logged in the classic way and not in JSON. It's a deeper refactoring to do this and outside the scope of this PR whose main goal is to clean up several spaghetti functions that are no longer maintainable.

Okay, gotcha!


> > Hm, I don't like the idea that a downstream app would use `messages` instead of the (future) REST API... that's the problem we have with the current API design—the implementation details are exposed.
> > If we recommend people read directly from `messages`, then we won't ever be able to change it (much).
> 
> I'm not saying that you have to directly read the `messages` table. The ideal is of course to use the future REST endpoint `/journals/` or even better imo `/events`. The idea is not to expose the structure of the base but all the values necessary to be able to build your own base perfectly adapted to your needs. `sqlite` is a database for a client application and optimized for parsing and maintaining consensus not for a web application. Directly using the Counterparty API to build a website is bad practice and makes no sense.

Yeah I don't know that we should expose `/journal` or `/events` as a REST endpoint, but we can decide that later...

adamkrellenstein (approved):

Issue #1481, opened 2024-03-08
Make Health Check Endpoint Heavier


adamkrellenstein:
We want it to check that Counterparty is *fully* up and running. What @jdogresorg has been using for Pingdom is a `create_send` call. We can have the `/healthz` endpoint do a `create_send` and verify that it was successful before returning `200 Healthy`.

```
{
    "method" : "create_send",
    "params" : {
		"source": "1JDogZS6tQcSxwfxhv6XKKjcyicYA4Feev",
		"destination": "1FwkKA9cqpNRFTpVaokdRjT9Xamvebrwcu",
		"asset": "XCP",
		"quantity": 100000000,
		"allow_unconfirmed_inputs": true,
		"fee": 1000
    },
    "jsonrpc" : "2.0",
    "id" : 0
}
```

adamkrellenstein:
Getting this problem intermittently, also:

```
7|counterparty-server start  | WARNING:root:Error composing send transaction via API: Insufficient BTC at address 1JDogZS6tQcSxwfxhv6XKKjcyicYA4Feev. (Need approximately 0.00036274 BTC.) T
o spend unconfirmed coins, use the flag `--unconfirmed`. (Unconfirmed coins cannot be spent from multi‐sig addresses.)                                                                      
7|counterparty-server start  | WARNING:root:Traceback (most recent call last):                                                                                                              
7|counterparty-server start  |   File "/home/adam/counterparty-core/counterparty-lib/counterpartylib/lib/api.py", line 636, in create_method                                                
7|counterparty-server start  |     return compose_transaction(self.db, name=tx, params=transaction_args, **common_args)                                                                     
7|counterparty-server start  |   File "/home/adam/counterparty-core/counterparty-lib/counterpartylib/lib/api.py", line 480, in compose_transaction                                          
7|counterparty-server start  |     return transaction.construct(db, tx_info, encoding=encoding,                                                                                             
7|counterparty-server start  |   File "/home/adam/counterparty-core/counterparty-lib/counterpartylib/lib/transaction.py", line 488, in construct                                            
7|counterparty-server start  |     inputs, change_quantity, n_btc_in, n_final_fee = construct_coin_selection(                                                                               
7|counterparty-server start  |   File "/home/adam/counterparty-core/counterparty-lib/counterpartylib/lib/transaction.py", line 231, in construct_coin_selection                             
[...]

PR #1560, opened 2024-03-26
implements healthz probes / graceful shutdown 

mattmarcello:
This ended up being a bit of a trip down the rabbit hole.  A few observations:

1) The `AddrIndexRsThread` will just poll indefinitely for a message response.  There is no limit on retries.  This means that when composing the `create_send` tx while `addrindexrs` is down, that thread just keeps polling for a response message.

2) `counterparty-server` does not reconnect to `addrindexrs` after `addrinexrs` is stopped and started. 

So, when using the heavy probe while addrindexrs is down, you will get an "Unhealthy" response, but it will also cause the `AddrIndexRsThread` to loop indefinitely as it waits for a response message ( blocking future messages ).

I could add a retry limit on the `AddrIndexRsThread` but given that we are replacing this soon, I'm not sure it's worth the effort. 

What does everyone think?


mattmarcello:
refs  #1481

adamkrellenstein:
I've been going back and forth on this, but I think we should probably fix those AddrIndexRs issues for this release. They sound Bad.

Also, for the light/heavy flag, I was thinking more like a query param `?type=light`

counterparty-lib/counterpartylib/lib/backend/addrindexrs.py
```diff
@@ -275,118 +276,172 @@ def getrawtransaction_batch(txhash_list, verbose=False, skip_missing=False, _ret
...
+            logger.debug(f"{self.name} -- Closing socket")
+            self.socket.close()
+            logger.debug(f"{self.name} -- Close socket success")
+        except Exception as e:
+            logger.exception(f"{self.name} -- Close socket failure -- {e}")
+
+
+    def recover_connection(self):
+        if self.socket:
+            logger.debug(f"{self.name} -- Recoving connection")
```
adamkrellenstein, line 374: typo

adamkrellenstein (approved):

ouziel-slama (approved):

adamkrellenstein (approved):

PR #1562, opened 2024-03-28
Fix random consensus error around block 823000

ouziel-slama:
The problem was a hidden timeout error. A first call to `get_oldest_tx()` returned `{}` because of a timeout and the next call to `get_oldest_tx()` returned the result of the previous call which had finally arrived.
So I have:
- increased the timeout from 5 to 20 seconds.
- made the `get_oldest_tx()` function much stricter. We only return `{}` if the error returned by `addrindexrs` is 'no txs for address'
If `addrindexrs` returns nothing after 20s we throw an error.


adamkrellenstein (approved):
nice find!

PR #1567, opened 2024-03-30
v10.0.0-rc.1

ouziel-slama:
Warning: this is a release  candidate to allow people who maintain a node to test version v10.0.0.

See the release notes: https://github.com/CounterpartyXCP/counterparty-core/blob/develop/release-notes/release-notes-v10.0.0.md

See remaining issues for final version:
https://github.com/CounterpartyXCP/counterparty-core/milestone/12


Dockerfile
```diff
@@ -32,5 +32,5 @@ RUN pip3 install .
 WORKDIR /counterparty-cli
 RUN pip3 install .
 
-ENTRYPOINT [ "counterparty-server", "start" ]
+ENTRYPOINT [ "counterparty-server"]
 CMD [ "-h" ]
```
adamkrellenstein, line 36: Do we want the CMD not to be `start`?

README.md
```diff
@@ -13,18 +13,38 @@ The simplest way to get your Counterparty node up and running is to use Docker C
...
+You can use the environment variable `COUNTERPARTY_NETWORK` to run a `testnet` node:
+
+```
+COUNTERPARTY_NETWORK=test docker-compose up -d
 ```
 
-Then wait for your node to catch up with the network. Note: this process currently takes a long time, beause it does not make use of the `bootstrap` or `kickstart` functionality. (See below.)
+NOTES:
+- By default, this Docker Compose script makes use of the `bootstrap` functionality, because Docker makes it hard to use `kickstart`. (See below.)
+- When working with a low-memory system, you can tell `addrindexrs` to use JSON-RPC to communicate with `bitcoind` using the environment variable `ADDRINDEXRS_JSONRPC_IMPORT`: `ADDRINDEXRS_JSONRPC_IMPORT=true docker-compose up -d`
```
adamkrellenstein, line 47: I think we shouldn't give users the option of not using `JSONRPC_IMPORT` because it'll be incompatible with `kickstart`

compose.yml
```diff
@@ -0,0 +1,79 @@
...
+      - addrindexrs
+    volumes:
+      - data:/root/.bitcoin
+      - data:/data
+    ports:
+      - "4000:4000"                                        # mainnet
+      - "14000:14000"                                      # testnet
+    command:
+      - start
+      - "--${COUNTERPARTY_NETWORK:-main}net"               # export COUNTERPARTY_NETWORK=test for testnet
```
adamkrellenstein, line 62: Hm, I still think this is going to be confusing... `NETWORK` means to me "mainnet" or "testnet", not "main" or "test".

Any ideas how to work around that?

Dockerfile
adamkrellenstein: you should just be able to do `docker run` and have the service start

compose.yml
```diff
@@ -0,0 +1,79 @@
+version: '3'
```
adamkrellenstein, line 1: let's rename this file `docker-compose.yml`? I think that's just as standard, and it's more perspicuous

Dockerfile
adamkrellenstein: wait, is `start` the default now?

compose.yml
```diff
@@ -0,0 +1,79 @@
...
+      - "XDG_LOG_HOME=/data/"
+
+
+volumes: 
+  data:
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      device: ${COUNTERPARTY_DOCKER_DATA:-~/counterparty-docker-data}/
```
adamkrellenstein, line 79: Hm, I think we should put this somewhere else by default... maybe `$XDG_CACHE_HOME/counterparty-docker-data` I don't want to pollute people's home dirs.

adamkrellenstein (changes requested):

Dockerfile
ouziel-slama: There is no default command for now. The `start` was removed to allow people to use the docker image for something other than `start` and without using the `--entrypoint` flag. It is not standard to include a command in the entrypoint.. this is what the CMD directive is intended for.

README.md
ouziel-slama: sorry I don't understand.. you want to force people to use ADDRINDEXRS_JSONRPC_IMPORT=true ? It's a lot slower and I don't understand the connection with `kickstart`.

adamkrellenstein: so I assumed that if addrindexrs is reading directly from the leveldb files (when `JSONRPC_IMPORT=false`) that'll block kickstart from working because only one process can access leveldb at a time...

Dockerfile
adamkrellenstein: hm, I don't follow... so if we set `CMD` to `start`, then wouldn't users be able to do something other than `start` without changing `ENTRYPOINT`? (isn't that the point of having `CMD`? 😅)

compose.yml
ouziel-slama: hm.. $XDG_CACHE_HOME is in people's home dir.. by default it's `~/.cache`.. and the counterparty data is not really cache..
what do you think of `${COUNTERPARTY_DOCKER_DATA:-~/.local/share/counterparty-docker-data}`

adamkrellenstein: hm yeah we could use `$XDG_DATA_HOME`, which is ~/.local/share` by default

(I was thinking "cache" only because the data *can* be reproduced, but `DATA` is probably better)

Dockerfile
ouziel-slama: yes we say the same thing :-)) I misunderstood you sorry 😅

README.md
ouziel-slama: `addrindexrs` is not reading from the leveldb files but only from the blk files and there is no lock problem on these files. Additionally addrindexrs only reads blk files for the first catch up, then it uses RPC.. and before kcikstart can begin addrindexrs must be up to date and therefore no longer read blk files anyway..

compose.yml
ouziel-slama: done

compose.yml
ouziel-slama: > yep, we could use $XDG_DATA_HOME, which is ~/.local/share` by default!

`XDG_DATA_HOME` is not set most of the time! and docker-compose will not replace `XDG_DATA_HOME` by the default folder but by an empty string..
we could have something like this:

`${COUNTERPARTY_DOCKER_DATA:-${XDG_DATA_HOME:-~/.local/share}}/counterparty-docker-data`

(in this cas data will be always in `counterparty-docker-data` subfolder)

compose.yml
ouziel-slama: no sorry, I didn't find anything!

README.md
adamkrellenstein: okay gotcha!

adamkrellenstein (approved):

compose.yml
adamkrellenstein: that works!

PR #1583, opened 2024-04-01
Fix and tweak 'check-db' command

ouziel-slama:


adamkrellenstein (approved):
does this need something in the release notes?

ouziel-slama:
yep! I updated what was already there!

PR #1586, opened 2024-04-02
Split Out Wallet

ouziel-slama:


ouziel-slama:
Split `counterparty-cli` into two subfolders: `counterparty-core` and `counterparty-wallet` which corresponds to the new names of the two CLIs:
`counterparty-client` becomes `counterparty-wallet` and `counterparty-server` becomes `counterparty-core`

PR #1607, opened 2024-04-05
[WIP] API Redesign

ouziel-slama:
- [x] Implement API v2
    - [x] ensure that all data is covered
    - [x] ensure that all API v1 endpoints have an equivalent
    - [x] Implement unpack for all message types
- [x] Deprecate API v1
    - [x] Use X-API-Warn
    - [x] Throw warnings when they're used
    - [x] Put behind a CLI flag
    - [x] Put behind /old/api/
- [ ] Test API v2
- [ ] Document API v2

# Quick description of API v2

## New flags

- `--api-host` (default: "localhost")
- `--api-port` (default: 4000)
- `--api-user` (default: "api")
- `--api-password` (default: "api")
- `--api-no-allow-cors` (default: False)
- `--api-not-ready-http-code` (default: 503)
- `--enable-v1-api` (default: False)

New default port for the API v1: 4100

## New Headers

All responses contain the following three headers:

`X-COUNTERPARTY-HEIGHT`: Last parsed block
`X-BACKEND-HEIGHT`: `backend.getblockcount()` + `backend.getindexblocksbehind()`
`X-COUNTERPARTY-READY`: `X-COUNTERPARTY-HEIGHT` >= `X-BACKEND-HEIGHT`

### Get Ledger State

- #### /blocks ####
    - `/blocks`
    - `/blocks/<int:block_index>`
    - `/blocks/<int:block_index>/transactions`
    - `/blocks/<int:block_index>/events`
    - `/blocks/<int:block_index>/events/counts`
    - `/blocks/<int:block_index>/events/<event>`
    - `/blocks/<int:block_index>/credits`
    - `/blocks/<int:block_index>/debits`
    - `/blocks/<int:block_index>/expirations`
    - `/blocks/<int:block_index>/cancels`
    - `/blocks/<int:block_index>/destructions`
    - `/blocks/<int:block_index>/issuances`
    - `/blocks/<int:block_index>/sends`
    - `/blocks/<int:block_index>/dispenses`
    - `/blocks/<int:block_index>/sweeps`
- #### /transactions ####
    - `/transactions/<tx_hash>`
- #### /addresses ####
    - `/addresses/<address>/balances`
    - `/addresses/<address>/balances/<asset>`
    - `/addresses/<address>/credits`
    - `/addresses/<address>/debits`
    - `/addresses/<address>/bets`
    - `/addresses/<address>/broadcasts`
    - `/addresses/<address>/burns`
    - `/addresses/<address>/sends`
    - `/addresses/<address>/receives`
    - `/addresses/<address>/sends/<asset>`
    - `/addresses/<address>/receives/<asset>`
    - `/addresses/<address>/dispensers`
    - `/addresses/<address>/dispensers/<asset>`
    - `/addresses/<address>/sweeps`
- #### /assets ####
    - `/assets`
    - `/assets/<asset>`
    - `/assets/<asset>/balances`
    - `/assets/<asset>/balances/<address>`
    - `/assets/<asset>/orders`
    - `/assets/<asset>/credits`
    - `/assets/<asset>/debits`
    - `/assets/<asset>/dividends`
    - `/assets/<asset>/issuances`
    - `/assets/<asset>/sends`
    - `/assets/<asset>/dispensers`
    - `/assets/<asset>/dispensers/<address>`
    - `/assets/<asset>/holders`
- #### /orders ####
    - `/orders/<tx_hash>`
    - `/orders/<tx_hash>/matches`
    - `/orders/<tx_hash>/btcpays`
- #### /bets ####
    - `/bets/<tx_hash>`
    - `/bets/<tx_hash>/matches`
    - `/bets/<tx_hash>/resolutions`
- #### /burns ####
    - `/burns`
- #### /dispensers ####
    - `/dispensers/<tx_hash>`:
    - `/dispensers/<tx_hash>/dispenses`
- #### /events ####
    - `/events`
    - `/events/<int:event_index>`
    - `/events/counts`
    - `/events/<event>`
- #### /mempool ####
    - `/mempool/events`
    - `/mempool/events/<event>`

### Compose and Parse Transactions

- `/transactions/compose/<transaction_name>`
- `/transactions/info`
- `/transactions/unpack`

### Get Server State

- `/`
- `/healthz`

### Backend Proxy

- `/backend/addresses/<address>/transactions`
- `/backend/addresses/<address>/transactions/oldest`
- `/backend/addresses/<address>/utxos`
- `/backend/addresses/<address>/pubkey`
- `/backend/transactions/<tx_hash>`
- `/backend/transactions`
- `/backend/estimatesmartfee`

### Get Ledger State

| API v1 | API v2 |
| ------ | ------ |
| `get_assets` | `/assets` <br /> `/assets/<asset>` <br /> `/events/ASSET_CREATION` <br /> `/blocks/<int:block_index>/events/ASSET_CREATION` |
| `get_balances` | `/assets/<asset>/balances` <br /> `/addresses/<address>/balances` <br /> `/addresses/<address>/balances/<asset>` <br/> `/assets/<asset>/balances/<address>` |
| `get_credits` | `/blocks/<int:block_index>/credits` <br /> `/addresses/<address>/credits` <br /> `/assets/<asset>/credits` <br /> `/events/CREDIT` <br /> `/blocks/<int:block_index>/events/CREDIT` |
| `get_debits` | `/blocks/<int:block_index>/debits` <br /> `/addresses/<address>/debits` <br /> `/assets/<asset>/debits` <br />  `/events/DEBIT` <br /> `/blocks/<int:block_index>/events/DEBIT` |
| `get_bets` | `/addresses/<feed_address>/bets` <br /> `/bets/<tx_hash>` <br /> `/events/OPEN_BET` <br /> `/blocks/<int:block_index>/events/OPEN_BET` |
| `get_bet_matches` | `/bets/<tx_hash>/matches` <br /> `/events/BET_MATCH` <br /> `/blocks/<int:block_index>/events/BET_MATCH` |
| `get_broadcasts` | `/addresses/<feed_address>/broadcasts` <br /> `/events/BROADCAST` <br /> `/blocks/<int:block_index>/events/BROADCAST` |
| `get_btcpays` | `/orders/<tx_hash>/btcpays` <br /> `/events/BTC_PAY` <br /> `/blocks/<int:block_index>/events/BTC_PAY` |
| `get_burns` |  `/addresses/<address>/burns` <br /> `/burns` <br /> `/events/BURN` <br /> `/blocks/<int:block_index>/events/BURN` |
| `get_cancels` | `/blocks/<int:block_index>/cancels` <br /> `/events/CANCEL_BET` <br /> `/blocks/<int:block_index>/events/CANCEL_BET` <br /> `/events/CANCEL_ORDER` <br /> `/blocks/<int:block_index>/events/CANCEL_ORDER` <br /> `/events/CANCEL_RPS` <br /> `/blocks/<int:block_index>/events/CANCEL_RPS` |
| `get_destructions` | `/blocks/<int:block_index>/destructions` <br /> `/events/ASSET_DESTRUCTION` <br /> `/blocks/<int:block_index>/events/ASSET_DESTRUCTION` |
| `get_dividends` | `/assets/<asset>/dividends` <br /> `/events/ASSET_DIVIDEND` <br /> `/blocks/<int:block_index>/events/ASSET_DIVIDEND` |
| `get_issuances` | `/blocks/<int:block_index>/issuances` <br /> `/assets/<asset>/issuances` <br /> `/events/ASSET_ISSUANCE` <br /> `/blocks/<int:block_index>/events/ASSET_ISSUANCE` |
| `get_orders` | `/assets/<asset>/orders` <br /> `/orders/<tx_hash>` <br /> `/events/OPEN_ORDER` <br /> `/blocks/<int:block_index>/events/OPEN_ORDER` |
| `get_order_matches` | `/orders/<tx_hash>/matches` <br /> `/events/ORDER_MATCH` <br /> `/blocks/<int:block_index>/events/ORDER_MATCH` |
| `get_sends` | `/blocks/<int:block_index>/sends` <br /> `/assets/<asset>/sends` <br /> `/addresses/<address>/sends` <br /> `/addresses/<address>/receives` <br /> `/addresses/<address>/sends/<asset>` <br /> `/addresses/<address>/receives/<asset>` <br /> `/events/SEND` <br /> `/blocks/<int:block_index>/events/SEND` <br /> `/events/MPMA_SEND` <br /> `/blocks/<int:block_index>/events/MPMA_SEND` <br /> `/events/ENHANCED_SEND` <br /> `/blocks/<int:block_index>/events/ENHANCED_SEND` |
| `get_bet_expirations` | `/blocks/<int:block_index>/expirations` <br /> `/events/BET_EXPIRATION` <br /> `/blocks/<int:block_index>/events/BET_EXPIRATION` |
| `get_order_expirations` | `/blocks/<int:block_index>/expirations` <br /> `/events/ORDER_EXPIRATION` <br /> `/blocks/<int:block_index>/events/ORDER_EXPIRATION` |
| `get_bet_match_expirations` | `/blocks/<int:block_index>/expirations` <br /> `/events/BET_MATCH_EXPIRATION` <br /> `/blocks/<int:block_index>/events/BET_MATCH_EXPIRATION` |
| `get_order_match_expirations` | `/blocks/<int:block_index>/expirations` <br /> `/events/ORDER_MATCH_EXPIRATION` <br /> `/blocks/<int:block_index>/events/ORDER_MATCH_EXPIRATION` |
| `get_bet_match_resolutions` | `/bets/<tx_hash>/resolutions` <br /> `/events/BET_MATCH_RESOLUTON` <br /> `/blocks/<int:block_index>/events/BET_MATCH_RESOLUTON` |
| `get_rps` | `/events/OPEN_RPS` <br /> `/blocks/<int:block_index>/events/OPEN_RPS` |
| `get_rpsresolves` | `/events/RPS_RESOLVE` <br /> `/blocks/<int:block_index>/events/RPS_RESOLVE` |
| `get_rps_matches` | `/events/RPS_MATCH` <br /> `/blocks/<int:block_index>/events/RPS_MATCH` |
| `get_rps_expirations` | `/blocks/<int:block_index>/expirations` <br /> `/events/RPS_EXPIRATION` <br /> `/blocks/<int:block_index>/events/RPS_EXPIRATION` |
| `get_rps_match_expirations` | `/blocks/<int:block_index>/expirations` <br /> `/events/RPS_MATCH_EXPIRATION` <br /> `/blocks/<int:block_index>/events/RPS_MATCH_EXPIRATION` |
| `get_sweeps` | `/addresses/<address>/sweeps`<br /> `/blocks/<int:block_index>/sweeps` <br /> `/events/SWEEP` <br /> `/blocks/<int:block_index>/events/SWEEP` |
| `get_dispensers` | `/assets/<asset>/dispensers` <br /> `/asset/<asset>/dispensers/<address>` <br /> `/addresses/<address>/dispensers` <br /> `/addresses/<address>/dispensers/<asset>` <br /> `/dispensers/<tx_hash>` <br /> `/events/OPEN_DISPENSER` <br /> `/blocks/<int:block_index>/events/OPEN_DISPENSER` |
| `get_dispenses` | `/blocks/<int:block_index>/dispenses` <br /> `/dispensers/<tx_hash>/dispenses` <br /> `/events/DISPENSE` <br /> `/blocks/<int:block_index>/events/DISPENSE` |
| `get_transactions` | `/blocks/<int:block_index>/transactions` <br /> `/transactions/<tx_hash>` <br /> `/events/NEW_TRANSACTION` <br /> `/blocks/<int:block_index>/events/NEW_TRANSACTION` |
| `get_messages` | `/blocks/<int:block_index>/events` <br /> `/blocks/<int:block_index>/events/<event>` <br /> `/events/<event>` |
| `get_messages_by_index` | `/events/<int:event_index>` |
| `get_supply` | `/assets/<asset>` |
| `get_xcp_supply` | `/assets/XCP` |
| `get_asset_info` | `/assets/<asset>` |
| `get_block_info` | `/blocks/<int:block_index>` <br /> `/blocks/<int:block_index>/events`  |
| `get_blocks`| `/blocks` |
| `get_asset_names` | `/assets` |
| `get_asset_longnames` | `/assets` |
| `get_holder_count` | `/assets/<asset>` |
| `get_holders` | `/assets/<asset>/holders` |
| `get_dispenser_info` | `/dispensers/<tx_hash>` |
| `get_mempool` | `/mempool/events` <br /> `/mempool/events/<event>` |
| `get_element_counts` | `/blocks/<int:block_index>/events/counts` <br /> `/events/counts` |
| `sql` | NA |

### Compose and Parse Transactions

| API v1 | API v2 |
| ------ | ------ |
| `create_bet` | `/transactions/compose/bet` |
| `create_broadcast` | `/transactions/compose/broadcast` |
| `create_btcpay` | `/transactions/compose/btcpay` |
| `create_burn` | `/transactions/compose/burn` |
| `create_cancel` | `/transactions/compose/cancel` |
| `create_destroy` | `/transactions/compose/destroy` |
| `create_dividend` | `/transactions/compose/dividend` |
| `create_issuance` | `/transactions/compose/issuance` |
| `create_order` | `/transactions/compose/order` |
| `create_send` | `/transactions/compose/send` |
| `create_rps` | `/transactions/compose/rps` |
| `create_rpsresolve` | `/transactions/compose/rpsresolve` |
| `create_sweep` | `/transactions/compose/sweep` |
| `create_dispenser` | `/transactions/compose/dispenser` |
| `get_tx_info` | `/transactions/info` |
| `unpack` | `/transactions/unpack` |

### Get Server State

| API v1 | API v2 |
| ------ | ------ |
| `get_running_info` | `/` |
| `/healthz` | `/healthz` |

### Backend Proxy

| API v1 | API v2 |
| ------ | ------ |
| `search_raw_transactions` | `/backend/addresses/<address>/transactions` |
| `get_oldest_tx` | `/backend/addresses/<address>/transactions/oldest` |
| `get_unspent_txouts` | `/backend/addresses/<address>/utxos` |
| `getrawtransaction` | `/backend/transactions/<tx_hash>` |
| `getrawtransaction_batch` | `/backend/transactions` |
| `search_pubkey` | `/backend/addresses/<address>/pubkey` |
| `fee_per_kb` | `/backend/estimatesmartfee` |


counterparty-lib/counterpartylib/lib/restapi.py
```diff
@@ -0,0 +1,121 @@
+import argparse
```
adamkrellenstein, line 1: I think we should call this file just `api.py` and move the existing `api.py` to `v1/api.py`?

counterparty-core/counterpartycore/server.py
```diff
@@ -227,6 +227,10 @@
             "help": "how long to keep a lock on a UTXO being tracked",
         },
     ],
+    [
+        ("--legacy-api",),
```
adamkrellenstein, line 231: hm maybe we should do `--enable-v1-api`?

counterparty-lib/counterpartylib/lib/restapi.py
ouziel-slama: yep! sounds good!

counterparty-core/counterpartycore/server.py
ouziel-slama: ok but it's one or the other of the two versions. Not both at the same time, otherwise you will have to provide another port..

adamkrellenstein: ah we can't have the other one available at `/v1/whatever`?

ouziel-slama: we could ofcbut the code is really much cleaner like that! the two APIs are well isolated and we just have to delete the `v1/` folder when we are ready...

adamkrellenstein: ah yeah I think we're going to have to support both simultaneously on the same port (with the `/v1/` prefix)

ouziel-slama: The current API code is spaghetti code. I don't know how to do this properly.

adamkrellenstein: ach yeah. maybe we should just use a second port for the v1 API

ouziel-slama: yeah sounds good! anyway we must add flags other than `--rpc-*` like `--api-port` etc..

ouziel-slama:
Here is an equivalence table between API v1 and v2. I will add it to the documentation when it is complete.

| API v1 | API v2 |
| ------ | ------ |
| `get_assets` | `/assets/<asset>` |
| `get_balances` | `/assets/<asset>/balances` <br /> `/addresses/<address>/balances` |
| `get_credits` |   |
| `get_debits` |   |
| `get_bets` |   |
| `get_bet_matches` |   |
| `get_broadcasts` |   |
| `get_btcpays` |   |
| `get_burns` |   |
| `get_cancels` |   |
| `get_destructions` |   |
| `get_dividends` |   |
| `get_issuances` |   |
| `get_orders` | `/assets/<asset>/orders` <br /> `/orders/<tx_hash>` |
| `get_order_matches` | `/orders/<tx_hash>/matches` |
| `get_sends` |   |
| `get_bet_expirations` |   |
| `get_order_expirations` |   |
| `get_bet_match_expirations` |   |
| `get_order_match_expirations` |   |
| `get_bet_match_resolutions` |   |
| `get_rps` |   |
| `get_rpsresolves` |   |
| `get_rps_matches` |   |
| `get_rps_expirations` |   |
| `get_rps_match_expirations` |   |
| `get_mempool` |   |
| `get_sweeps` |   |
| `get_dispensers` |   |
| `get_dispenses` |   |
| `get_transactions` | `/blocks/<int:block_index>/transactions` |
| `create_bet` |   |
| `create_broadcast` |   |
| `create_btcpay` |   |
| `create_burn` |   |
| `create_cancel` |   |
| `create_destroy` |   |
| `create_dividend` |   |
| `create_issuance` |   |
| `create_order` |   |
| `create_send` |   |
| `create_rps` |   |
| `create_rpsresolve` |   |
| `create_sweep` |   |
| `create_dispenser` |   |
| `sql` |   |
| `get_messages` |   |
| `get_messages_by_index` |   |
| `get_supply` |   |
| `get_xcp_supply` |   |
| `get_asset_info` |   |
| `get_block_info` | `/blocks/<int:block_index>` <br /> `/blocks/<int:block_index>/events`  |
| `fee_per_kb` |   |
| `get_blocks`| `/blocks` |
| `get_running_info` |   |
| `get_element_counts` |   |
| `get_asset_names` |   |
| `get_asset_longnames` |   |
| `get_holder_count` |   |
| `get_holders` |   |
| `search_raw_transactions` |   |
| `get_oldest_tx` |   |
| `get_unspent_txouts` |   |
| `getrawtransaction` |   |
| `getrawtransaction_batch` |   |
| `get_tx_info` |   |
| `unpack` |   |
| `search_pubkey` |   |
| `get_dispenser_info` |   |
| `/healthz` |   |

ouziel-slama:
# New Routes

## Get Ledger State

- ### /blocks ###
    - `/blocks`
    - `/blocks/<int:block_index>`
    - `/blocks/<int:block_index>/transactions`
    - `/blocks/<int:block_index>/events`
    - `/blocks/<int:block_index>/events/counts`
    - `/blocks/<int:block_index>/events/<event>`
    - `/blocks/mempool/events`
    - `/blocks/mempool/events/<event>`
    - `/blocks/<int:block_index>/credits`
    - `/blocks/<int:block_index>/debits`
    - `/blocks/<int:block_index>/expirations`
    - `/blocks/<int:block_index>/cancels`
    - `/blocks/<int:block_index>/destructions`
    - `/blocks/<int:block_index>/issuances`
    - `/blocks/<int:block_index>/sends`
    - `/blocks/<int:block_index>/dispenses`
    - `/blocks/<int:block_index>/sweeps`
- ### /transactions ###
    - `/transactions/<tx_hash>`
- ### /addresses ###
    - `/addresses/<address>/balances`
    - `/addresses/<address>/balances/<asset>`
    - `/addresses/<address>/credits`
    - `/addresses/<address>/debits`
    - `/addresses/<address>/bets`
    - `/addresses/<address>/broadcasts`
    - `/addresses/<address>/burns`
    - `/addresses/<address>/sends`
    - `/addresses/<address>/receives`
    - `/addresses/<address>/sends/<asset>`
    - `/addresses/<address>/receives/<asset>`
    - `/addresses/<address>/dispensers`
    - `/addresses/<address>/dispensers/<asset>`
    - `/addresses/<address>/sweeps`
- ### /assets ###
    - `/assets`
    - `/assets/<asset>`
    - `/assets/<asset>/balances`
    - `/assets/<asset>/balances/<address>`
    - `/assets/<asset>/orders`
    - `/assets/<asset>/credits`
    - `/assets/<asset>/debits`
    - `/assets/<asset>/dividends`
    - `/assets/<asset>/issuances`
    - `/assets/<asset>/sends`
    - `/assets/<asset>/dispensers`
    - `/assets/<asset>/dispensers/<address>`
    - `/assets/<asset>/holders`
- ### /orders ###
    - `/orders/<tx_hash>`
    - `/orders/<tx_hash>/matches`
    - `/orders/<tx_hash>/btcpays`
- ### /bets ###
    - `/bets/<tx_hash>`
    - `/bets/<tx_hash>/matches`
    - `/bets/<tx_hash>/resolutions`
- ### /burns ###
    - `/burns`
- ### /dispensers ###
    - `/dispensers/<tx_hash>`:
    - `/dispensers/<tx_hash>/dispenses`
- ### /events ###
    - `/events/<int:event_index>`
    - `/events/counts`
    - `/events/<event>`

## Compose and Parse Transactions
## Get Server State
## Backend Proxy

ouziel-slama:
OP updated.

counterparty-core/counterpartycore/server.py
```diff
@@ -132,6 +132,39 @@
...
+        {
+            "default": "rpc",
+            "help": f"required password (for --api-user) to use the {config.APP_NAME}  API (via HTTP basic auth)",
+        },
+    ],
+    [
+        ("--api-no-allow-cors",),
+        {"action": "store_true", "default": False, "help": "allow ajax cross domain request"},
+    ],
+    [
```
adamkrellenstein, line 164: interesting! why should this be configurable?

adamkrellenstein: and I think maybe it should be 503...

counterparty-lib/counterpartylib/server.py
```diff
@@ -450,6 +460,57 @@ def initialise_config(
...
+    if api_host:
+        config.API_HOST = api_host
+    else:
+        config.API_HOST = "localhost"
+
+    if api_port:
+        config.API_PORT = rpc_port
+    else:
+        if config.TESTNET:
+            if config.TESTCOIN:
```
adamkrellenstein, line 477: Do you know what this is?

counterparty-lib/counterpartylib/lib/api/routes.py
```diff
@@ -0,0 +1,255 @@
...
+        "function": ledger.get_events_counts,
+    },
+    "/events/<event>": {
+        "function": ledger.get_events,
+        "args": [("last", None), ("limit", 100)],
+    },
+    ### /healthz ###
+    "/healthz": {
+        "function": util.handle_healthz_route,
+        "args": [("check", "heavy")],
```
adamkrellenstein, line 222: should that be `check_type`?

adamkrellenstein (changes requested):

mattmarcello:
I wonder if mempool should be behind it's own resource, instead of blocks since it only indexes `events` and `events/<event>` ( i.e. is not totally symmetrical with other `block` resources.

```
/mempool/events
/mempool/events/<event>

```

counterparty-core/counterpartycore/server.py
ouziel-slama: > interesting! why should this be configurable?

Because I think 202 is good, but I was sure and certain that someone would think otherwise LOL. I will set 503 as default.

counterparty-lib/counterpartylib/server.py
ouziel-slama: no idea! I copy/past from RPC_*

mattmarcello:
Just a few minor thoughts for the `/transactions/compose/` API.  You are going to need to post data anyway here, so I don't think it's necessary to have `/compose`.  Could just be a POST to `/transactions`.  To that end, would it be more clear to call this endpoint something like `/rawtransactions` to make it explicit that this is not tantamount to a create and broadcast?  Also, maybe it would be nicer from the callers standpoint to pass the transaction type as data in the POST request?  Not 100% sure.  
```
curl -d '{"type":"bet", "args": ...betargs }' -H "Content-Type: application/json" -X POST http://localhost:4000/rawtransactions
```

<h3 dir="auto">Compose and Parse Transactions</h3>

API v1 | API v2 | --
-- | -- | -- 
create_bet | /transactions/compose/bet | curl -d '{"type":"bet", "args": ...betargs }' -H "Content-Type: application/json" -X POST http://localhost:4000/rawtransactions 
create_broadcast | /transactions/compose/broadcast | curl -d '{"type":"broadcast", "args": ...broadcastargs  }' -H "Content-Type: application/json" -X POST http://localhost:4000/rawtransactions
...

counterparty-lib/counterpartylib/lib/api/routes.py
ouziel-slama: it could! Isn't a "heavy check" the same as a "heavy type check" ?

ouziel-slama:
> @ouziel-slama, this looks great. Question for you: are these resources paginated?

A basic pagination only for some queries (with the parameters `last` and `limit`)

counterparty-lib/counterpartylib/server.py
adamkrellenstein: lol, okay I'll create an issue for removing it later :)

ouziel-slama:
> I wonder if mempool should be behind it's own resource, instead of blocks since it only indexes `events` and `events/<event>` ( i.e. is not totally symmetrical with other `block` resources.
> 
> ```
> /mempool/events
> /mempool/events/<event>
> ```

I hesitated.. we can say that the mempool contains the next block, it's a mental view, but in the code it is managed like that with a block_index=999999999.. as you wish I can change easily ..

counterparty-lib/counterpartylib/lib/api/routes.py
adamkrellenstein: ah no, it's the type of the check—heavy vs. light

a heavy `healthz` check is slower but tests more

ouziel-slama:
> Just a few minor thoughts for the `/transactions/compose/` API. You are going to need to post data anyway here, so I don't think it's necessary to have `/compose`. Could just be a POST to `/transactions`. To that end, would it be more clear to call this endpoint something like `/rawtransactions` to make it explicit that this is not tantamount to a create and broadcast? Also, maybe it would be nicer from the callers standpoint to pass the transaction type as data in the POST request? Not 100% sure.


hum..  I am sure that all our requests must be GET and there must be no POST. We open the database in read only mode and no queries modify the database. Also in most cases the url is formed with a script/CLI or something else, I don't think this is a problem.

adamkrellenstein:
> > I wonder if mempool should be behind it's own resource, instead of blocks since it only indexes `events` and `events/<event>` ( i.e. is not totally symmetrical with other `block` resources.
> > ```
> > /mempool/events
> > /mempool/events/<event>
> > ```
> 
> I hesitated.. we can say that the mempool contains the next block, it's a mental view, but in the code it is managed like that with a block_index=999999999.. as you wish I can change easily ..

Yeah I think we should have a dedicated route and later fix  that magic number block index, which bothers me :/ (it's my fault originally!)

mattmarcello:
> > I wonder if mempool should be behind it's own resource, instead of blocks since it only indexes `events` and `events/<event>` ( i.e. is not totally symmetrical with other `block` resources.
> > ```
> > /mempool/events
> > /mempool/events/<event>
> > ```
> 
> I hesitated.. we can say that the mempool contains the next block, it's a mental view, but in the code it is managed like that with a block_index=999999999.. as you wish I can change easily ..

Interesting.  I presume these requests would be errors? 

`/blocks/999999999/sends` 
`/blocks/mempool/sends`

ouziel-slama:
> Interesting. I presume these requests would be errors?
> 
> `/blocks/999999999/sends` `/blocks/mempool/sends`

yep

mattmarcello:
> > Just a few minor thoughts for the `/transactions/compose/` API. You are going to need to post data anyway here, so I don't think it's necessary to have `/compose`. Could just be a POST to `/transactions`. To that end, would it be more clear to call this endpoint something like `/rawtransactions` to make it explicit that this is not tantamount to a create and broadcast? Also, maybe it would be nicer from the callers standpoint to pass the transaction type as data in the POST request? Not 100% sure.
> 
> hum.. I am sure that all our requests must be GET and there must be no POST. We open the database in read only mode and no queries modify the database. Also in most cases the url is formed with a script/CLI or something else, I don't think this is a problem

I think this is slightly murky territory.  You are not looking up an existing resource, but creating an ephemeral one.  In any case, I think passing data as JSON will be more familiar to web developers building on top of the platform relative to serializing in the URL.  We could pass JSON in a GET, but that too would be non-standard.

ouziel-slama:
> > > Just a few minor thoughts for the `/transactions/compose/` API. You are going to need to post data anyway here, so I don't think it's necessary to have `/compose`. Could just be a POST to `/transactions`. To that end, would it be more clear to call this endpoint something like `/rawtransactions` to make it explicit that this is not tantamount to a create and broadcast? Also, maybe it would be nicer from the callers standpoint to pass the transaction type as data in the POST request? Not 100% sure.
> > 
> > 
> > hum.. I am sure that all our requests must be GET and there must be no POST. We open the database in read only mode and no queries modify the database. Also in most cases the url is formed with a script/CLI or something else, I don't think this is a problem
> 
> I think this is slightly murky territory. You are not looking up an existing resource, but creating an ephemeral one. In any case, I think passing data as JSON will be more familiar to web developers building on top of the platform relative to serializing in the URL. We could pass JSON in a GET, but that too would be non-standard.
> 
> Just my 2 cents 🤑

nah I don't think so.. there is no question of resource here.. a raw transaction is not one.. etc.. we make a REST API so let's respect the basic principles.. here we only read and use data (from the bitcoind and counterparty db) to construct a response... no matter the size of the response or the request, this is not what determines whether we must do a GET or a POST (this is the case for example of charts API which uses huge urls). It's easy for a developer to convert a json, a dict or anything into an url.

mattmarcello:
The basic principles of REST do not dictate that a resource is created after a POST.  

https://stackoverflow.com/questions/19843480/s3-rest-api-and-post-method/19844272#19844272

And while it is possible for developers to serialize data in URLs, it is non standard from the perspective of most web developers and requires an explicit encoding step.  Obviously it can be done it just creates a bit of friction.  

In any case, it's your call! :)

counterparty-lib/counterpartylib/lib/api/routes.py
ouziel-slama: done

counterparty-lib/counterpartylib/server.py
ouziel-slama: ok :)

counterparty-core/counterpartycore/server.py
ouziel-slama: done

adamkrellenstein:
Hm, this is a tough one. I think we're not retrieving any (existing) resource, so `GET` feels wrong to me. And there are already times when using `POST` for idempotent operations is totally fine. Also think that passing the params in the URL is indeed awkward.

I think we should use `POST /transactions/compose`. `POST /transactions/` makes it look like you're writing to the DB, as would `POST /rawtransactions/`. These would obviously be more ReSTful, but I think sticking with good HTTP is more important than following ReST principles where they don't apply.

ouziel-slama:
> Hm, this is a tough one. I think we're not retrieving any (existing) resource, so `GET` feels wrong to me. And there are already times when using `POST` for idempotent operations is totally fine. Also think that passing the params in the URL is indeed awkward.
> 
> I think we should use `POST /transactions/compose`. `POST /transactions/` makes it look like you're writing to the DB, as would `POST /rawtransactions/`. These would obviously be more ReSTful, but I think sticking with good HTTP is more important than following ReST principles where they don't apply.

Yes we retrieve utxo from the bitcoind database and we check data from the counterparty database. Using a GET is not weird and using a POST makes no sense other than preventing developers from doing what is necessary, which is to say building a URL. Using a POST to make a request that does not do any writing to the database is a design error.

adamkrellenstein:
I'm not aware of any specification and convention that requires a `POST` to trigger any change to server state. I'm going by <https://www.rfc-editor.org/rfc/rfc7231#section-4.3.3>, which is pretty clear that all that is necessary is that the server "target resource process the representation enclosed in the request according to the resource's own specific semantics" and that no resources need to be created as a result of processing the request.

I also think it's actually nice to use `GET` for retrieving network state and `POST` for generating ad hoc data, esp. when the latter requires a proper payload.

ouziel-slama:
I'm going by https://www.rfc-editor.org/rfc/rfc7231#section-4.3.1
The compose method returns a representation of inputs and outputs in transaction form. I think using a POST here is ugly, non-standard and useless.

ouziel-slama:
> esp. when the latter requires a proper payload.

This is not a good enough reason to use a POST. It is nowhere written that a url does not contain a payload. We are not talking about a domain name but a URL systematically generated by code.

adamkrellenstein:
> > esp. when the latter requires a proper payload.
> 
> This is not a good enough reason to use a POST. It is nowhere written that a url does not contain a payload. We are not talking about a domain name but a URL systematically generated by code.

I agree with all of that, and that's why I wrote "esp."! :)


> I'm going by https://www.rfc-editor.org/rfc/rfc7231#section-4.3.1 

```
The GET method requests transfer of a current selected representation for the target resource.  
```

So this does not describe what we are doing—the newly constructed transaction is not a set of already extant inputs and outputs. The request to compose a tx is really a request to "process", not to "transfer". In any case, I *totally* hear what you're saying, but I think we should proceed with `POST /.../compose`. Yes, it'd be simpler and cleaner in some sense to just use `GET`s, but I think it's not quite as _correct_, and it's no uglier than HTTP in general. ;)

adamkrellenstein:
Let's include this only into v10.0.4.

ouziel-slama:
Note about API design

The idea behind the new API is “one route, one function”. Example:

in `api/routes.py`:

```
...
ROUTES = util.prepare_routes(
    {
        ### /blocks ###
        "/blocks": ledger.get_blocks,
        "/blocks/<int:block_index>": ledger.get_block,
        "/blocks/<int:block_index>/transactions": ledger.get_transactions_by_block,
        "/blocks/<int:block_index>/events": ledger.get_events_by_block,
        ...
    }
)
...
```

in `ledger.py`:
```
def get_blocks(db, last: int = None, limit: int = 10):
    """
    Returns the list of the last ten blocks
    :param int last: The index of the most recent block to return
    :param int limit: The number of blocks to return
    """
    ...

```

The new API uses type annotations and docstrings to generate route arguments, default values, and documentation.

`curl http://api:api@localhost:4000/` returns:

```
{
  "backend_height": 839940,
  "counterparty_height": 839940,
  "network": "mainnet",
  "server_ready": true,
  "version": "10.1.0"
  "routes": [
    ...
    {
      "path": "/blocks"
      "description": "Returns the list of the last ten blocks",
      "args": [
        {
          "default": null,
          "description": "The index of the most recent block to return",
          "name": "last",
          "required": false,
          "type": "int"
        },
        {
          "default": 10,
          "description": "The number of blocks to return",
          "name": "limit",
          "required": false,
          "type": "int"
        }
      ],
      
    },
    ...
  ]
}
````

this output will be used to automatically generate a blueprint/swagger.

ouziel-slama:
moved here https://github.com/CounterpartyXCP/counterparty-core/pull/1716

PR #1632, opened 2024-04-11
Trigger publish docker image on push in `publish-docker-image` branch

ouziel-slama:


adamkrellenstein (approved):
Does it *only* do the docker image build and push on pushes to this branch or also to `master`?

ouziel-slama:
only on pushes to the 'publish-docker-image'.. I'm not sure this is a good idea anymore after all.. you can do it manually whenever necessary.. less risk of publishing an image by mistake.. thoughts?

adamkrellenstein:
Yeah I mean I think if we can do it manually then that's sufficient—no need for the extra branch.

PR #1633, opened 2024-04-11
Fix egde case

ouziel-slama:
In the `order.expire()` function we first expire the `orders` and then the `order_matches`.
It can therefore happen that an order with the status `filled`, and therefore not expired, is re-opened because of an expired `order_match`.

In v9 we used `WHERE expire_index < :block_index` and therefore the order was expired at the next block (this is why the order `e928918dd9a0605dc737eca5034cc444080d20f4ff7a7090b11f10d2a8e637e6` is expired at block 835513 instead of being expired at block 835512).

In v10, for optimization reasons and because in theory a `<` should not be necessary, we use `WHERE expire_index = :block_index - 1`. And therefore the order is not expired in the next block.

To reproduce this edge case while optimizing performance we now also check `WHERE expire_index = :block_index - 2`.
The correct fix should be to first expire `order_matches` and then expire `orders` but this is a protocol change. I will open an issue for this.

### DIFF
--- a/counterparty-core/requirements.txt
+++ b/counterparty-core/requirements.txt
@@ -5,4 +5,4 @@ colorlog==6.8.0
 python-dateutil==2.8.2
 requests==2.31.0
 termcolor==2.4.0
-counterparty-lib==10.0.0
+counterparty-lib==10.0.1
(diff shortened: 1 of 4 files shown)


ouziel-slama:
Fix https://github.com/CounterpartyXCP/counterparty-core/issues/1631

adamkrellenstein (approved):

PR #1644, opened 2024-04-12
Add checkpoint for block 838863

ouziel-slama:
Same hashes as https://www.xcp.dev/block/838863 and https://xchain.io/block/838863 running v9


ouziel-slama:
Testnet checkpoint comes from the new bootstrap db.

adamkrellenstein (approved):

PR #1664, opened 2024-04-15
No mempool flag

ouziel-slama:


adamkrellenstein:
I think we should also log a WARNING that the mempool is disabled.

adamkrellenstein (approved):

ouziel-slama:
> I think we should also log a WARNING that the mempool is disabled.

yep! done!

Issue #1315, opened 2024-01-14
Re-Enable DB Integrity Check


adamkrellenstein:
#1192 

Can we do this in the background in a separate thread?

adamkrellenstein:
Hmm, what if we just used `PRAGMA schema.quick_check;` instead? https://www.sqlite.org/pragma.html#pragma_quick_check

And maybe a full integrity check after bootstrap.

I don't know if we need a separate command, since the user can just do the integrity check with the `sqlite3` REPL if they want.

ouziel-slama:
FWIW there is already command `check-db` which do an integrity check.

PR #1674, opened 2024-04-15
run PRAGMA quick_check when initializing db (#1315)

mattmarcello:
This takes a minute or two when run initially.  Subsequent runs are faster.  I could add a flag to skip it?


adamkrellenstein:
I think it's good!

PR #1677, opened 2024-04-15
wip

mattmarcello:


mattmarcello:
Presently we are not wired into any backend ( aside from sentry ), but all the glue is there to do so easily.  

Also, I will open a separate ticket for rpc call exceptions not being propagated to sentry.

counterparty-core/counterpartycore/server.py
```diff
@@ -576,6 +579,14 @@ def start_all(catch_up="normal"):
...
         db = initialise_db()
 
+        telemetry_daemon = TelemetryDaemon(
+            interval=60,
+            collector=TelemetryCollectorLive(db=db),
+            client=TelemetryClientLocal(),
+        )
+
+        telemetry_daemon.start()
+
```
ouziel-slama, line 589: shouldn't we call `.stop()` in `finally:` clause ?

ouziel-slama (approved):
Nice!

counterparty-core/counterpartycore/server.py
mattmarcello: Good catch.  Yes!

counterparty-core/counterpartycore/lib/api.py
```diff
@@ -1069,6 +1063,12 @@ def get_running_info():
                 "version_major": config.VERSION_MAJOR,
                 "version_minor": config.VERSION_MINOR,
                 "version_revision": config.VERSION_REVISION,
+                "addrindexrs_version_major": int(addrindexrs_version[0]),
+                "addrindexrs_version_minor": int(addrindexrs_version[1]),
+                "addrindexrs_version_revision": int(addrindexrs_version[2]),
+                "uptime": int(get_uptime()),
+                "docker": is_docker(),
```
adamkrellenstein, line 1070: I'd probably call this `dockerized`

counterparty-core/counterpartycore/lib/sentry.py
```diff
@@ -0,0 +1,43 @@
...
+    event["tags"]["network"] = data["network"]
+    event["tags"]["force_enabled"] = data["force_enabled"]
+
+    event["extra"] = event.get("extra", {})
+    event["extra"]["last_block"] = data["last_block"]
+
+    return event
+
+
+sentry_sdk.init(
```
adamkrellenstein, line 37: Can we record the hostname for Sentry? (but *not* for the node telemetry, which is on by default)

counterparty-core/counterpartycore/lib/telemetry/util.py
```diff
@@ -0,0 +1,43 @@
+import os
```
adamkrellenstein, line 1: What ReST endpoint should we expose this at? (@ouziel-slama)

adamkrellenstein (approved):

counterparty-core/counterpartycore/lib/telemetry/util.py
ouziel-slama: hum.. for now I put the equivalent of `get_running_info` at the root path '/'. We could have `/server` resource also.

adamkrellenstein: I think the root path might be sufficient! Unless there's some standard I don't know about...

counterparty-core/counterpartycore/server.py
mattmarcello: Actually, yeah, I think it's cleaner just to not handle it explicitly.  Otherwise i need to do something like this in `stop`:

```
    def stop(self):
        self.is_running = False
        try :
            self.thread.join()
        except KeyboardInterrupt:
            pass
```

counterparty-core/counterpartycore/lib/sentry.py
mattmarcello: Sentry automatically picks up hostname so I think we currently have the behavior you want.

adamkrellenstein: ah hm, I'm seeing `API-SERVER-{H,J,5,6}` etc.

counterparty-core/counterpartycore/lib/telemetry/util.py
adamkrellenstein: yep, that sounds good!

counterparty-core/counterpartycore/lib/sentry.py
mattmarcello: That is the error ID.  API-SERVER is the name of the project and the alphanumeric chars are the unique ID for a given error class ( as opposed to instance ).  We could rename API-SERVER to whatever we want. the hostname ( i.e. api3 or api4 ) is captured under `server_name` tag.

counterparty-core/counterpartycore/lib/telemetry/util.py
adamkrellenstein: ah great! I didn't see the server name

PR #1735, opened 2024-04-27
Increase default poll interval

ouziel-slama:


adamkrellenstein (approved):

counterparty-core/counterpartycore/server.py
```diff
@@ -303,6 +303,7 @@ def initialise_config(
 
     # Backend Poll Interval
     if backend_poll_interval:
+        assert backend_poll_interval >= 3.0
```
adamkrellenstein, line 306: @ouziel-slama can we specify the minimum value in the CLI `--help` and exit gracefully if the condition isn't met?

ouziel-slama: sure!

PR #1744, opened 2024-04-28
INFO on stdout but everything else on stderr

ouziel-slama:
fix https://github.com/CounterpartyXCP/counterparty-core/issues/1733


ouziel-slama:
Better to keep the default!

PR #1754, opened 2024-04-30
Ready if the last block is less than 1 minute old

ouziel-slama:
- [x] API v1
- [x] API v2

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/util.py
+++ b/counterparty-core/counterpartycore/lib/api/util.py
@@ -2,6 +2,7 @@ import decimal
 import inspect
 import json
 import logging
+import time
 from logging import handlers as logging_handlers
 
 import flask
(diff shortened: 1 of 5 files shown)


counterparty-core/counterpartycore/lib/api/util.py
```diff
@@ -11,17 +12,20 @@
 logger = logging.getLogger(config.LOGGER_NAME)
 
 
-def check_last_parsed_block(blockcount):
+def check_last_parsed_block(db, blockcount):
     """Checks database to see if is caught up with backend."""
+    last_block = ledger.get_last_block(db)
```
adamkrellenstein, line 17: two minutes or one minute? Am I reading this right?

adamkrellenstein: (two minutes seems like a lot!)

release-notes/release-notes-v10.1.2.md
```diff
@@ -19,6 +19,7 @@ To easily migrate to the new API, an equivalence table is available in the docum
 
 ## Codebase
 * New ReST API
+* APIs return ready if the last block is less than 2 minutes old
```
ouziel-slama, line 23: oh got it lol! fixing that!

counterparty-core/counterpartycore/lib/api/util.py
ouziel-slama: fixed the title and the release notes.

PR #1755, opened 2024-05-01
Add test to compare local hashes with a list of servers

ouziel-slama:
- this test is executed only with the flag `--comparehashes`
- this test needs a running local mainnet server
- will be executed after `dredd` in the Docker Compose test

### DIFF
--- a/counterparty-core/counterpartycore/test/compare_hashes_test.py
+++ b/counterparty-core/counterpartycore/test/compare_hashes_test.py
@@ -1,0 +1,59 @@
+import json
+
+import pytest
+import requests
+
+from counterpartycore.test import (
+    conftest,  # noqa: F401
+)
+
+# This two servers must be up and running
+API_V9_URL = "http://rpc:rpc@api1.counterparty.io:4000"
+API_V10_URL = "http://api:api@localhost:4000"
+
+
+def test_compare_hashes(skip):
+    if skip:
+        pytest.skip("Skipping test book")
+        return
+
+    # get v9 last block
+    headers = {"content-type": "application/json"}
+    payload = {
+        "method": "get_running_info",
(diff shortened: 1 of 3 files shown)


adamkrellenstein (changes requested):
This looks great, except I think we should test *all* supported versions and not just v9 vs v10. (That is, I think we should test v9.61.3, v10.00, v10.0.1, etc.)

ouziel-slama:
> This looks great, except I think we should test _all_ supported versions and not just v9 vs v10. (That is, I think we should test v9.61.3, v10.00, v10.0.1, etc.)

yep that's mean to setup an instance for each version.. it's possible right now.. just need to add a list of servers in `CHECK_SERVERS` variable. `v9` means here to use the API v1 and `v10` means to use the API v2. I will fix that.

counterparty-core/counterpartycore/test/compare_hashes_test.py
```diff
@@ -0,0 +1,87 @@
...
+
+from counterpartycore.test import (
+    conftest,  # noqa: F401
+)
+
+LOCAL_API_URL = "http://api:api@localhost:4000"
+
+# [server_url, api_version]
+CHECK_SERVERS = [
+    ["http://rpc:rpc@api1.counterparty.io:4000", "v1"],  # v9.61.3
```
adamkrellenstein, line 14: hmm I think we should record which version is supposed to be running on each server and have this test verify that before passing

adamkrellenstein (approved):

counterparty-core/counterpartycore/test/compare_hashes_test.py
ouziel-slama: yep good idea! done!
(so api1 is running v9.61.1)

PR #1756, opened 2024-05-01
Inject details when ?verbose=true

ouziel-slama:


ouziel-slama:
comes with https://github.com/CounterpartyXCP/Documentation/pull/195

counterparty-core/counterpartycore/lib/api/util.py
```diff
@@ -229,3 +239,133 @@ def default(self, o):
...
+    issuance_by_asset = ledger.get_assets_last_issuance(db, asset_list)
+
+    # inject issuance
+    for result_item in result_list:
+        item = result_item
+        if "params" in item:
+            item = item["params"]
+        for field_name in ["asset", "give_asset", "get_asset"]:
+            if field_name in item and item[field_name] in issuance_by_asset:
+                item[field_name + "_issuance"] = issuance_by_asset[item[field_name]]
```
adamkrellenstein, line 280: I think we should use `get_asset_info` and `give_asset_info`.

adamkrellenstein (approved):

PR #1759, opened 2024-05-02
Add TRACE log level

ouziel-slama:


ouziel-slama:
The `--verbose` flag can be repeated to increase verbosity. `-vv` is also supported.

ouziel-slama:
`counterparty-server start -v` set log level to DEBUG
`counterparty-server start -vv` to TRACE

counterparty-core/counterpartycore/cli.py
```diff
@@ -33,9 +33,9 @@ def float_range_checker(arg):
         ("-v", "--verbose"),
         {
             "dest": "verbose",
-            "action": "store_true",
-            "default": False,
-            "help": "sets log level to DEBUG",
+            "action": "count",
+            "default": 0,
+            "help": "verbose output (repeat for increased verbosity)",
```
adamkrellenstein, line 38: Let's specify that `-v` means `DEBUG` and `-vv` means `TRACE`... otherwise someone might try `-vvv` :)

adamkrellenstein (approved):

counterparty-core/counterpartycore/cli.py
ouziel-slama: done

PR #1763, opened 2024-05-03
Sentry: Filter '/healthz' transactions + Enable in API v2

ouziel-slama:


counterparty-core/counterpartycore/lib/api/api_server.py
```diff
@@ -244,6 +255,8 @@ def handle_route(**kwargs):
 
 
 def run_api_server(args):
+    logger.info("Starting API Server v2.")
```
adamkrellenstein, line 258: I think we shouldn't specify the version here. It's just the default now :)

counterparty-core/counterpartycore/lib/sentry.py
```diff
@@ -29,6 +31,17 @@ def before_send(event, _hint):
     return event
 
 
+def filter_transactions(event, _hint):
```
adamkrellenstein, line 34: Do we need to filter client-side? Sentry says it'll do it server-side.

adamkrellenstein (approved):

counterparty-core/counterpartycore/lib/api/api_server.py
ouziel-slama: :-) yes!

ouziel-slama: done

counterparty-core/counterpartycore/lib/sentry.py
ouziel-slama: yes we could just rename the transaction with a filtered name but for bandwith usage perhaps better to filter here ?

ouziel-slama: I renamed the routes to "healthcheck", so should be filtered server side, and removed the filter client side.

adamkrellenstein: eh I think it's a tiny amount of bandwidth, and it's only for servers we host ourselves, really. Also if we don't filter client-side, we can turn the filter off server-side later if we want.

Issue #1721, opened 2024-04-21
Clean Rust panic error


ouziel-slama:
during testnet kickstart:

```
src/utils.rs:46:32:
range end index 22 out of range for slice of length 17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

warrenpuffett:
Ok so this is coming from:

```rust
let address = Address {
    payload: Payload::WitnessProgram {
        version,
        program: script[2..22].to_vec(),
    },
    network: network_enum,
}
```

specifically `script[2..22]`. It relies on `script.is_witness_program()` to ensure the length so it can assume it is at least of len 22. However, looking at the implementation of this method, it only guarantees that the data portion of the script is between 2 and 40 bytes long. 

```rust
/// Checks whether a script pubkey is a Segregated Witness (segwit) program.
#[inline]
pub fn is_witness_program(&self) -> bool {
    // A scriptPubKey (or redeemScript as defined in BIP16/P2SH) that consists of a 1-byte
    // push opcode (for 0 to 16) followed by a data push between 2 and 40 bytes gets a new
    // special meaning. The value of the first push is called the "version byte". The following
    // byte vector pushed is called the "witness program".
    let script_len = self.0.len();
    if !(4..=42).contains(&script_len) {
        return false
    }
    let ver_opcode = opcodes::All::from(self.0[0]); // Version 0 or PUSHNUM_1-PUSHNUM_16
    let push_opbyte = self.0[1]; // Second byte push opcode 2-40 bytes
    WitnessVersion::try_from(ver_opcode).is_ok()
        && push_opbyte >= opcodes::all::OP_PUSHBYTES_2.to_u8()
        && push_opbyte <= opcodes::all::OP_PUSHBYTES_40.to_u8()
        // Check that the rest of the script has the correct size
        && script_len - 2 == push_opbyte as usize
}
```

@ouziel-slama I'm not sure what to do here as I don't understand the history around:

```
// This block below is necessary to reproduce a prior truncation bug in the python codebase.
```

ouziel-slama:
@warrenpuffett is really just a logging problem! I think we just need to catch the error and raise a Python ValueError instead a Panic Error.

warrenpuffett:
@ouziel-slama Thanks for the info. I will make that change. I just wanted to be sure that there isn't a bigger issue here.

PR #1773, opened 2024-05-05
Avoid rust panic

warrenpuffett:
#1721

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@ env
 venv
 docs/_build/*
 target/
+typings/
+pyrightconfig.json
+.DS_Store
 
 profile.txt
 
--- a/counterparty-rs/src/utils.rs
+++ b/counterparty-rs/src/utils.rs
@@ -40,10 +40,14 @@ fn script_to_address(script_pubkey: Vec<u8>, network: &str) -> PyResult<String> 
             Err(_) => return Err(PyErr::new::<pyo3::exceptions::PyValueError, _>("Invalid version value")),
         };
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -2,6 +2,9 @@ env
 venv
 docs/_build/*
 target/
+typings/
+pyrightconfig.json
```
warrenpuffett, line 6: My editor use pyright to link/typecheck so need to ignore some config files related to it.

counterparty-rs/src/utils.rs
```diff
@@ -40,10 +40,14 @@ fn script_to_address(script_pubkey: Vec<u8>, network: &str) -> PyResult<String>
             Err(_) => return Err(PyErr::new::<pyo3::exceptions::PyValueError, _>("Invalid version value")),
         };
 
+        let n = 22;
+        if script.len() < n{
```
ouziel-slama, line 44: hum.. should not be here 24 ? because we take script[2..22] ?

warrenpuffett: @ouziel-slama 22 is an index not the number of bytes to take: https://doc.rust-lang.org/std/ops/struct.Range.html. We're taking 20 bytes here, from the byte at index 2 up to the byte at index 22 (exclusive). We skipped 2 bytes so the length of the script must at least be 22.

ouziel-slama: ah right!! I forgot that you should never do a review before the second coffee lol!

Issue #779, opened 2015-05-14
Checksum Bootstrap Files after Download


adamkrellenstein:
At least do a checksum; ideally check a signature

adamkrellenstein:
Or should we just build and push with GitHub Actions??

warrenpuffett:
We will add signature verification logic to the server's bootstrap downloader. New bootstrap files will come with a signature file manually created by the team.

warrenpuffett:
@adamkrellenstein got a draft PR for this that adds a `verify_signature` util function and a corresponding test. I still need to add this to the `bootstrap` server function. First though, I need the contents of the public key we will use to create signatures. I'm thinking we just hardcode the contents of the public key file in the code.

We can use this command for creating signatures:

```
gpg --output snapshot.sig --detach-sign snapshot.tar.gz
```

And you can use this command to "export" the public key:

```
gpg --armor --export key-id > public_key.asc
```

adamkrellenstein:
[AD52047D.asc.txt](https://github.com/CounterpartyXCP/counterparty-core/files/15218073/AD52047D.asc.txt)

(I had to add the `.txt` extension so GitHub would let me upload the file.)

adamkrellenstein:
@ouziel-slama would you upload yours?

warrenpuffett:
Ok the PR has been updated to integrate the signature verification into bootstrap. Currently this is causing one of our tests to fail: https://github.com/CounterpartyXCP/counterparty-core/actions/runs/8977198956/job/24655491058?pr=1774

This is because we do not have the signature uploaded for whatever snapshot this is downloading.  Could someone please do that for me? Thanks!

ouziel-slama:
@warrenpuffett done! I uploaded sig files and also added my public key. Test works now.

PR #1774, opened 2024-05-05
Snapshot verification

warrenpuffett:
#779


ouziel-slama (approved):

adamkrellenstein:
Do we want to be able to bypass this with `—force`?

adamkrellenstein:
'k, I'm fine with leaving it out.

PR #1776, opened 2024-05-06
Refactor Mempool, Follow and Backend

ouziel-slama:


counterparty-core/counterpartycore/lib/mempool.py
```diff
@@ -0,0 +1,78 @@
...
+import time
+
+from counterpartycore.lib import blocks, config, exceptions, ledger
+from counterpartycore.lib.kickstart import blocks_parser
+
+logger = logging.getLogger(config.LOGGER_NAME)
+
+
+def parse_mempool_transactions(db, raw_tx_list):
+    logger.debug("Parsing mempool block")
```
adamkrellenstein, line 11: Let's not call it a "mempool block". It's not a block. (I think it was a bad idea for me to have implemented it that way originally.)

counterparty-core/counterpartycore/lib/mempool.py
```diff
@@ -0,0 +1,78 @@
...
+    now = time.time()
+    transaction_events = []
+    cursor = db.cursor()
+    try:
+        with db:
+            # insert fake block
+            cursor.execute(
+                """INSERT INTO blocks(
```
adamkrellenstein, line 19: Is there a way to do the mempool without having this virtual block? It seems to me that was a bad way to do it originally. (My fault!)

ouziel-slama: nope.. because of the foreign key constraints and the block_index field present in all tables we are obliged to have a corresponding block when we parse a mempool transaction.. it is also the most practical way to parse several transactions of the mempool at a time. I modified the "Parsing mempool block" log with "Parsing {len(raw_tx_list)} raw transactions from mempool...".
Of course this block, like all the mempool txs are never really recorded in the database.. we only retrieve the events and cancel the SQL transaction.

ouziel-slama:
Ok a lot of stuff in this PR.

The `blocks.follow()` function has been removed and replaced with:

     - The `blocks.catch_up()` function whose only role is to catch the last block.
     - The `follow.py` module which then takes over to follow the blockchain.

The `follow.py` module uses ZMQ to receive a notification for each transaction that enters the mempool and for each new block.

The `follow.py` module uses the `mempool.py` module to parse the mempool transactions in groups of 100 and to clean it with each new block.

The `follow.py` module, the `blocks.catch_up()` function (and also the kickstart) use the `blocks.parse_new_block()` function to parse the new blocks. Whatever the parsing method, this function must be used to add a new block.

`blocks.parse_new_block()` uses the new module `deserialize.py` (good candidate to be ported to Rust). This module is extracted from the kickstart to avoid a circular dependency and also to ensure that whatever the parsing method (catch_up, follow, kickstart) we always use the same deserializer.
It is also this function, and it alone, which is responsible for incrementing `CURRENT_BLOCK_INDEX` and managing the reorganization of the Blockchain.

The `prefetch.py` module is responsible for managing the block and transaction cache. It contains two functions `get_decoded_block(block_index)` and `get_decoded_tx(tx_hash)` which are used systematically in the code. This once again ensures that we use the same deserialization, the same cache, etc. everywhere.

The `backend` module has been completely refactored:
- The `__init__.py` file is empty.
- The `bitcoind.py` module contains functions to make RPC calls to Bitcoin Core
- The `addrindexrs.py` module contains functions to make calls to Addrindexrs
(The `addrindexrs.py` module also contains functions to make RPC calls to `bitcoind` in batch. I left these functions here because they are only used by functions inside `addrindexrs.py` and that we can delete them too.)

I added a `tx_hash` field in the `messages` table. This field allows you to easily retrieve all events for a given transaction in order to save them (or delete them) from the `mempool` table. Along the way, I added a route `/v2/transactions/<tx_hash>/events`.

I need to update the doc and release notes and test it again but other than that I think it's finished.

adamkrellenstein (approved):
❤️

PR #1792, opened 2024-05-15
Dispenser must be created by source

ouziel-slama:
Fix https://github.com/CounterpartyXCP/counterparty-core/issues/1785
For the moment I set from block 900000, we can change before the release.

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/dispenser.py
+++ b/counterparty-core/counterpartycore/lib/messages/dispenser.py
@@ -215,6 +215,12 @@ def validate(
         problems.append(
             f"address doesn't has enough balance of {asset} ({available[0]['quantity']} < {escrow_quantity})"
         )
+    elif (
+        util.enabled("dispenser_must_be_created_by_source")
+        and open_address is not None
+        and source != open_address
+    ):
+        problems.append("dispenser must be created by source")
     else:
         if status == STATUS_OPEN_EMPTY_ADDRESS and not (open_address):
             open_address = source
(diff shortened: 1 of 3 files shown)


ouziel-slama:
So the idea is that the status of dispensers can no longer be equal to `STATUS_OPEN_EMPTY_ADDRESS`.
Two possibilities:

- `status == STATUS_OPEN_EMPTY_ADDRESS` and `source != open_address`. In this case we throw an error.
- `status == STATUS_OPEN_EMPTY_ADDRESS` and `source == open_address`. In this case we set status to `STATUS_OPEN`.

Since `STATUS_OPEN_EMPTY_ADDRESS` is no longer possible we never enter in the `if` where `get_oldest_tx()` is called.

adamkrellenstein (approved):

adamkrellenstein:
FYI, this is a very sensitive issue for the community, so we have to be super duper extra sure that we get this implementation correct and that there are no unplanned / undocumented regressions. We'll need tutorials for migrating to new workflows in the docs, explanation in the release notes, etc.

This will encourage users to have more than one dispenser on the same address, so we'll need to make sure that the new `create_dispense()` allows for buying multiple assets simultaneously.

adamkrellenstein (changes requested):
I actually think it'd be best if we include all dispenser protocol changes in a single PR—then we can more easily review everything together

loon3:
> * `status == STATUS_OPEN_EMPTY_ADDRESS` and `source == open_address`. In this case we set status to `STATUS_OPEN`.

when would this statement ever be true?

A couple other notes...

- activation of this feature will be tricky unless you have a sort of staged rollout because someone could theoretically open a dispenser at block 899999 on an empty address meaning you won't know what the csv will need to be until the activation block.
- there should be a way, post activation, to close existing dispensers that were opened on an empty address so that the assets transfer back to origin address upon close (not sure if this is the default behavior or if you can currently close from either dispenser address or origin address)

PR #1799, opened 2024-05-16
API Tweaks

ouziel-slama:
1. New routes

`/v2/assets/<asset>/dispenses`
`/v2/addresses/<address>/dispenses/sends`
`/v2/addresses/<address>/dispenses/receives`
`/v2/addresses/<address>/dispenses/sends/<asset>`
`/v2/addresses/<address>/dispenses/receives/<asset>`
`/v2/transactions/<int:tx_index>/events`
`/v2/transactions/<tx_hash>/events`
`/v2/transactions/<int:tx_index>/events/<event>`
`/v2/transactions/<tx_hash>/events/<event>`
`/v2/transactions/<int:tx_index>`
`/v2/blocks/<block_hash>`
`/v2/blocks/last`
`/v2/mempool/transactions/<tx_hash>/events`
`/v2/addresses/<address>/issuances`

2. Pagination

- Replace `last` arguments by `cursor`
- All responses contain a `next_cursor` field.
- All responses contain a `result_count` field.
- All queries that return lists from the database accept the "cursor" and "limit" arguments or "offset" and "limit" arguments

3. Miscellaneous

- Documents the list of events with an example format for each of them.
- Updated `rowtracer` so that `apsw` returns boolean instead of integer for `BOOL` type fields.
- The `asset`, `assets`, `give_asset` and `get_asset` parameters are not case sensitive.


ouziel-slama:
OP updated.

adamkrellenstein (approved):
❤️

PR #1811, opened 2024-05-23
Handle correctly invalid RPC call; Accept trailing slashes

ouziel-slama:
Fix https://github.com/CounterpartyXCP/counterparty-core/issues/1810


adamkrellenstein:
I think we should move away from raising exceptions due to API calls? Also here:
image

But we can do that later.

adamkrellenstein (approved):

ouziel-slama:
> I think we should move away from raising exceptions due to API calls? Also here: image
> 
> But we can do that later.

yep!!
I added here https://github.com/CounterpartyXCP/counterparty-core/pull/1811/commits/47557277676ddef7f5c13725a7c2898d80a2473c all excepted error!

PR #1822, opened 2024-05-25
ZMQ Server

ouziel-slama:
- Publish all events with ZMQ
- Each event is published on a specific topic
- The events are in JSON format:
```
{
    "event":"DEBIT",
    "event_index":17549202,
    "params":{
       "block_index":845084,
       "address":"1A9snquSqbdW4qnGQ5qmzJG6w1QWEdG16H",
       "asset":"GIVEKUDOS",
       "quantity":1,
       "action":"send",
       "event":"0a2fb26f5237e3d0b7b4704e1f899381fc91de6e41f1e9744759a5d6169b6200",
       "tx_index":2734371
    },
    "mempool":false,
    "description":"Debit 1 GIVEKUDOS from 1A9snquSqbdW4qnGQ5qmzJG6w1QWEdG16H",
    "block_index":845084
}
```
- the mempool events do not contain `block_index` and `event_index` fields and the mempool field is equal to True.
- `tools/zmqclient.py` contains a working example in Python.
- It remains to add a paragraph in the documentation.


adamkrellenstein (approved):

ouziel-slama:
> This looks good. Just one thought: should we consider a design that will allow a user to query all messages? This could be useful for doing things like building an event feed. 

nah that's not the philosophy behind the Pub/Sub model and that will add too much complexity server side. The idea is to use the REST API to catch up and then ZMQ to watch particular events.

> I haven't worked with ZMQ but I'd wager the subscribe API would support some kind of filtering that would enable us to narrow to specific message type from a general message channel.

yep! see https://github.com/CounterpartyXCP/Documentation/blob/master/docs/advanced/api-v2/node-api.md#zmq-publisher

PR #1833, opened 2024-05-30
Fix redirection to API v1

ouziel-slama:
should fix https://github.com/CounterpartyXCP/counterparty-core/issues/1816


adamkrellenstein (approved):
Why was it non-deterministic? Do you know?

ouziel-slama:
> Why was it non-deterministic? Do you know?

Nope! I don't understand why they don't get **always** result in result..  
hum.. perhaps a load balancer with different versions behind ?

PR #1835, opened 2024-05-30
Add events filter

ouziel-slama:
Use `/v2/events?only=EVENT1,EVENT2,..` to filter events.


adamkrellenstein:
maybe instead of `only` we call the filter `event_name`?

adamkrellenstein (approved):

ouziel-slama:
> maybe instead of `only` we call the filter `event_name`? in case we want to add more filters later

done!

adamkrellenstein:
@ouziel-slama we also need a line in the release notes!

Issue #1830, opened 2024-05-29
Update pyo3


warrenpuffett:
pyo3 has been updated and its interface has changed. As this is a critical dependency which allows the python process to communicate with the rust process we do not want to fall behind.

PR #1838, opened 2024-05-31
Update pyo3

warrenpuffett:
#1830

### DIFF
--- a/counterparty-core/counterpartycore/lib/script.py
+++ b/counterparty-core/counterpartycore/lib/script.py
@@ -179,7 +179,7 @@ def base58_check_decode_py(s, version):
 
 def base58_check_decode(s, version):
     try:
-        decoded = b58.b58_decode(s)
+        decoded = bytes(b58.b58_decode(s))
     except ValueError:
         raise Base58Error("invalid base58 string")  # noqa: B904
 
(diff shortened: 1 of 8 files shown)


counterparty-core/counterpartycore/lib/script.py
```diff
@@ -179,7 +179,7 @@ def base58_check_decode_py(s, version):
 
 def base58_check_decode(s, version):
     try:
-        decoded = b58.b58_decode(s)
+        decoded = bytes(b58.b58_decode(s))
```
warrenpuffett, line 182: Now that we are relying on pyo3's [mapping of rust types to python types](https://pyo3.rs/v0.21.2/conversions/tables) we need to cast this list of bytes to bytes here.

counterparty-rs/src/b58.rs
```diff
@@ -1,45 +1,42 @@
...
     match decoded {
-        Ok(s) => Ok(PyBytes::new(py, &s)),
+        Ok(s) => Ok(s),
         Err(_) => Err(PyValueError::new_err("Bad input")),
     }
 }
 
 #[pyfunction]
-fn b58_decode_list<'p>(py: Python<'p>, encoded_list: Vec<&str>) -> PyResult<Vec<&'p PyBytes>> {
+fn b58_decode_list(encoded_list: Vec<String>) -> PyResult<Vec<Vec<u8>>> {
```
warrenpuffett, line 25: Using rust types here's significantly simplifies the code as we don't have to do any casting or introduce lifetime parameters.

ouziel-slama (approved):

PR #1846, opened 2024-05-31
Minimalistic connection pool

ouziel-slama:
Fix https://github.com/CounterpartyXCP/counterparty-core/issues/1817


adamkrellenstein (approved):
Nice!

Issue #1831, opened 2024-05-29
Manually deserialize block object to PyDict


warrenpuffett:
Right now we are relying on `json.loads` in the python process to deserialize block objects in `fetcher.get_block`. This is causing a significant performance hit. The underlying rust method should instead return a pyo3 PyDict instead of a json string. Depends on #1830.

PR #1847, opened 2024-06-01
Rust fetcher now returns a Dict instead of JSON bytes

warrenpuffett:
#1831

### DIFF
--- a/counterparty-rs/src/indexer/block.rs
+++ b/counterparty-rs/src/indexer/block.rs
@@ -1,5 +1,5 @@
 use pyo3::{
-    types::{PyAnyMethods, PyDict},
+    types::{PyAnyMethods, PyBytes, PyDict},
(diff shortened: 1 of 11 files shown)


counterparty-rs/src/indexer/block.rs
```diff
@@ -1,20 +1,46 @@
...
+        dict.set_item("sequence", self.sequence).unwrap();
+        dict.set_item("script_sig", self.script_sig).unwrap();
+        dict.unbind().into()
+    }
+}
+
+#[derive(Clone)]
 pub struct Vout {
     pub value: u64,
     pub script_pub_key: String,
```
ouziel-slama, line 29: is it possible to return bytes here ? also for `script_sig` ?

ouziel-slama (approved):

ouziel-slama:
❤️

PR #1848, opened 2024-06-01
Optimize `get_pending_dispensers()`

ouziel-slama:
Need a reparse from block 819300. So we need to bump minor version.


adamkrellenstein:
k

adamkrellenstein (approved):
Release notes! ❤️

PR #1869, opened 2024-06-05
Fixes and tweaks

ouziel-slama:
* Fix non-cacheable API v2 routes.
* Fix the execution interval of the `APIStatusPoller` thread.
* Fix `docker compose stop`
* Don't run quick check
* Add `first_issuance_block_index` and `last_issuance_block_index` in assets information.


.github/workflows/test_compose.sh
```diff
@@ -98,7 +98,7 @@ REPARSE_FROM=$(($CURRENT_HEIGHT-50))
 
 # Stop, reparse and start counterparty-core mainnet
 docker compose --profile mainnet stop counterparty-core
-docker compose --profile mainnet run counterparty-core reparse $REPARSE_FROM \
+docker compose --profile mainnet run counterparty-core reparse $REPARSE_FROM --force \
```
adamkrellenstein, line 101: why are we doing `--force` here?

counterparty-core/counterpartycore/server.py
```diff
@@ -782,6 +783,16 @@ def start_all(args):
...
         fetcher.stop()
+        try:
+            database.check_wal_file()
+        except exceptions.WALFileFoundError:
+            logger.error(
+                "Database WAL file detected. To ensure no data corruption has occurred, run `counterpary-server check-db`."
+            )
+        except exceptions.DatabaseError:
+            logger.error(
+                "Database in use by another process. Please run `counterpary-server check-db` to verify its integrity."
```
adamkrellenstein, line 794: Does this exception get thrown only when the DB is already open?

If it's already opened by another running process, how will running `check-db` help?

adamkrellenstein (changes requested):

counterparty-core/counterpartycore/server.py
ouziel-slama: yep!
to check if the other process didn't corrupt the db ?

adamkrellenstein: Hm, I don't follow. Why would the other process corrupt the DB? If there's another process using the DB (e.g. the `sqlite` CLI), seems like the correct soln is just to close it and try again.

ouziel-slama: we can't close a connection opened by another process (`sqlite3` for example). But I agree the message is not really relevant. Maybe juste a warning "another process is using the database, unable to close it correctly." ?

.github/workflows/test_compose.sh
ouziel-slama: removed

counterparty-core/counterpartycore/server.py
adamkrellenstein: yep! that's what I meant! just `"Database in use by another process. Please close it and try again."`

ouziel-slama: oh! you mean refuse to terminate `counterparty-server` until the other process is closed? hmm.. this can cause problems with `docker stop` which at one time or another will send a SIGKILL in any case.. It's generally annoying to have a program that refuses to quit...

adamkrellenstein: ah right this is just for shutdown! let's make the error: "Database is in use by another process and was unable to be closed correctly."

ouziel-slama: done

adamkrellenstein (approved):
❤️

counterparty-core/counterpartycore/lib/api/util.py
```diff
@@ -361,7 +361,7 @@ def inject_normalized_quantities(result):
                 continue
 
             is_divisible = True
-            if field_name not in ["burned", "earned"]:
+            if field_name not in ["burned", "earned", "btc_amount"]:
```
adamkrellenstein, line 364: did you update the release notes for this?

ouziel-slama: now yes :)

PR #1886, opened 2024-06-07
Fix ZMQ addresses with Docker

ouziel-slama:


adamkrellenstein (approved):
looks like the docker compose test is failing

ouziel-slama:
> looks like the docker compose test is failing

fixed. there is sometimes a timeout with dredd. On it.

PR #1890, opened 2024-06-09
Add `confirmation_target` argument

ouziel-slama:
```
$ curl 'http://localhost:14000/v2/addresses/n38bRTm24T8Hs4D6kxDKDTnX78wvoM98kS/compose/send?destination=tb1qutk7tnyhfxkp0grfp3nz69jech84yqlulffk0u&asset=BTC&quantity=5000' | jq
{
  "error": "Insufficient BTC at address n38bRTm24T8Hs4D6kxDKDTnX78wvoM98kS. (Need approximately 0.00437483 BTC.) To spend unconfirmed coins, use the flag `--unconfirmed`. (Unconfirmed coins cannot be spent from multi‐sig addresses.)"
}
```

Increase the the number of blocks to target for confirmation to reduce fees:

```
$ curl 'http://localhost:14000/v2/addresses/n38bRTm24T8Hs4D6kxDKDTnX78wvoM98kS/compose/send?destination=tb1qutk7tnyhfxkp0grfp3nz69jech84yqlulffk0u&asset=BTC&quantity=5000&confirmation_target=10' | jq
{
  "result": {
    "rawtransaction": "0100000001a7d080b98fc17e54e1db603206f226fbc3c5384b3420050c370bf24676bd96b9010000001976a914ed18315a862eab6395edff98b2ad5285f96dd46b88acffffffff028813000000000000160014e2ede5cc9749ac17a0690c662d1659c5cf5203fccb490000000000001976a914ed18315a862eab6395edff98b2ad5285f96dd46b88ac00000000",
    "params": {
      "source": "n38bRTm24T8Hs4D6kxDKDTnX78wvoM98kS",
      "destination": "tb1qutk7tnyhfxkp0grfp3nz69jech84yqlulffk0u",
      "asset": "BTC",
      "quantity": 5000,
      "memo": null,
      "memo_is_hex": false,
      "use_enhanced_send": true
    },
    "name": "send"
  }
}
```


ouziel-slama:
The number in "(Need approximately 0.00437483 BTC.)" comes from:

```
$ curl --user rpc:rpc --data-binary '{"jsonrpc": "1.0", "id": "curltest", "method": "estimatesmartfee", "params": [6]}' -H 
{
  "result": {
    "feerate": 0.00433884,
    "blocks": 6
  },
  "error": null,
  "id": "curltest"
}
```

adamkrellenstein (approved):
Does this really resolve the whole testnet problem? :)

adamkrellenstein:
Also, release notes!

ouziel-slama:
> Does this really resolve the whole testnet problem? :)

there is already a solution which is to use the `fee` argument but with this new argument this avoids having to calculate it..
Sometimes testnet is very very expensive and there is no other solution than to wait: wait before the fees drop to compose another transaction, or force fewer fees and in this case wait until the transaction is confirmed :)

PR #1893, opened 2024-06-10
Fixes

ouziel-slama:
- [x] fix CURRENT_BLOCK_INDEX on reorg
- [x] kill API server if necessary
- [x] don't truncate hashes in logs


adamkrellenstein:
> don't truncate hashes in logs

sorry what's the reasoning for this?

PR #1997, opened 2024-07-03
Fixes

ouziel-slama:
- [x] Catch TimeoutError on version checking; re-check on error
- [x] Catch correctly bitcoind error from an API call
- [x] Log v2 API Server Location
- [x] Fix `util.check_last_parsed_block()`
- [x] Add logging in `blocks.initialise(db)`

---------


counterparty-core/counterpartycore/lib/check.py
```diff
@@ -959,9 +959,14 @@ def software_version():
...
         )
         versions = json.loads(response.text)
-    except (requests.exceptions.ConnectionError, ConnectionRefusedError, ValueError) as e:  # noqa: F841
+    except (
+        requests.exceptions.ConnectionError,
+        ConnectionRefusedError,
+        ValueError,
+        TimeoutError,
+    ) as e:  # noqa: F841
         logger.warning("Unable to check version! " + str(sys.exc_info()[1]))
```
adamkrellenstein, line 968: Is this what we want to log?

adamkrellenstein (approved):

counterparty-core/counterpartycore/lib/check.py
ouziel-slama: nope! I think `logger.error("Unable to check version:" + str(e)`

adamkrellenstein: yeah I think `error`

adamkrellenstein (approved):

Issue #1764, opened 2024-05-03
Kill AddrIndexRs


adamkrellenstein:
### Consensus-Critical
* [X] Rollback the `dispenser_origin_permission_extended` protocol change so we can drop `get_oldest_tx()`

### TX Creation
* [ ] Replace `get_unspent_txouts()`
* [ ] Get rid of `searchrawtransactions()`

For TX creation, wallets should either make a call either to bitcoind (if the address in question is in the wallet) or to a public API or electrs. Counterparty can't return BTC balances for arbitrary addresses, because Bitcoin Core won't do it.

_Note_ this will only be possible once the v10.4.0 protocol change has actually gone into effect and we have had the opportunity to hard-code historical dispenser creations

**Update**: See https://github.com/CounterpartyXCP/counterparty-core/issues/2245 for information on how to modify transaction construction to work without AddrIndexRs.

warrenpuffett:
Leather wallet is using this API for getting utxos: https://github.com/Blockstream/esplora/blob/master/API.md#get-addressaddressutxo

We could use the rate limited public API or host an Esplora instance of our own.

jotapea:
Is this suggesting making the reference protocol implementation dependent on third party services? Or am I misunderstanding something?

Shouldn't the only data dependency to `counterparty-core` be Bitcoin (Core), no third party involved?

B0BSmiths:
Please correct me if i am wrong but a counterparty node needs to know about the UTXOs in order to build the a TX hex due to the use of arc4 when it comes to encoding data

Requiring users to dox themselves to 3rd party APIs in order to use Counterparty is a privacy nightmare and a massive downgrade over current functionality 

If we are encouraging wallets to not use the Counterparty API to build transactions then we are encouraging them to skip sanity checks the API provides

adamkrellenstein:
It will be like every other Bitcoin wallet. If you have a local node, you use only that. If you don't, you have to rely on someone else's node (in which case the information shared is the same info as is shared where you broadcast the TX you just created.)

Using those sanity checks would still be encouraged in this proposal. But they *are* fundamentally optional.

jdogresorg:
> * [ ]  Rollback the `dispenser_origin_permission_extended` protocol change so we can drop `get_oldest_tx()`

Can you point me to ANY documentation or discussion of removing this `origin` functionality... anything to indicate that there has been SOME discussions on this issue with community members.

[...]

PR #2140, opened 2024-08-08
Require Dispenser to be Source Address

ouziel-slama:
# Motivation

For an unknown reason, it has always been possible to open a dispenser on an address that you yourself do not own. Of course, this is a major security vulnerability as it can allow anyone on the network to force other users to sell assets without their consent (potentially with major legal consequences). Generally speaking, this is just a Bad Idea(TM).

A hacky workaround was implemented in November 2023 (with the unhelpful ChangeLog entry "Redefined EMPTY address to mean no XCP or BTC history”) which prevents dispensers from being opened on any address that does not have any XCP or BTC history. This probably resolves the security issue but [[that little patch](https://github.com/CounterpartyXCP/counterparty-core/pull/1255/files#diff-446a2aa52fd57fd6379f4732ae36f794df6f0fb1585494f798f4aa3f7b94fca2R136)] unintentionally (and retroactively!) made the entirety of AddrIndexRs (and its ~200GB database) consensus-critical and mandatory for parsing. AddrIndexRs is a broken, unmaintained fork of a third-party codebase that is *not* deterministic and has been the cause of a [[large number of critical bugs in Counterparty](https://github.com/CounterpartyXCP/counterparty-core/issues?q=is%3Aissue+addrindexrs+label%3Abug+)], not to mention the fact that it makes Counterparty deployment a total PITA. Thankfully, once this protocol change goes into effect, we can kill AddrIndexRs, per [[#1764](https://github.com/CounterpartyXCP/counterparty-core/issues/1764)](https://github.com/CounterpartyXCP/counterparty-core/issues/1764).

# Design

A protocol change will be added `dispenser_must_be_created_by_source` to the `protocol_changes.json` file. From this block onwards, the `compose.validate` function will verify that `source == open_address`.

The only known downside to this change is that it will require a moving tokens to a wallet (e.g. a cold wallet) before using that wallet as a dispenser. (By differentiating between `source` and `origin`  you can simulate Ordinals-like fair minting; but we’re going to be implementing an *actual* Fair Minting contract, so no functionality will be lost.)

## API Changes

*No API changes*.

## Database Changes

*No database changes*.

------------


niftyboss1:
I fully support this change. 
Question: will it impact dispensers where source<>origin that were created before the change goes into effect?

ouziel-slama:
moved here https://github.com/CounterpartyXCP/counterparty-core/pull/2141

PR #2153, opened 2024-08-09
Lockable Issuance Descriptions Protocol Change

warrenpuffett:
## Motivation

From a collectors point of view, it is often important to be sure that all characteristics of an asset are immutable to establish the permanency of the object being traded.

To lock an asset, there are currently two possibilities: perform an `issuance` with the description equal to "lock" or directly pass the parameter `lock=True`. When an Asset is locked, however, it is still possible to modify the description.

## Design

1. Added way to lock description: `description="lock_description"`
2. A `description_locked` field is added to the `issuances` table.
3. The `issuances.validate` is modified to check that when the `description_locked` issuance property is `True` the description is `None` and that issuances with locked descriptions cannot be reset.
4. A protocol change `lockable_issuance_descriptions` will be added in the `protocol_changes.json` file.

## API Changes

No API changes, except for the documentation to indicate that `description="lock_description"` causes description only to be locked.

## Database Changes

A `description_locked` column will be added to the `issuances` table.


counterparty-core/counterpartycore/lib/messages/issuance.py
```diff
@@ -769,6 +781,14 @@ def unpack(db, message, message_type_id, block_index, return_dict=False):
     )
 
 
+def _get_last_description(db, asset, default):
+    issuances = ledger.get_issuances(db, asset=asset, status="valid", first=True)
```
ouziel-slama, line 785: there is already a function `ledger.get_asset_description()` I think.

ouziel-slama (commented):
LGTM!

PR #2173, opened 2024-08-21
Rust fetcher now respect JSON_LOGS config option

warrenpuffett:


counterparty-rs/src/indexer/bitcoin_client.rs
```diff
@@ -1,24 +1,24 @@
-use std::{collections::HashMap, sync::Arc, thread::JoinHandle};
```
warrenpuffett, line 1: Changes in this file are just formatting and linting

adamkrellenstein (approved):

Issue #1842, opened 2024-05-31
Subassets on Numerics


adamkrellenstein:
https://github.com/mikeinspace/Glyphs/blob/main/README.md

reinamora137:
Yes please. Will certainly free up usage of the root level namespace with no apparent downside.

mikeinspace:
Definitely in favor of sub-assets on numerics. There's no reason to treat numerics as lesser than named assets in my view. 

XCP fee has been clarified to be an anti-squatting mechanism (not an anti-spam mechanism as some have previously claimed) and since you cannot really "squat" a numeric, then all the fee really does is add friction for users.

b1u3y:
Would subassets for numerics be an integer or a string?

If no XCP was paid for the main asset, should subassets be free too? This decision will greatly affect usage, and adoption. Free subasset mass adoption could lead to assets being tracked in code & db by main asset, instead of UTXO or address.

mikeinspace:
"If no XCP was paid for the main asset, should subassets be free too"

@b1u3y I think its important to understand the role of XCP within the context of asset registration. It isn't an "anti-spam" mechanism but, instead, an "anti-squat" mechanism. Therefore, I see no distinction between sub-assets on named assets vs numeric assets. You can't "squat yourself".

jotapea:
(Would have written here much earlier if I wasn't blocked, but [some already know](https://github.com/jotapea/counterparty-core/issues/3#issuecomment-2144003597) my opinion about this.)

---

I believe this change is detrimental to XCP, as I see clear downsides in the incentives for obtaining and utilizing the XCP token for named asset registration. Which is, as of today, still its main utility.

The tradeoff of being able to issue an asset XCP-free is that your name must be a number. Aside from this, a numeric asset (by itself) is treated exactly the same as any other non-numeric asset. And this is great!

But then, the users that spend/burn XCP to issue an asset, have the privilege of continuing to issue subassets related to their (super)asset. If this privilege is extended by making these subassets XCP-free, then, I believe, the main utility of the XCP token is elevated:

_XCP will be burned for the issuance of named **super-assets**._

So, I am in full agreement with [eliminating the fee for subassets](https://github.com/CounterpartyXCP/counterparty-core/issues/1840). But I'm against this one, because I believe it clearly lowers the utility of the XCP token.

---

> https://github.com/mikeinspace/Glyphs/blob/main/README.md

[...]

PR #2195, opened 2024-09-03
Allow subassets on numerics

warrenpuffett:
#1842 

## Motivation

There is no particular reason to disallow subassets on numeric assets and there is a strong community demand to be able to do so (see for example https://github.com/mikeinspace/Glyphs/blob/main/README.md).

## Design

Currently the `validate_subasset_parent_name` function checks that the parent asset is not numeric. This is the only place where this check is performed.

A protocol change `allow_subassets_on_numerics` will be added in the `protocol_changes.json` file. From this block on, the verification will be removed from the function `validate_subasset_parent_name`.

### DIFF
--- a/counterparty-core/counterpartycore/lib/util.py
+++ b/counterparty-core/counterpartycore/lib/util.py
@@ -178,7 +178,7 @@ def date_passed(date):
 # checks and validates subassets (PARENT.SUBASSET)
 #   throws exceptions for assset or subasset names with invalid syntax
 #   returns (None, None) if the asset is not a subasset name
-def parse_subasset_from_asset_name(asset):
+def parse_subasset_from_asset_name(asset, allow_subassets_on_numerics=False):
     subasset_parent = None
     subasset_child = None
     subasset_longname = None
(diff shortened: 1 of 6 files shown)


adamkrellenstein (dismissed):

ouziel-slama (approved):

counterparty-core/counterpartycore/lib/util.py
```diff
@@ -225,8 +225,21 @@ def validate_subasset_longname(subasset_longname, subasset_child=None):
...
+def is_numeric(s):
+    pattern = r"^A(\d{17,20})$"
+    match = re.match(pattern, s)
+    if match:
+        numeric_part = match.group(1)
+        numeric_value = int(numeric_part)
+        lower_bound = 26**12 + 1
+        upper_bound = 256**8
+
+        return lower_bound <= numeric_value <= upper_bound
```
ouziel-slama, line 237: Nice! I didn't know this syntax!

ouziel-slama (approved):

PR #2203, opened 2024-09-05
Tests, fixes and doc

ouziel-slama:
- [x] Document Atomic Swap (https://gist.github.com/ouziel-slama/88cc8a7674c4db37cce1bc8a3a38907a)
- [x] Test Atomic Swap
- [x] Test Atomic Swap with Counterparty TX
- [x] Add regetst tests for `issuance` protocol changes (`lockable_issuance_descriptions`, `free_subassets`, `allow_subassets_on_numerics`)
- [x] Fix minimum BTC amount in order and order match
- [x] Add regetst tests for `order` protocol changes (`expire_order_matches_then_orders`, `fix_min_btc_quantity`)
- [x] Add `quantity_by_price` in Fairminter

--------


adamkrellenstein (approved):

counterparty-core/counterpartycore/lib/messages/fairminter.py
```diff
@@ -275,9 +284,9 @@ def unpack(message, return_dict=False):
             lock_description,
             lock_quantity,
             divisible,
-        ) = data_content[0:15]
+        ) = data_content[0:16]
```
warrenpuffett, line 287: For here and the change below we could use `len - 1 ` to calculate this value instead of needing to update the number every time we add something

counterparty-core/counterpartycore/lib/messages/order.py
```diff
@@ -573,6 +573,10 @@ def parse(db, tx, message):
             min_btc_quantity = 0.001 * config.UNIT  # 0.001 BTC
             if util.enabled("btc_order_minimum_adjustment_1"):
                 min_btc_quantity = 0.00001 * config.UNIT  # 0.00001 BTC
+            if util.enabled("fix_min_btc_quantity", tx["block_index"]):
+                min_btc_quantity = int(
+                    D("0.00001") * D(1e8)
+                )  # else min_btc_quantity==1000.0000000000001
```
warrenpuffett, line 579: 0.00001 * 1e8 is the value in the comment. I think the comment can be removed because it makes me think an extra case was commented out.

counterparty-core/counterpartycore/lib/messages/order.py
```diff
@@ -762,6 +766,9 @@ def match(db, tx, block_index=None):
 
             if block_index >= 313900 or config.TESTNET or config.REGTEST:  # Protocol change.
                 min_btc_quantity = 0.001 * config.UNIT  # 0.001 BTC
+                if util.enabled("fix_min_btc_quantity", block_index):
+                    # we subtract 1 because the <= instead of < like when checking orders
```
warrenpuffett, line 770: Maybe we make a function `new_min_btc_quantity` as this computation is performed in multiple places?

counterparty-core/counterpartycore/lib/transaction.py
```diff
@@ -1299,6 +1308,14 @@ def compose(db, source, transaction_name, api_v1=False, **kwargs):
     )
 
 
+def get_key_name(**construct_args):
```
warrenpuffett, line 1311: Can be written more concisely like so:

```python
def get_key_name(**construct_args):
    if construct_args.get("return_only_data"):
        return "data"
    if construct_args.get("return_psbt"):
        return "psbt"
    return "rawtransaction"
 ```

File: counterparty-core/counterpartycore/test/conftest.py [hunk 5]
```diff
@@ -123,7 +123,7 @@ def shouldCheckForMockProtocolChangesAtBlock(change_name):
 _generate_random_asset = util.generate_random_asset
 
 
-def generate_random_asset():
+def generate_random_asset(subasset_longname=None):
```
warrenpuffett, line 126: Why are we adding this parameter and then not using it?

counterparty-core/counterpartycore/test/regtest/testscenarios.py
```diff
@@ -29,27 +38,88 @@ def compare_strings(string1, string2):
...
+    return item
+
+
+def control_result(item, node, context, block_hash, block_time, tx_hash):
+    for control in item["controls"]:
+        control_url = control["url"].replace("$TX_HASH", tx_hash)
+        for i, address in enumerate(node.addresses):
+            control_url = control_url.replace(f"$ADDRESS_{i+1}", address)
+        result = node.api_call(control_url)
+        try:
```
warrenpuffett, line 63: Maybe more this `try` to just before the `assert`?

counterparty-core/counterpartycore/test/regtest/testscenarios.py
```diff
@@ -113,6 +154,11 @@ def run_scenarios():
 
         for item in SCENARIOS:
             context = run_item(regtest_node_thread.node, item, context)
+
+        # print("Server ready, ctrl-c to stop.")
```
warrenpuffett, line 158: remove?

counterparty-core/counterpartycore/lib/messages/order.py
ouziel-slama: yeah right! Added that to explain why I use `D("0.00001")` instead `D(0.00001)`.

counterparty-core/counterpartycore/test/conftest.py
ouziel-slama: it's a mock function so it should respect the signature event if we don't use the argument.

counterparty-core/counterpartycore/test/regtest/testscenarios.py
ouziel-slama: I used it often! I need to add some arguments to the script `testsenarios.py`

counterparty-core/counterpartycore/lib/messages/order.py
ouziel-slama: hum.. no if fact it's not exactly the same in the two places (there is a -1 here)

warrenpuffett (approved):

PR #2209, opened 2024-09-09
Asset Name Sorting of Balances

droplister:
When sorting balances, I think it would be a better experience to sort in a way that considers asset_longname.


adamkrellenstein (changes requested):
Nice! We just need an entry in the release notes for everything. There's a checklist for new PRs that should be used.

adamkrellenstein (approved):

PR #2247, opened 2024-09-19
Add `inputs_set` support

ouziel-slama:


adamkrellenstein (approved):

adamkrellenstein:
I would say `inputs_set` is a bit better :)

adamkrellenstein:
Is there a max number you're allowed to pass?

ouziel-slama:
> > Is there a max number you're allowed to pass?
> 
> nope! but we make an RPC request by utxo (to retrieve the value) so maybe we should!

Just added a limit a 100. We could adjust if necessary.

PR #2493, opened 2024-10-20
support "ALL" quantity for send

valperez715:
this would be very useful when a proxy address is managing tx for a user address.


adamkrellenstein:
I think this should really just be done in the UI. It's two API calls instead of one, but we keep the API smaller and simpler.

PR #2626, opened 2024-10-29
UTXO Support Fixes and Tweaks

ouziel-slama:
- [x] Spend UTXO when detaching assets
- [x] Attach: use `destination` field to specify the vout to use instead an arbitrary UTXO
- [x] Attach: remove `source` field; use the transaction source instead
- [x] Detach: remove `source` field; detach all UTXO with balances in the inputs
- [x] Detach: make `destination` optional; detach to `utxo_address`.
- [x] Don't use other utxo with balance when composing utxo move

----


counterparty-core/counterpartycore/lib/messages/utxo.py
```diff
@@ -60,19 +87,13 @@ def validate(db, source, destination, asset, quantity, block_index=None):
...
-        if asset_balance < quantity + fee:
-            problems.append("insufficient funds for transfer and fee")
-    else:
-        if asset_balance < quantity:
-            problems.append("insufficient funds for transfer")
-        if source_is_address:
-            xcp_balance = ledger.get_balance(db, source, config.XCP)
-            if xcp_balance < fee:
-                problems.append("insufficient funds for fee")
+    if source_is_address and util.enabled("spend_utxo_to_detach"):
```
wilfreddenton, line 90: Can we refactor this and the next if block to make the logic more clear and avoid the early return?

counterparty-core/counterpartycore/lib/blocks.py
```diff
@@ -107,8 +107,22 @@ def parse_tx(db, tx):
...
+                except struct.error:  # Deterministically raised.
+                    message_type_id = None
+                    message = None
+            else:
+                message_type_id = None
+                message = None
+
             if "utxos_info" in tx and tx["utxos_info"]:
-                utxo.move_assets(db, tx)
+                if not util.enabled("spend_utxo_to_detach"):
```
wilfreddenton, line 121: This if block seems to be the opposite of the if block below and yet they seem to perform the same action?

counterparty-core/counterpartycore/lib/messages/utxo.py
```diff
@@ -176,87 +265,62 @@ def parse(db, tx, message):
...
-            "fee_paid": fee,
-        }
+        fee_paid = pay_fee(db, tx, action, source, recipient)
+
         # update counter
         if action == "attach to utxo":
             gas.increment_counter(db, ID, tx["block_index"])
 
-    ledger.insert_record(db, "sends", bindings, event)
+        if action == "attach to utxo" or not util.enabled("spend_utxo_to_detach"):
```
wilfreddenton, line 275: maybe it's best avoid `or not` operator? We want to `proceed_send` on any action other than `"attach to utxo"` when `spend_utxo_to_detach` is not enabled?

counterparty-core/counterpartycore/lib/blocks.py
ouziel-slama: Yes. Before the protocol change "spend_utxo_to_detach" we always move utxo before parsing the data.
After the protocol change we move utxo before parsing only if the transaction is not an Attach or Detach, else we move utxos after the parsing.

counterparty-core/counterpartycore/lib/messages/utxo.py
ouziel-slama: well seen, done!

counterparty-core/counterpartycore/lib/messages/utxo.py
ouziel-slama: We want to proceed_send on ALL actions when spend_utxo_to_detach is not enabled.
We want to proceed_send only on "attach to utxo"  when spend_utxo_to_detach is enabled.

counterparty-core/counterpartycore/lib/transaction_helper/transaction_inputs.py
```diff
@@ -262,6 +263,31 @@ def construct_coin_selection(
     else:
         use_inputs = unspent
 
+    # forced  utxo always in first position
+    if force_utxo is not None:
+        txid, vout = force_utxo.split(":")
+        included_pos = None
+        for i, v in enumerate(unspent):
```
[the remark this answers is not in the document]
ouziel-slama: I think the comment is misleading. We always want `force_utxo` to be the first item of unspent. I updated the comment

counterparty-core/counterpartycore/lib/blocks.py
wilfreddenton: Ok maybe the comment on line 124 is confusing me then. also I think you can combine the if/elif into an or?

ouziel-slama: yeah right! I re-write the comment and used one if.

counterparty-core/counterpartycore/lib/messages/utxo.py
```diff
@@ -160,9 +243,14 @@ def parse(db, tx, message):
 
     # detach if source is a UTXO
     if util.is_utxo_format(source):
-        source_address, _value = backend.bitcoind.get_utxo_address_and_value(source)
-        if source_address != tx["source"]:
-            problems.append("source does not match the UTXO source")
+        if util.enabled("spend_utxo_to_detach"):
```
adamkrellenstein, line 246: Can you comment this code a bit? It's still pretty opaque to me.

counterparty-core/counterpartycore/lib/messages/utxo.py
```diff
@@ -109,44 +129,107 @@ def compose(db, source, destination, asset, quantity):
...
+    ledger.credit(
+        db,
+        recipient,
+        asset,
+        quantity,
+        tx["tx_index"],
+        action=action,
+        event=tx["tx_hash"],
+    )
+    # we store parameters only if the transaction is valid
```
adamkrellenstein, line 217: where is the validity determined? in `ledger.insert_record()`?

counterparty-core/counterpartycore/lib/messages/utxo.py
adamkrellenstein: let's include a comment for this

counterparty-core/counterpartycore/lib/messages/utxo.py
```diff
@@ -176,87 +264,62 @@ def parse(db, tx, message):
...
+        if action == "attach to utxo" or not util.enabled("spend_utxo_to_detach"):
+            proceed_send(db, tx, action, source, recipient, asset, quantity, event, fee_paid)
+
+        if action == "detach from utxo" and util.enabled("spend_utxo_to_detach"):
+            # we detach all the assets from the source UTXO
+            balances = ledger.get_utxo_balances(db, source)
+            for balance in balances:
+                if balance["quantity"] == 0:
+                    continue
+                proceed_send(
```
adamkrellenstein, line 283: call this something else? `move_asset()`?

counterparty-core/counterpartycore/lib/transaction.py
```diff
@@ -195,7 +198,14 @@ def construct(
     exclude_utxos_with_balances=False,
 ):
     # Extract tx_info
-    (source, destinations, data) = tx_info
+    (address_or_utxo, destinations, data) = tx_info
+
+    force_utxo = None
```
adamkrellenstein, line 203: what does this do?

adamkrellenstein (changes requested):
I left some specific notes but I think in general it'd be great to comment this code very verbosely... it's super tricky!

counterparty-core/counterpartycore/lib/api/compose.py
```diff
@@ -630,8 +630,8 @@ def compose_utxo(
     db,
     source: str,
     destination: str,
-    asset: str,
-    quantity: int,
+    asset: str = None,
+    quantity: int = None,
```
DerpHerpenstein, line 634: The unpack function in utxo.py still unpacks the message with an asset and a quantity, so i could presumably compose a detach manually that contains these fields and they will be unpacked and bubbled up?  is that what is desired here?  

https://github.com/CounterpartyXCP/counterparty-core/blob/6904df8242e3d0cd79b5e07fc842587f3bf1bd99/counterparty-core/counterpartycore/lib/messages/utxo.py#L147

counterparty-core/counterpartycore/lib/messages/utxo.py
```diff
@@ -160,9 +243,14 @@ def parse(db, tx, message):
 
     # detach if source is a UTXO
     if util.is_utxo_format(source):
-        source_address, _value = backend.bitcoind.get_utxo_address_and_value(source)
-        if source_address != tx["source"]:
-            problems.append("source does not match the UTXO source")
+        if util.enabled("spend_utxo_to_detach"):
+            utxos = tx["utxos_info"].split(" ")
+            if len(utxos) < 2 and source not in utxos[:-1]:
```
DerpHerpenstein, line 248: shouldn't you throw the error if the utxo length is less than two OR the source is not in the input utxos? It looks like both conditions should invalidate the tx?

Am I correct in assuming the move_assets for attach/detach is moved until after parsing to ensure the above check clears before debiting/crediting the ledger?

counterparty-core/counterpartycore/lib/messages/utxo.py
ouziel-slama: done!

counterparty-core/counterpartycore/lib/messages/utxo.py
ouziel-slama: I remove this comment from here. Come from a copy/past and it's not relevant here.

counterparty-core/counterpartycore/lib/messages/utxo.py
ouziel-slama: Done

counterparty-core/counterpartycore/lib/transaction.py
ouziel-slama: Added a comment in the code to explain.
The idea is that now `*.compose()` function could return an UTXO as source, if so the construct function will force the presence of this utxo in the transaction inputs.

counterparty-core/counterpartycore/lib/messages/utxo.py
ouziel-slama: right! should be an OR. Fixed, thank you.

Yes right! see https://github.com/CounterpartyXCP/counterparty-core/pull/2626/files#diff-5af753d407909a443a7a31b3406b6eb0866c131536670744cf593ea0fd627ef0R120

counterparty-core/counterpartycore/lib/api/compose.py
ouziel-slama: yeah right!
for now the validate and parse function just ignore them. I hesitated to force them to None but I let like this to avoid another `if util.enabled("spend_utxo_to_detach")`.. but given your comment seems confusing :) I will update this.

(the root problem is that initially attach and detach had the same signature.. not it would be cleaner to have two functions but don't want to change too much code :-/)

ouziel-slama: see https://github.com/CounterpartyXCP/counterparty-core/pull/2626/commits/b9b702782e333de96a4ccc43493c9d2f1efd9405

ouziel-slama:
@adamkrellenstein @wilfreddenton @DerpHerpenstein with all the good ideas described here https://github.com/CounterpartyXCP/counterparty-core/issues/2624#issuecomment-2446892916 there are too many changes to do this properly in `utxo.py`.. so I took the opportunity to split the contract into two `attach.py` and `detach.py`, it's much cleaner and readable. From the activation it will be these two contracts that will be used instead of `utxo.py`. Sorry for this first review before a refactor!

btcopenstamp (commented):
We primarily reviewed `attach.py` and `detach.py`. Our feedback mainly focuses on the following two points for your reference:

1. To ensure detach transactions without non-OP_RETURN outputs can be processed correctly, "sources" should be returned when there's no "non-opreturn" in the `get_utxos_info` method. Consequently, the logic in `attach.py` related to checking for bindable UTXOs will need adjustment.

2. Avoid attaching assets to already spent UTXOs whenever possible—either burn the assets or return them to the relevant address.

counterparty-core/counterpartycore/lib/gettxinfo.py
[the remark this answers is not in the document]
ouziel-slama: good catch thanks! I understand the problem but I'm not sure how to solve it. `utxos_info` contains nothing if there is no destination to be sure that when there is only one element it is a destination and we can use it for an attach.. I will look for a way to allow a detach without non-OP_RETURN output..

counterparty-core/counterpartycore/lib/messages/attach.py
```diff
@@ -0,0 +1,228 @@
...
+        utxos_info = tx["utxos_info"].split(" ") if tx["utxos_info"] else []
+        if len(utxos_info) == 0:
+            problems.append("no UTXO to attach to")
+        else:
+            # last element of utxos_info field is the first non-OPT_RETURN output
+            destination = utxos_info[-1]
+    else:
+        # IMPORTANT: if the vout provided doesn't exist in the transaction or
+        # if is an OP_RETURN output, the attached assets will be unspendable.
+        # We don't check this here because:
```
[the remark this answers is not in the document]
ouziel-slama: yeah good idea! will store also the vout number of an eventual OP_RETURN, so we could also make this check.

counterparty-core/counterpartycore/lib/messages/attach.py
```diff
@@ -0,0 +1,228 @@
...
+
+    # determine destination
+    if destination_vout is None:
+        # if no destination_vout is provided, we use the first non-OPT_RETURN output
+        utxos_info = tx["utxos_info"].split(" ") if tx["utxos_info"] else []
+        if len(utxos_info) == 0:
+            problems.append("no UTXO to attach to")
+        else:
+            # last element of utxos_info field is the first non-OPT_RETURN output
+            destination = utxos_info[-1]
```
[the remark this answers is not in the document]
ouziel-slama: yes right!

counterparty-core/counterpartycore/lib/messages/detach.py
```diff
@@ -0,0 +1,133 @@
...
+        },
+    )
+
+
+def parse(db, tx, message):
+    destination = unpack(message)
+
+    # utxos_info is a space-separated list of UTXOs, last element is the destination,
+    # the rest are the inputs with a balance
+    utxos_info = tx["utxos_info"].split(" ") if tx["utxos_info"] else []
```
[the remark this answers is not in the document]
ouziel-slama: right, see my comment above.

counterparty-core/counterpartycore/lib/messages/detach.py
```diff
@@ -0,0 +1,133 @@
...
+
+def validate(source, destination):
+    problems = []
+
+    # check if source is a UTXO
+    if not util.is_utxo_format(source):
+        problems.append("source must be a UTXO")
+
+    # check if destination is an address
+    try:
```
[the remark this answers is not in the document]
ouziel-slama: hum.. that possible but I would say only if destination is not provided at all else could be a typo in the address.

ouziel-slama:
> We primarily reviewed `attach.py` and `detach.py`. Our feedback mainly focuses on the following two points for your reference:
> 
> 1. To ensure detach transactions without non-OP_RETURN outputs can be processed correctly, "sources" should be returned when there's no "non-opreturn" in the `get_utxos_info` method. Consequently, the logic in `attach.py` related to checking for bindable UTXOs will need adjustment.
> 2. Avoid attaching assets to already spent UTXOs whenever possible—either burn the assets or return them to the relevant address.

Thank you for this review! 

1. see my comment above
2. it's not possible anymore! the assets are always attached to the first non-OP_RETURN output or to the `destination_vout`  output

counterparty-core/counterpartycore/lib/gettxinfo.py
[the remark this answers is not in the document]
ouziel-slama: fixed here https://github.com/CounterpartyXCP/counterparty-core/pull/2626/commits/2a5d59a4ca282f99699a1ce50f1b36b32f4c2383
the new format of  `utxo_infos` field is:
`<comma_separted_inputs_with_balances> <first_non_op_return_output> <outputs_count> <op_return_vout>`

counterparty-core/counterpartycore/lib/messages/detach.py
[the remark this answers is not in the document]
ouziel-slama: done here https://github.com/CounterpartyXCP/counterparty-core/pull/2626/commits/d58103b65d6f0760c0755297f61c2f235e374d01

counterparty-core/counterpartycore/lib/api/compose.py
ouziel-slama: so we have 2 clean contracts now :)

counterparty-core/counterpartycore/lib/messages/move.py
```diff
@@ -0,0 +1,84 @@
...
+    action = "utxo move"
+    msg_index = 0
+    # we move all assets from each source to the destination
+    for source in sources:
+        balances = ledger.get_utxo_balances(db, source)
+        for balance in balances:
+            if balance["quantity"] == 0:
+                continue
+            # debit asset from source
+            ledger.debit(
```
DerpHerpenstein, line 46: In blocks.py you have attach.parse and detach.parse.
When looking in detach, parse goes over all sources and calls detach, and detach debits/credits the ledger.
then in blocks.py, you call move.move_assets which once again debits/credits the ledger.

It looks like the assets are being moved twice? What am I missing?

ouziel-slama: you are absolutely right! see the comment here: https://github.com/CounterpartyXCP/counterparty-core/pull/2626/files#diff-d605bdc22d2945dfe8e33a259ec63f2bda3d3a9a9ea8aea1022ec17541224b3dR141

The assets will not be debited twice, nothing will happen at all because the balances will already be emptied by the detach.
On the other hand, I think it is better to leave the call of `utxo.move()` after the transaction parsing even for the detach because if the detach is invalid (for example with an invalid destination) there is no reason not to do a possible utxo move.

btcopenstamp: What if there’s no NON-OPRETURN in an invalid destination detach transaction? In that case, it seems the assets would remain in spent UTXOs, which we should aim to avoid after this fix. I suggest detaching the assets to the origin input addresses, similar to how it’s handled when no destination is specified in `detach.py`.

ouziel-slama: with the new `utxos_info` field format we don't check the outputs at all for detach. All the inputs are detached to the  provided destination or, if not provided, by default to the address corresponding to the vin.

ouziel-slama: I'm not sure we should fallback to the address corresponding to the vin if the destination is invalid. I feel is cleaner to mark the transaction as invalid and do nothing. Could be a typo and the user will fix in a new tx for example.

btcopenstamp: > I'm not sure we should fallback to the address corresponding to the vin if the destination is invalid. I feel is cleaner to mark the transaction as invalid and do nothing. Could be a typo and the user will fix in a new tx for example.

They can’t perform any action with a new transaction because the asset UTXO has already been spent.

counterparty-core/counterpartycore/lib/messages/move.py
```diff
@@ -0,0 +1,84 @@
...
+def move_assets(db, tx):
+    if "utxos_info" not in tx or not tx["utxos_info"]:
+        return
+
+    sources, destination, _outputs_count, _op_return_output = util.parse_utxos_info(
+        tx["utxos_info"]
+    )
+
+    # do nothing if no source or destination
+    if len(sources) == 0 or not destination:
```
btcopenstamp, line 33: If no valid destination in a move transaction, either burning all the assets or returning them to the address associated with the UTXO to avoid assets remaining in a spent UTXO

` if len(sources) == 0 or not destination:
        return`

counterparty-core/counterpartycore/lib/messages/detach.py
```diff
@@ -0,0 +1,143 @@
...
+    # check if source is a UTXO
+    if not util.is_utxo_format(source):
+        problems.append("source must be a UTXO")
+
+    # check if destination is an address
+    if destination is not None:
+        try:
+            script.validate(destination)
+        except script.AddressError:
+            problems.append("destination must be an address")
```
btcopenstamp, line 23: If the destination is an invalid address, either burning all the assets or returning them to the address associated with the UTXO to avoid assets remaining in a spent UTXO

btcopenstamp (commented):
To avoid leaving assets in spent UTXOs, I suggest either burning all the assets or returning them to the address corresponding to the vin UTXO in the following cases:
1. Move transaction without NON-OPRETURN output.
2. Detach transaction with an invalid destination address or no destination specified.

counterparty-core/counterpartycore/lib/messages/move.py
ouziel-slama: yeah make sense!

counterparty-core/counterpartycore/lib/messages/detach.py
ouziel-slama: done here https://github.com/CounterpartyXCP/counterparty-core/pull/2626/commits/82ecd19af35ffabf43577567b08b7477d6b7f561

counterparty-core/counterpartycore/lib/messages/move.py
ouziel-slama: Why does it bother you so much to have assets attached to an unspendable output? It's actually a kind of burning! I mean if someone goes to the effort of crafting a transaction with a single OP_RETURN it's because they want it to become impossible to move assets and that's the definition of burning.

btcopenstamp: My main considerations are:

1. Assets bound to spent UTXOs are essentially garbage data and unnecessary, as users cannot actually use them, which may also cause some confusion.
2. Third-party platforms (such as OpenStamp) need to determine whether assets are bound to spent UTXOs when calculating users’ valid assets; we don’t want to display unusable assets to users.


Of course, these are just my suggestions for everyone’s discussion and reference.

adamkrellenstein: Yeah, I've thought a lot about this and I do think that preventing burning of UTXO–attached assets is probably a good idea. There are other ways to burn.

counterparty-core/counterpartycore/lib/blocks.py
```diff
@@ -238,13 +252,19 @@ def parse_tx(db, tx):
                 util.CURRENT_TX_HASH = None
                 return False
 
+            # if attach or detach we move assets after parsing
+            if util.enabled("spend_utxo_to_detach") and message_type_id in [attach.ID, detach.ID]:
```
btcopenstamp, line 256: There's likely no need to call "move.move_assets(db, tx)" if it’s a detach transaction.

`if util.enabled("spend_utxo_to_detach") and message_type_id in [attach.ID, detach.ID]:
                move.move_assets(db, tx)`

ouziel-slama: yeah right, no need anymore now that detach can't fail! will fix that.

ouziel-slama:
> It all seems good. @ouziel-slama Appreciate your work.

Thank you so much  for your great feedbacks!!

counterparty-core/counterpartycore/lib/blocks.py
ouziel-slama: done here https://github.com/CounterpartyXCP/counterparty-core/pull/2626/commits/d65072563ff8c56949ba555ad13227656c41bd7c

ouziel-slama:
@adamkrellenstein @DerpHerpenstein @wilfreddenton ok for you to merge ?

adamkrellenstein (approved):

adamkrellenstein:
@ouziel-slama do the release notes need tweaking?

ouziel-slama:
> @ouziel-slama do the release notes need tweaking?

I don't think. Just added this line at the end of the last paragraph:
"A UTXO move with a transaction that contains only a single OP_RETURN output behaves like a `detach`"

PR #2677, opened 2024-11-10
Gas System: tests, fixes and tweaks

ouziel-slama:


counterparty-core/counterpartycore/test/gas_test.py
```diff
@@ -0,0 +1,97 @@
+import tempfile
```
adamkrellenstein, line 1: let's also add a test that makes sure the values are right (prints out the fee at lots of different tx levels so we can plot it easily to verify the behavior)

ouziel-slama: done

adamkrellenstein (approved):

PR #2680, opened 2024-11-10
upgrade rust bitcoin deps

warrenpuffett:


adamkrellenstein:
We need to update the changelog and we need to test a full reparse

PR #2700, opened 2024-11-12
Fixes

ouziel-slama:
- Fix CORS for OPTIONS requests
- Fix rounding error on normalized quantity
- Use `null` instead of '' for `asset_longname` and `asset_parent` fields
- Correctly catch `ValueError` in unpack endpoint
- Correctly catch `InvalidBase58Error` in compose endpoints
- Correctly catch `BitcoindRPCError` in get transaction info endpoint
- Fix typo in dispenser error messages (`has` -> `have`)
- Fix get balances endpoint when using `sort=asset`
- Catch all errors when using unpack endpoint with invalid data
- Restart RSFetcher when it returns None
- Add `sort` parameter for the get holders endpoint (sortable fields: `quantity`, `holding_type`, and `status`)
- Exclude blocks that are not finished being parsed
- Support the `SENTRY_SAMPLE_RATE` environment variable to set the Sentry sample rate
- Clean up blocks without ledger hash before starting the catch up

------------

* [ ] Ensure that the test suite passes


counterparty-core/counterpartycore/lib/messages/dispenser.py
droplister, line 1: There are two mentions of "has" asset in this file, see line 232.

counterparty-core/counterpartycore/lib/messages/dispenser.py
```diff
@@ -232,7 +232,7 @@ def validate(
         problems.append(f"address doesn't has the asset {asset}")
```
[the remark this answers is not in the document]
ouziel-slama: thanks! fixed here https://github.com/CounterpartyXCP/counterparty-core/pull/2700/commits/7bc8f16eb8acd5dfd7331a35587582953e3905e7

adamkrellenstein (approved):

counterparty-core/counterpartycore/lib/api/api_server.py
```diff
@@ -384,19 +384,35 @@ def init_flask_app():
...
+        methods = ["OPTIONS", "GET"]
         # Add routes
-        app.add_url_rule("/v2/", view_func=handle_route, strict_slashes=False)
         app.add_url_rule(
-            "/v2/blueprint", view_func=handle_doc, methods=["GET"], strict_slashes=False
+            "/v2/",
+            view_func=handle_route,
+            methods=methods,
+            strict_slashes=False,
+            provide_automatic_options=False,
```
adamkrellenstein, line 394: what does this arg do?

adamkrellenstein (approved):

counterparty-core/counterpartycore/lib/api/api_server.py
ouziel-slama: by default Flask manage alone OPTIONS request and doesn't pass them to the application.

PR #2809, opened 2024-12-09
Fixes

ouziel-slama:
- Catch up with RPC when ZMQ is late
- Restart RSFetcher when it returns None too many times
- Fix shutdown with Gunicorn

--------

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/wsgi.py
+++ b/counterparty-core/counterpartycore/lib/api/wsgi.py
@@ -54,6 +54,7 @@ class CurrentStateThread(threading.Thread):
         self.stop_event = threading.Event()
 
     def run(self):
+        logger.debug("Starting CurrentStateUpdater thread...")
         try:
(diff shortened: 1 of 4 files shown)


counterparty-core/counterpartycore/lib/api/wsgi.py
```diff
@@ -54,6 +54,7 @@ def __init__(self):
         self.stop_event = threading.Event()
 
     def run(self):
+        logger.debug("Starting CurrentStateUpdater thread...")
```
adamkrellenstein, line 57: This should be renamed -> `NodeStatusChecker` I think

adamkrellenstein (approved):

counterparty-core/counterpartycore/lib/api/wsgi.py
ouziel-slama: done

PR #2873, opened 2024-12-19
Composer V2

ouziel-slama:
- Replacement of `transaction.py` and `transaction_helper/*` with `composer.py`
- No more dependency on Addrindexrs, prioritizing the use of Bitcoin Core to retrieve UTXOs or search for a public key, and then with Electr if it is configured.
- Use of the `bitcoin-utils` library to generate transactions
- API backward compatible with the old composer
- Addition of the following parameters:
    * `change_address`: allows defining the change address
    * `more_outputs`: allows adding an arbitrary number of outputs in the form `<value>:<address>` or `<value>:<script_pubkey>`
    * `use_all_inputs_set`: forces the use of all UTXOs provided with `inputs_set`
    * `sat_per_vbyte`: allows defining transaction fees
    * `max_fee`: defines the maximum fees to be paid
    * `verbose`: includes transaction details, notably `data` and `psbt`
    * `multisig_pubkey`: public key allowing the redemption of multisig data outputs
- The following parameters are deprecated but still work: `fee_per_kb`, `fee_provided`, `dust_return_pubkey`, `return_psbt`, `regular_dust_size`, `multisig_dust_size`, `extended_tx_info`, `old_style_api`, `p2sh_pretx_txid`, `segwit`, `unspent_tx_hash`
- Composed transactions no longer contain the `script_pubkey` (lock script) where the `script_sig` (unlock script) should be.
- With `verbose=true`, the composer also returns a `lock_scripts` field that contains the `script_pubkey` of the UTXOs used by the transaction.
- Composed transactions use a version byte 2 instead of version 1
- Error messages for UTXOs now contain the reason for the error in parentheses: `invalid UTXOs: <utxo(s)> (<reason>)`

-------------


counterparty-core/counterpartycore/lib/backend/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+def list_unspent(source, allow_unconfirmed_inputs):
+    # first try with Bitcoin Core
+    unspent_list = backend.bitcoind.list_unspent(source, allow_unconfirmed_inputs)
+    if len(unspent_list) > 0:
+        return unspent_list
+
+    # then try with Electrs
+    if config.ELECTRS_URL is None:
+        raise exceptions.ComposeError(
+            "No UTXOs found with Bitcoin Core and Electr is not configured, use the `inputs_set` parameter to provide UTXOs"
```
adamkrellenstein, line 13: `Electrs`

counterparty-core/counterpartycore/lib/backend/__init__.py
```diff
@@ -0,0 +1,30 @@
+from counterpartycore.lib import backend, config, exceptions
+
+
+def list_unspent(source, allow_unconfirmed_inputs):
+    # first try with Bitcoin Core
```
adamkrellenstein, line 5: have you tested that this works?

counterparty-core/counterpartycore/lib/p2sh.py
```diff
@@ -0,0 +1,179 @@
+"""
```
adamkrellenstein, line 1: So I think we're going to be removing P2SH encoding soon-ish (once we add support for encoding in witness data). In the meantime, we can re-enable this? Will reduce fees for some users.

counterparty-core/counterpartycore/test/regtest/apidoc/group-compose.md
```diff
@@ -1,21 +1,26 @@
...
-    - The resulting transaction is a `P2SH` encoded message, using the redeem script on the transaction inputs as data carrying mechanism.
-    - Sign the transaction following the `Bitcoinjs-lib on javascript, signing a P2SH redeeming transaction` section
-    - **NOTE**: Don't leave pretxs hanging without transmitting the second transaction as this pollutes the UTXO set and risks making bitcoin harder to run on low spec nodes.
+***Deprecated parameters***
+
+The following parameters are deprecated in the new composer and will no longer be supported in a future version:
+
+- `fee_per_kb`: Use `sat_per_vbyte` instead
+- `fee_provided`: Ue `max_fee` instead
+- `unspent_tx_hash`: Use `inputs_set` instea
```
adamkrellenstein, line 18: typo

adamkrellenstein (approved):
beautiful! super clean!

counterparty-core/counterpartycore/lib/backend/__init__.py
ouziel-slama: In unit test we are testing with a mock of electr and the the fallback is tested
In regtest we are testing for now only with bitcoin core 
and there is No UTXOs found with Bitcoin Core and Electr

counterparty-core/counterpartycore/lib/backend/__init__.py
ouziel-slama: fixed! thanks!

PR #2899, opened 2025-01-04
Fixes

ouziel-slama:
- Search also by `destination` in get sends by address endpoints
- Add `source_address` and `destination_address` in `sends` table

---------


counterparty-core/counterpartycore/lib/api/queries.py
```diff
@@ -302,6 +302,7 @@ def select_rows(
         query = f"{query} OFFSET ?"
         bindings.append(offset)
 
+    print(query, bindings)
```
adamkrellenstein, line 305: extra?

adamkrellenstein (approved):

counterparty-core/counterpartycore/lib/api/queries.py
ouziel-slama: oops! cleaned, thanks!

PR #2991, opened 2025-02-05
Fixes and tweaks

ouziel-slama:
- `parse/gettxinfo.py`, `parser/gettxinfo_legacy.py` and `ledger/` 100% covered
- Fix search pubkey for never spent segwit address
- Refactor how required actions are triggered on upgrade
- Add Get Dispense By Hash endpoint
- Integrate coveralls.io in Github workflow
- Tweak badges in README.md
- Fix  shutdown with Gunicorn
- Same log file for all Gunicorn processes
- Fix double log in access file on API query
- Don't call `getblockcount` from each API thread or process

-------


counterparty-core/counterpartycore/lib/ledger/currentstate.py
```diff
@@ -14,6 +19,34 @@ def get_backend_height():
     return block_count + blocks_behind
 
 
+class BackendHeigt(threading.Thread):
```
adamkrellenstein, line 22: *BackendHeight

adamkrellenstein (approved):
❤️

counterparty-core/counterpartycore/lib/ledger/currentstate.py
ouziel-slama: fixed! thank you!

PR #3058, opened 2025-02-28
Fix blueprint

ouziel-slama:


ouziel-slama:
> Do we also have to change `cursor` to be designated as as string?

hum.. depends the endpoint.. some endpoint have cursor as int and some as string.

ouziel-slama:
> Do we also have to change `cursor` to be designated as as string?

I checked every function in queries.py and put the correct type for cursor.

PR #3127, opened 2025-04-02
Counterparty Client

ouziel-slama:
# `xcp` — a Counterparty command-line client & wallet (Rust)

This PR introduces **`counterparty-client/`**, a standalone Rust CLI that composes, signs and broadcasts Counterparty transactions against a Counterparty v2 API server, backed by a local, encrypted Bitcoin wallet. It ships as two identical binaries — **`xcp`** (short) and **`counterparty-client`** — sharing one implementation.

> ⚠️ **Beta.** New code handling private keys. Test on `signet`/`testnet4`/`regtest` and back up your keys before touching mainnet funds.

## Why

Today, using Counterparty from the command line means talking to the API by hand and signing transactions with a separate tool. `xcp` closes that gap: a single, dependency-free binary that turns the **entire v2 API into CLI commands** and adds a **local signing wallet**, so `compose → sign → broadcast` is one command — with private keys that never leave the machine.

## What it does

- **Full v2 API as commands.** Every v2 API function is exposed as a command under `xcp api <function>` (`api --help` lists them; the endpoint list is cached and refreshable with `--update-cache`). Coloured YAML on a terminal, plain when piped, `--json` for machine-readable output on stdout.
- **Local encrypted wallet.** Per-address independent keys (no wallet-wide seed), each with a selectable address type (`--address-type` `bech32` (default, BIP84) / `p2pkh` (BIP44) / `taproot` (BIP86)). Subcommands: `new_address` (prints a BIP39 phrase once), `import_address` (private key or mnemonic), `list_addresses`, `address_balances`, `export_address` (reveals the WIF), `change_password`, `disconnect`.
- **End-to-end transactions.** `wallet transaction <type>` composes via the API, **independently verifies the composed transaction matches the request**, signs locally, prints a summary, asks for confirmation, then broadcasts — one subcommand per `compose_*` function (`send`, `issuance`, `order`, `dispenser`, `dividend`, …). `sign` and `broadcast` are also available separately.
- **Divisibility-aware amounts.** In `wallet transaction`, `--quantity` is human-readable (e.g. `10` = 10.0 units of a divisible asset, up to 8 decimals); the client looks up each asset's divisibility and converts to satoshis for you. Raw `api compose_*` remains the expert path and expects satoshis.
- **Address labels.** `--address`/`--destination` accept a wallet label instead of a raw address.
- **All networks.** `--mainnet` (default) / `--signet` / `--testnet4` / `--regtest`, each defaulting to the official public HTTPS endpoint (regtest → `localhost`). Defaults are written to a `config.toml` on first run and can be repointed at your own server.
- **Automation-friendly.** A non-interactive password (`XCP_WALLET_PASSWORD`) and `-y`/`--yes` skip the prompts for CI and headless servers — but `--yes` only auto-confirms a transaction the client could **fully** verify (see below).
- **Shell autocompletion** (`xcp completion`) for bash, zsh, fish, …

## Signer

A self-contained `bitcoinsigner` module signs the standard script types — **P2PKH, P2WPKH, P2SH** (incl. P2SH-P2WPKH / P2SH-P2WSH), **P2WSH** and **P2TR** (both BIP86 key-path and single-key script-path spends). Each transaction is built, finalized and extracted through an in-process PSBT (nothing is imported or exported). ECDSA nonces are RFC6979-deterministic; Schnorr signing uses BIP340 aux-rand and the shared `secp256k1` context is randomized for side-channel hardening. Signing is fixed to `SIGHASH_ALL` / `SIGHASH_DEFAULT` (a caller-supplied sighash byte is ignored). Every signature is verified against its own key before it is embedded, and the key/scripts are checked to reconstruct the input's `scriptPubKey` before signing. Keys are used to sign locally and are **never sent to the API server**.

*(One exception to local signing: a Taproot inscription/envelope **reveal** is composed **and signed by the server** and broadcast verbatim — the client never signs it, it only bounds its scope, as described below.)*

## Security model

- **Server-transaction verification:** before signing, the client re-decodes the server-composed transaction (ARC4 de-obfuscation → message-type unpack → CBOR/struct decode) and checks it encodes exactly what the user requested — asset, quantity, destination, sweep flags — and that BTC/change only pays the source or the destination. The composed **message type is bound to the command you ran**, so a `send` cannot be silently verified as a wallet-draining `sweep`, and a plain-BTC send that hides a value-bearing Counterparty payload (in an `OP_RETURN` or bare-multisig/P2PK output) is refused. A mismatch aborts; a type the client cannot independently decode is refused for verifiable commands and otherwise requires an explicit confirmation. `--yes` never auto-confirms a transaction the client could not fully check — an unverifiable type, an asset it could not resolve offline, or a fee it could not bound. An absolute miner-fee cap is enforced on every transaction, and any server-signed reveal is bounded to spending only the commit's first output (a dust amount already reviewed in the commit).
- **Transport:** every public network requires `https://` (enforced before any request and by an `https_only` client, which also blocks an https→http redirect downgrade); the cleartext exemption is limited to a regtest endpoint on a genuinely local host.
- **At rest:** the wallet DB is a single encrypted file. The password is first stretched with **Argon2id** (memory-hard; pinned OWASP parameters over a fresh random salt, regenerated on every write) before deriving the `cocoon` (ChaCha20-Poly1305) key, so `cocoon`'s own fixed PBKDF2 work factor is no longer the ceiling on an offline attack against a stolen file (older-format wallets are read and transparently upgraded on next save). Written atomically (`O_EXCL` temp file + rename, no TOCTOU window) with owner-only permissions on Unix and current-user-only ACLs on Windows; a concurrent-writer guard refuses to clobber a change another `xcp` process made since load (so a freshly generated key is never silently lost). New-wallet passwords must be **≥ 12 characters**, sufficiently varied and not a well-known weak password — passphrase strength is the primary defence, with Argon2id raising the cost of guessing it.
- **Password handling:** cached in the **OS keyring** only after it actually unlocks the wallet; `wallet disconnect` clears it and always works (even with no keyring backend).
- **Secret hygiene:** secret flags accept an `@file` form (and prompt without echo when omitted) so keys/mnemonics stay out of shell history and the process list; in-memory secrets use `zeroize`/`secrecy` (the CLI-supplied key/mnemonic is held in a `Zeroizing` buffer and the mnemonic is redacted from `Debug`). The signed transaction hex is withheld from output until after the broadcast confirmation, so it never leaks through scrollback or logs.
- **Supply chain:** `rustls` (no system OpenSSL — `openssl`/`openssl-sys`/`native-tls` are banned by `cargo-deny`); `syntect` trimmed to pure-Rust deps to drop `cargo-audit`-flagged transitive crates. Release builds keep arithmetic overflow checks enabled, so an unexpected amount/fee overflow panics instead of silently wrapping.

## Testing & CI

- **Hermetic unit tests** for the signer (including independent ECDSA/Taproot signature verification), the independent server-transaction verifier (asset/quantity/destination/BTC-routing/message-type/sweep-flags, with tests named after the attack they close), and the API/compose/broadcast paths (via an in-process HTTP mock), plus quantity/divisibility conversion.
- **New CI jobs:** `Rust Test` runs `fmt --check`, `clippy -D warnings` and the client test suite **with coverage** (`cargo-llvm-cov`, wired into Codecov under a `client` flag with an 88% floor), an **MSRV** job pinning the advertised minimum (Rust 1.88), a macOS/Windows build+test matrix, and a `cargo-audit` + `cargo-deny` dependency-audit job (also run daily on a schedule so a newly-disclosed advisory is caught without a push).
- **`Client Regtest Integration`** runs against a live regtest `counterparty-server`: a library test for divisibility lookup + `compose` reachability, **and** a full **fund → compose → sign → broadcast → accept** end-to-end test (`tests/e2e_regtest.rs`) that drives the compiled `xcp` binary and signs with the client's **own** signer (not Bitcoin Core's wallet), then asserts the destination was credited over the live v2 API.

## Try it

```sh
cd counterparty-client && cargo build --release   # or: cargo install --path .
xcp --version

xcp --testnet4 wallet new_address --label savings
xcp --testnet4 wallet transaction send --address <src> --destination <dst> --asset XCP --quantity 10
xcp --testnet4 api get_asset_info --asset XCP
```

Full docs: https://github.com/CounterpartyXCP/Documentation/pull/263

---

* [x] Update the project [release notes](release-notes/release-notes-v11.2.1.md)
* [x] Update the project documentation, as appropriate, with a corresponding Pull Request in the [Documentation repository](https://github.com/CounterpartyXCP/Documentation/pull/263)


adamkrellenstein:
* [x] We should handle divisible vs. indivisible assets
* [x] More helpful error messages when unable to parse JSON response, e.g.
```
Config: /Users/adam/Library/Application Support/counterparty-client/config.toml
API: http://localhost:4000
Wallet DB: /Users/adam/Library/Application Support/counterparty-client/mainnet/mainnet/
Network: Mainnet
Version: 11.0.1
------------ Wallet send_transaction -------------

Error: Failed to send API request

Caused by:
    0: error sending request for url (http://localhost:4000/v2/addresses/bc1q49lll76029jxw2554wsflqg7zdptpc3fqhwhpp/compose/issuance?verbose=true&quantity=0&asset=A18100000000000000000&multisig_pubkey=03a8f17dfde77df77e32841aa4f7e0450af1f626c859acd0a670d01037ccc13e5f&address=bc1q49lll76029jxw2554wsflqg7zdptpc3fqhwhpp): error trying to connect: tcp connect error: Connection refused (os error 61)
    1: error trying to connect: tcp connect error: Connection refused (os error 61)
    2: tcp connect error: Connection refused (os error 61)
    3: Connection refused (os error 61)
```
```
------------- API sendrawtransaction -------------

Error: Failed to parse API response

Caused by:
    0: error decoding response body: expected value at line 1 column 1
    1: expected value at line 1 column 1
```
* [x] `send_transaction` -> `transaction`
* [x] Add `sign` and `broadcast` operations
* [x] Better error message for insufficient *BTC* balance
```
 > target/release/counterparty-client wallet send_transaction issuance --asset A18100000000000000000 --quantity 0 --address bc1q49lll76029jxw2554wsflqg7zdptpc3fqhwhpp
Config: /Users/adam/Library/Application Support/counterparty-client/config.toml
API: http://localhost:4000
Wallet DB: /Users/adam/Library/Application Support/counterparty-client/mainnet/mainnet/
Network: Mainnet
Version: 11.0.1
------------ Wallet send_transaction -------------

API request failed with status: 400 Bad Request
Error: API error: "Insufficient funds for the target amount: 30000 < 157696"
```

adamkrellenstein (approved):

PR #3244, opened 2026-01-05
Support Sub-1 sat/vByte Fee Rates

adamkrellenstein:
Replace \`int()\` with \`math.ceil()\` when calculating transaction fees in the composer. This properly supports fractional sat/vByte rates like 0.5 or 0.1 that can come from Bitcoin Core's \`estimatesmartfee\`.

Without this fix, \`int(0.5 * 100)\` would work correctly (50), but edge cases like \`int(0.1 * 5)\` would truncate to 0 instead of rounding up to 1.


ouziel-slama (approved):
nice!

PR #3250, opened 2026-01-12
Lots of Performance Optimizations

adamkrellenstein:
### Startup & Shutdown
- **Startup time**: 8+ minutes → 33 seconds (removed AddressEventsCache, store event in DB directly)
- **Shutdown time**: 182 seconds → 21 seconds (`SIGTERM` handler for Kubernetes)

### Memory & Performance
- **Memory usage**: ~5GB → ~3.5GB (N.B. may still be growing unbounded)
- Connection pool monitoring: POOL_STATS every 60s with peak tracking and contention warnings
- Fix: DB config params now passed to API subprocess (was causing connection pool contention)
- Memory profiler: `--memory-profile` flag for RSS and cache size logging (lightweight, no tracemalloc)
- AssetCache: loads all ~246k assets (~70MB) at startup
- UTXOBalancesCache: unbounded dict (~350k entries, ~47MB) - naturally bounded by blockchain UTXO set
- NotSupportedTransactionsCache: O(n) list → O(1) set

### API Performance
- Use `COLLATE NOCASE` instead of `UPPER()` for case-insensitive queries
- Add indexes for `assets_info`, `balances`, `dispensers` tables
- Optimize verbose mode with sets for deduplication

### Configuration
- `--db-connection-pool-size` - pool size per thread (default: 10)
- `--db-max-connections` - max connections across all threads (default: 50)
- `--memory-profile` - enable memory usage logging

### Other
- Block parsing timing instrumentation at debug level
- BuildKit caching for faster Docker builds


ouziel-slama (approved):
wow a lot of nice things here!!

adamkrellenstein:
Still getting some latency spikes with the latest version, even with:
```
--waitress-threads=20
--db-connection-pool-size=30
--db-max-connections=200
```
Screenshot 2026-01-13 at 5 04 28 PM

adamkrellenstein:
Screenshot 2026-01-13 at 11 54 44 PM

Okay, I think I fixed this, but I'm going to let it run for a while to confirm.

PR #3266, opened 2026-01-28
Allow more MIME types for ordinal inscriptions

antron3000:
Allow MIME types with parameters for ordinal inscriptions. 

e.g. **audio/ogg;codecs=opus**
not only **audio/ogg;**

Fix validation to support MIME types with codec paramters when creating ordinal inscriptions with counterparty tokens via API.

Previously, MIME types with parameters would fail validation because 
the entire string (including parameters) was checked against Python's 
mimetypes library, which only recognizes base MIME types.

Now only the base MIME type is checked against Python's mimetypes library instead of the entire string.

Fixes issue where audio/ogg;codecs=opus was rejected as invalid.


metaver5o:
**Issue: Opus Audio Inscriptions Not Properly Recognized**

**Problem:** When minting Opus audio files (.opus) as Counterparty inscriptions, the API accepts the transaction but the resulting inscriptions are not properly recognized or playable in ordinal viewers.

**Steps to Reproduce:**

1. Prepare an Opus audio file:
```bash
HEX_DATA=$(xxd -p file.opus | tr -d '\n')
```

2. Mint using the Counterparty API:
```bash
curl -X POST "https://api.counterparty.io:4000/v2/addresses/$WALLET_ADDRESS/compose/issuance" \
-H "Accept: application/json" \
-H "Content-Type: application/x-www-form-urlencoded" \
--data-urlencode "asset=OPUSAUDIO" \
--data-urlencode "quantity=1" \
--data-urlencode "divisible=false" \
--data-urlencode "description=$HEX_DATA" \
--data-urlencode "encoding=taproot" \
--data-urlencode "inscription=true" \
--data-urlencode "mime_type=audio/opus" \
--data-urlencode "fee_rate=1"
```

**Current Behavior:**
- API successfully processes the request without MIME type validation errors
- Transaction is created with OP_RETURN and INSCRIPTION tags
- Ordinal viewers display the inscription but cannot play the audio
- Example: https://ordinals.com/inscription/48a853970082b9234927782698a9291edd261c25591eef09479364a9e4a12f1ai0

**Expected Behavior:**
- Opus audio inscriptions should be properly recognized and playable
- MIME type `audio/opus` should be fully supported

**Root Cause:**
The `audio/opus` MIME type appears to be missing from the supported MIME types list in `helpers.py`, causing ordinal viewers to not recognize the file format properly even though the API accepts it.

**Request:**
Add support for `audio/opus` MIME type in `helpers.py` to ensure Opus audio inscriptions are properly recognized and playable in ordinal viewers.

**Additional Context:**
- Opus is an open, royalty-free audio codec widely used for web audio
- Common container formats: Ogg (.opus), WebM (audio only)
- Related MIME types that may also need support: `audio/ogg` (for Ogg container with Opus)

TatianaMoroz:
@adamkrellenstein ^

adamkrellenstein:
Thanks for identifying the issue and submitting this PR @antron3000 Unfortunately this is a protocol change, so it'll take some time to release (and the code to be refactored.

ouziel-slama:
cherry-picked and completed here https://github.com/CounterpartyXCP/counterparty-core/pull/3287
thank you @antron3000 !

Issue #2749, opened 2024-11-23
DEX Market Makers


DerpHerpenstein:
I had a thought recently and have been trying to think it through figured I would drop it here for discussion. Proper AMM's will require quite a bit of new logic, testing and new database tables.

There may be a way to get the vast majority of the functionality into the current dex without a lot of changes by making a discreet version. It would require a new api endpoint, a new column in the dex, and a few lines of code to create a new order on the opposite side when an order of this type is filled. The disadvantage is there will be alot of orders created in the dex... (if this gets popular there could be tens of thousands of filled dex orders per day, so that side of things needs to be considered?  It would be a good thing, but the db would certainly increase in size quickly ). If this were to ever be implemented, it would require an xcp fee large enough to discourage people opening up tons of LPs for tens of dollars in assets. (or perhaps when X pools are open on a pair, new pools must have a value greater than the smallest X/Y pools? just an idea)

Anyway, the idea is pretty simple, a user would compose a transaction to create a "DEX Market Maker".  They would specify the asset pair, the minimum price, the maximum price, spread, and the type (linear or logarithmic spacing).  In this example lets say the user wants to list 250 MINTS in the range 1 XCP to 10 XCP per mint with a spread of 1% and a linear order type.  This order is creating a group of DEX orders selling 25 MINTS each.  The table below shows the values

|Order #|Price in XCP-linear|Price in XCP-log|
|-------|--------------------|-----------------|
|1      |1.00                |1.00             |
|2      |1.38                |1.10             |
|3      |1.75                |1.20             |
|4      |2.13                |1.30             |
|5      |2.50                |1.50             |
|6      |2.88                |1.60             |
|7      |3.25                |1.80             |
|8      |3.63                |2.00             |
|9      |4.00                |2.20             |
|10     |4.38                |2.40             |
|11     |4.75                |2.70             |
|12     |5.13                |3.00             |
|13     |5.50                |3.30             |
|14     |5.88                |3.60             |
|15     |6.25                |3.90             |
|16     |6.63                |4.30             |
|17     |7.00                |4.70             |
|18     |7.38                |5.10             |
|19     |7.75                |5.60             |
|20     |8.13                |6.20             |
|21     |8.50                |6.80             |
|22     |8.88                |7.50             |
|23     |9.25                |8.20             |
|24     |9.63                |9.10             |
|25     |10.00               |10.00            |

[...]

Issue #2339, opened 2024-10-12
Fair Mint AMM


DerpHerpenstein:
Had an idea and just wanted to drop it here.  
In the future if AMM's become a part of cp, having a flag in the XCP fair mint that allows for all the XCP to go directly into an AMM as collateral with a fixed amount of the asset would be a great feature.

pinnate-modo:
This is a great idea!

adamkrellenstein:
@droplister

Issue #2122, opened 2024-07-31
AMMs / LPs


adamkrellenstein:
consider making optional 'governance' component where payouts to contributors are determined by vote

adamkrellenstein:
Dupe with https://github.com/CounterpartyXCP/counterparty-core/issues/1892

Issue #1892, opened 2024-06-09
Automated Market Making (?)


adamkrellenstein:
consider making optional 'governance' component where payouts to contributors are determined by vote

DerpHerpenstein:
I really like the idea of adding automated market makers to counterparty.  It would look something like this:

**Basic Liquidity Pool (LP) Implementation**
A simple LP design utilizing the time tested function x*y=k (see uniswap v2 whitepaper) could be implemented into counterparty. Any asset pair could be created, but it would work smoothest with divisible assets.

**Add Liquidity:** User adds assets to the pool and LP tokens proportional to their stake are sent to them
**Remove Liquidity:** User sends liquidity tokens to be destroyed in exchange for that proportion of the liquidity pool assets being send back to the user
**Swap:** Exchange one asset for the other asset, allows for transfer to a wallet or to a tx output (see below)

Here are some quick resources if your not familiar with the basic idea.

https://betterprogramming.pub/uniswap-v2-in-depth-98075c826254

https://medium.com/@icmoore/analyze-uniswap-in-python-using-uniswappy-804acb8f2e5c


The drawback to this type of system is that it incentivizes MEV sandwiching.  When a user creates a swap they identify the number of tokens they are trading and the minimum they must get in return. Since users don't want to waste money on transaction fees for a failed swap, they add a slippage percentage to ensure their tx goes through.  This results in miners looking for a tx where the slippage is larger than the cost of two transactions, and if it is, they create two transactions to extract the leftover value.  

Ideally the implementation here would be created in such a way that does not incentivize this behavior. A possible solution would be: if multiple users swap token a for token b, and others swap b for token a, the calculation xy >k will be performed using all inputs in a block, rather than per tx sequentially. If the swap fails because xy<=k the transactions with the exchange rates furthest will be removed one at a time until x*y >k is satisfied.  This functionality would ensure that any MEV is only performed at the block level and not the transaction level.

adamkrellenstein:
See https://github.com/CounterpartyXCP/counterparty-core/pull/3278/changes

PR #3278, opened 2026-03-17
AMM Liquidity Pools with Order Book Matching

droplister:
## Overview

This PR adds **constant-product AMM liquidity pools** to the Counterparty DEX, with pools integrated directly into the existing order matching engine.

Orders can fill from:

- the resting order book,
- an AMM pool,
- or both,

with execution routed deterministically by **best price first** while respecting the order’s limit price.

This PR also introduces two new message types:

- `pooldeposit`
- `poolwithdraw`

References: #1892, #2122, #2339, #2749

---

## Design Summary

This is a **Uniswap v2-style constant-product AMM** adapted to Counterparty’s existing DEX model.

Key choices:

- **Constant-product pools** with DB-tracked reserves
- **LP shares** represented as standard transferable numeric assets
- **No concentrated liquidity**
- **No protocol fee extraction**
- **Pool + order book interleave** in a single matching flow
- **Two fee tiers**:
  - **0.50%** for XCP pairs
  - **1.00%** for non-XCP pairs

Pools are meant to complement the existing book, not replace it.

---

## Matching Model

When a new order arrives, matching is interleaved between the book and the pool according to price.

### Routing principle

For each step of execution:

- if the pool offers a better effective price than the next resting book order, fill from the pool up to that price boundary
- otherwise fill from the book
- after eligible resting book liquidity is exhausted, any remaining quantity may continue filling from the pool, subject to the order’s limit price

This gives:

- deterministic behavior
- price-priority across both liquidity sources
- no pool fill at a worse price than the order allows

### High-level flow

```text
New Order
  │
  ├─ Pool exists for pair? ── No ──→ Book match only
  │   Yes
  │
  ├─ Compare next book price vs current pool price
  │     Pool better? ─ Yes → fill from pool up to book-price boundary
  │                   No  → fill from book
  │
  ├─ Repeat until no eligible resting book liquidity remains
  │
  └─ Fill remainder from pool if still within order's price limit
```

This allows a single order to be filled by **book only**, **pool only**, or a **blend of both**.

---

### First deposit (pool creation)

The first liquidity provider sets the initial reserve ratio and therefore the opening pool price.

```text
Depositor supplies qty_a + qty_b
  ├─ Debit qty_a and qty_b
  ├─ Create LP token (numeric asset, issuer = UNSPENDABLE)
  ├─ Mint LP = isqrt(qty_a * qty_b)
  └─ Create pool record with initial reserves
```

### Subsequent deposit

Subsequent deposits are **ratio-preserving**.

The submitted quantities are treated as **maximums**, not mandatory exact debits. The protocol computes the largest proportional deposit that fits within both submitted amounts, debits only those proportional amounts, and leaves any excess in the depositor’s balance.

```text
Depositor supplies max qty_a + max qty_b
  ├─ Compute largest proportional deposit at current reserve ratio
  ├─ Debit only actual_a and actual_b
  ├─ Mint LP proportional to that contribution
  ├─ Credit LP tokens to depositor
  └─ Add only actual_a and actual_b to reserves
```

This means:

- deposits do **not** move price
- off-ratio excess is **not donated** to existing LPs
- price correction must happen through swaps, not through liquidity addition

This is intentional: **swaps move price and pay fees; deposits add liquidity and preserve ratio**.

### Withdrawal

```text
LP holder redeems N LP tokens
  ├─ Debit N LP tokens
  ├─ Destroy N LP tokens
  ├─ Credit proportional reserves:
  │     asset_a = N * reserve_a / total_supply
  │     asset_b = N * reserve_b / total_supply
  └─ Reduce pool reserves
```

LP token destruction reduces total supply, so any externally destroyed LP tokens increase the claim of remaining LP holders on the reserves.

---

## Fee Schedule

| Pair type | Fee | Rationale |
| --- | --- | --- |
| XCP pair | 0.50% (50 bps) | Lower fee encourages XCP-denominated liquidity |
| Non-XCP pair | 1.00% (100 bps) | Higher fee compensates LPs in thinner markets |

Fees are retained in pool reserves. No fee is extracted by the protocol.

This means fees increase pool value for LPs over time by growing reserves and, in normal operation, increasing `k`.

### Example

Swapping `0.1 XCP` through a `100 XCP / 100 POOLTEST` pool:

```text
Input:           10,000,000 sat (0.1 XCP)
Fee (0.50%):         50,000 sat
Effective input:  9,950,000 sat
Output:           9,049,567 sat POOLTEST

Pool after:      110,000,000 XCP / 90,950,433 POOLTEST
k before:        10,000,000,000,000,000
k after:         10,004,547,630,000,000
```

The important behavior is:

- fee is included in execution
- fee remains in reserves
- larger pools reduce slippage for the same trade size

---

### New message types

- **`pooldeposit`**
  - first deposit creates the pool and LP token
  - later deposits mint LP tokens proportionally at the current reserve ratio

- **`poolwithdraw`**
  - destroys LP tokens and returns proportional reserves
  - pool record persists even if reserves go to zero

### Order matching changes

- `match()` now supports deterministic **book/pool interleave**
- pool fills never violate the order’s limit price
- orders can finish as book-only, pool-only, or blended fills
- filled/open accounting now correctly handles cases where `get_remaining <= 0` even if `give_remaining > 0`

### LP token model

- LP tokens are ordinary transferable numeric assets
- `issuer = UNSPENDABLE`
- LP tokens cannot themselves be pooled

### Fees

- deposit/withdraw compose endpoints use the same sigmoid-style fee pattern as attach/detach
- swap fees are pool-native trading fees retained in reserves

### New tables

- `pools`
- `pool_deposits`
- `pool_withdrawals`
- `pool_matches`

---

## Important Invariants

- Pool reserves are **tracked explicitly in the DB**, not inferred from balances
- First deposit sets the opening price
- Subsequent deposits are **ratio-preserving**
- Off-ratio deposit excess is **left with the depositor**
- Deposits too small to mint at least 1 LP token are rejected before debits
- Optional slippage protection via minimum LP mint amount prevents stale quote surprises
- Book/pool interleave is deterministic and respects price priority
- Pool fills never violate the incoming order’s limit price
- Filled/open order accounting is consistent with existing order-book logic
- Exceptions propagate through the outer DB transaction for full rollback safety

---

## Quote Semantics

This PR adds quote endpoints for swap, deposit, and withdrawal.

```text
GET /v2/pools/{give_asset}/{get_asset}/quote/swap?give_quantity=N
GET /v2/pools/{asset1}/{asset2}/quote/deposit?quantity_a=N
GET /v2/pools/{asset1}/{asset2}/quote/withdraw?quantity=N
```

These quotes are **current-state estimates** based on the pool and book state at query time.

### Swap quote behavior

Swap quotes explicitly expose whether the result comes from:

- pool liquidity,
- book liquidity,
- or a blend of both

using fields such as:

- `pool_output`
- `book_output`
- `book_orders_matched`
- `fee_bps`
- `fee_amount`
- `effective_price`
- `price_impact`

Fee is included in the quote result.

### Deposit quote behavior

Deposit quotes return the proportional counterpart amount required at current reserves.

That quote is useful for wallet-side UX and avoids accidental off-ratio submissions, but final execution still depends on current state at confirmation time.

### Withdraw quote behavior

Withdraw quotes reflect the current pool reserve state and LP supply state.

---

## Why deposits do not move price

A deliberate design choice here is that **deposits add liquidity but do not perform price discovery**.

If off-ratio deposits were allowed to move price directly, traders could bypass swap fees by “correcting” the pool through deposits rather than through trades.

By keeping deposits ratio-preserving:

- **price discovery happens through swaps**
- **swaps pay fees**
- **LPs earn from re-pricing activity**
- **liquidity provision remains neutral**

That separation produces cleaner incentives and better matches AMM economics.

---

## API Additions

```text
# Quotes
GET /v2/pools/{give_asset}/{get_asset}/quote/swap?give_quantity=N
GET /v2/pools/{asset1}/{asset2}/quote/deposit?quantity_a=N
GET /v2/pools/{asset1}/{asset2}/quote/withdraw?quantity=N

# Compose
GET /v2/addresses/{address}/compose/pooldeposit
GET /v2/addresses/{address}/compose/poolwithdraw

# Pool data
GET /v2/pools
GET /v2/pools/{asset1}/{asset2}
GET /v2/pools/{asset1}/{asset2}/deposits
GET /v2/pools/{asset1}/{asset2}/withdrawals
GET /v2/pools/{asset1}/{asset2}/matches
GET /v2/pools/{asset1}/{asset2}/price_history

# Address-scoped
GET /v2/addresses/{address}/pools
GET /v2/addresses/{address}/pool_deposits
GET /v2/addresses/{address}/pool_withdrawals

# Order-scoped / global
GET /v2/orders/{order_hash}/pool_matches
GET /v2/pool_matches
```

---

## Tests

| Type | Count | Covers |
| --- | --- | --- |
| `pooldeposit_test` | 19 | validation, compose/unpack, parse, proportional deposits, edge cases |
| `poolwithdraw_test` | 20 | validation, compose/parse, partial withdrawal, LP destruction |
| `pool_match_test` | 8 | pool fills, price limits, fee tiers, BTC exclusion, filled/open accounting |
| `queries_test` | 8 | pool queries and pagination behavior |
| `scenario_26_pools` | 11 steps, 20+ controls, 5 error cases | end-to-end lifecycle with event assertions |

Additional local hardening covered:

- extreme-value math
- matching at scale
- adversarial / property-style edge cases

---

## Notes / Non-Goals

This PR does **not** introduce:

- concentrated liquidity
- custom fee tiers per pool
- protocol fee extraction
- oracle-based pricing safeguards
- special protections against tiny-pool mispricing beyond normal market arbitrage

This PR is intentionally a **simple, deterministic v1 AMM** that fits naturally into Counterparty’s existing DEX architecture.


Ouziel:
nice!!

droplister:
The main awkwardness to this, I think, is the pooled reserves and understanding it in holders/escrow views. I attribute it to UNSPENDABLE and use first pooldeposit tx_hash. Because reserves aren't attributable directly to an address, addresses have lp tokens with claims on a share of reserves. And you could technically deposit and sell your lp tokens, so having made a deposit doesn't make sense to show your address/tx_hash as having an escrowed value. I think that's okay.

The other thing was around quote endpoints and queries and where should the code for that live and what should route names be. It's a new pattern, but really useful to have for UIs and LPs.

Open to design changes or a totally different design that achieves AMMs. This way feels mostly Counterparty-esque.

droplister:
Indivisible assets can get into a weird state where they are illiquid, but not bricked, you can deposit or withdraw but they don't match because of protections against reserve assets being zero.

And you can possibly end up with an amount of lp tokens which can redeem nothing because you like sell or send half of your lp tokens away and you need X > Y tokens to withdraw reserve. But this is like having a dust output and not a design issue really.

Ouziel:
LGTM! elegant implementation!

droplister:
Replaced by https://github.com/CounterpartyXCP/counterparty-core/pull/3281 for better CI.

PR #3279, opened 2026-03-24
Fairmint Option to Trustlessly Fund Liquidity Pool

droplister:
## Overview

This PR adds an optional **automatic pool funding mechanism** to the fairminter, extending the existing fairminter message (type 90) with two new parameters. No new message types are introduced.

When configured, a fairminter can reserve a portion of its token supply for an AMM pool. At soft cap resolution, the reserved tokens are paired with all collected XCP and deposited into a new constant-product pool -- trustlessly, atomically, with no issuer intervention required.

This is an **XCP-native feature**. Fairminters are paid in XCP, and the collected XCP becomes the quote-side reserve of the pool. BTC-based minting cannot provide this because BTC payments are not tracked as protocol-level balances. The soft cap refund mechanism (which returns XCP if participation is insufficient) makes this a low-risk coordination game rather than a blind commitment.

Builds on: #3278 (AMM Liquidity Pools)
References: #2339

---

## Motivation

Every token launch faces the cold-start problem: after minting, there is no market. Someone has to go first -- seed a pool, place the first order, take the risk of being the initial market maker. Usually nobody does, and the token sits inert.

A creator could promise to manually create an LP after a fairmint. But promises are not enforceable. The creator could keep the XCP, never make the pool, or make it and then withdraw liquidity later.

This feature makes pool creation **protocol-enforced**:

- Pool tokens are escrowed at the unspendable address from block one -- the issuer never touches them
- XCP payments are escrowed during the soft cap period -- the issuer never touches them
- Pool creation is automatic and atomic at soft cap resolution
- lock_pool_liquidity=true sends LP tokens to the unspendable address -- nobody can ever withdraw the liquidity
- If the soft cap is not met, everything refunds -- minters lose only a BTC miner fee (~$0.04)

The result is **collective liquidity creation with downside protection**: the community's act of minting IS the act of creating a market, with a protocol guarantee that the pool will exist and cannot be rugged.

---

## New Parameters

Two optional fields added to the fairminter message (CBOR serialization, backwards-compatible):

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| pool_quantity | int | 0 | Tokens reserved for the pool token side. All collected XCP becomes the pool XCP side. |
| lock_pool_liquidity | bool | false | If true, LP tokens are sent to the unspendable address (permanent, irrevocable liquidity). |

When pool_quantity = 0 (default), fairminter behavior is unchanged.

---

### Setup

Creator configures a fairminter with pool_quantity > 0. The pool tokens are minted and escrowed at the unspendable address immediately. The mintable supply is hard_cap - pool_quantity - premint_quantity.

### Minting

Community members mint tokens by paying XCP. During the soft cap period, both XCP payments and minted tokens are escrowed. Every minter pays the same price.

### Resolution (soft cap met)

At the soft cap deadline block:
1. Escrowed tokens are credited to minters
2. Escrowed XCP + pool tokens are deposited into a new AMM pool
3. LP tokens are credited to the unspendable address (if locked) or the issuer
4. The pool opens at total_XCP / pool_quantity -- always above mint price when pool_quantity < mintable

### Resolution (soft cap not met)

All XCP is refunded to minters. All tokens (including pool tokens) are destroyed. The fairminter closes. Cost to each minter: one BTC miner fee.

---

## Pool Economics

The pool opens at a **premium** above mint price. The premium is determined by the ratio of mintable tokens to pool tokens:

| Pool % | Mintable % | Premium | Meaning |
|--------|-----------|---------|---------|
| 10% | 90% | 9.0x | High premium, thin pool |
| 20% | 80% | 4.0x | Strong premium |
| 31% | 69% | 2.2x | XCP-69 style |
| 40% | 60% | 1.5x | Moderate premium |
| 50% | 50% | 1.0x | Pool at mint price |

The premium creates a buffer against early sell pressure. With a 20% pool allocation, approximately 25% of minters can sell before the price drops to the original mint price.

If literally every minter dumps all tokens into the pool, average sellers recover approximately (100 - pool_pct)% of their XCP. The pool survives with permanent liquidity.

---

## Validation Rules

- pool_quantity > 0 requires price > 0 (cannot fund pool from free fairminters)
- pool_quantity > 0 requires soft_cap > 0 (escrow mechanism required)
- pool_quantity > 0 requires hard_cap > 0
- pool_quantity > 0 is incompatible with burn_payment = true (XCP goes to pool, not burn)
- premint_quantity + pool_quantity + soft_cap <= hard_cap
- Pool cannot already exist for the TOKEN/XCP pair
- Pool deposits are blocked for any asset with an active fairminter that has pool_quantity > 0
- Pool deposit gas fee is charged upfront to the fairminter creator
- Protocol-gated behind fairmint_pool activation flag

---

## Compatibility

- **No new message type.** Two optional fields appended to the existing fairminter CBOR payload.
- **Backwards compatible.** Old messages without the new fields unpack with defaults (0, False) -- pool mode disabled.
- **Works with all existing fairminter parameters:** premint, commission, max_per_address, max_per_tx, start/end blocks, lock_quantity, lock_description, divisibility.
- **Existing fairminter unit tests updated** for the new CBOR payload size. All pass.
- **Consolidated tables auto-discover** the new columns via PRAGMA -- no additional API migration needed.

---

## Edge Cases and Protections

- **No issuance during fair minting.** The issuer cannot inflate supply while the fairminter is active (fair_minting = True blocks issuance).
- **Pool-already-exists fallback.** If a pool somehow exists at resolution time despite validation, assets are credited to the issuer rather than lost.
- **Debit/credit accounting verified.** Escrow at UNSPENDABLE always balances: total debited = total credited + pool reserves. The built-in asset conservation checker validates this every block.
- **No overflow risk.** All arithmetic stays within MAX_INT bounds due to validation constraints.
- **Uneconomical configurations are allowed but visible.** A creator can set pool_quantity > mintable (pool opens below mint price) or soft_cap = 1 (minimal participation required). These are transparent in the fairminter params -- minters can evaluate before participating. The protocol does not prevent bad economics, only unsafe mechanics.

---

## Example: XCP-69 Configuration

| Parameter | Value |
|-----------|-------|
| hard_cap | 100,000,000 tokens |
| pool_quantity | 31,000,000 tokens (31%) |
| soft_cap | 69,000,000 tokens (= mintable, all-or-nothing) |
| max_per_address | 1,000,000 tokens (1%) |
| price | 1 sat per unit (1 XCP per 10,000 tokens) |
| lock_pool_liquidity | true |
| lock_quantity | true |

- **Minimum 69 participants** (1% cap)
- **Raises 6,900 XCP** (~$13,800 at $2/XCP)
- **Pool opens at 2.23x** mint price
- **100% of XCP** goes to permanent, locked liquidity
- **If soft cap fails**, everyone gets XCP back minus ~$0.04

---

## Files Changed

| File | Change |
|------|--------|
| fairminter.py | New params in validate/compose/unpack/parse. Pool token escrow. Gas fee for pool deposit. Pool creation at soft cap resolution. |
| pooldeposit.py | create_pool_from_fairminter() -- creates pool from block-level operation. Validation blocks pool deposits during active fairmint-pool. |
| compose.py | API endpoint exposes pool_quantity and lock_pool_liquidity. |
| protocol_changes.json | fairmint_pool activation flag. |
| 0010.fairminter_pool.sql | Adds columns to fairminters table. |
| fairminter_test.py | Updated compose/unpack expected bytes for new CBOR fields. |
| scenario_27_fairminter_pool.py | Regtest scenario: soft cap failure + soft cap success with pool creation. |
| scenarios_test.py | Registers scenario 27. |

---

## Tests

| Type | Status |
|------|--------|
| Existing fairminter unit tests | All pass (updated for new CBOR payload) |
| Regtest scenario (soft cap failure) | Pass -- tokens destroyed, XCP refunded |
| Regtest scenario (soft cap success) | Pass -- pool created, LP tokens locked |
| ruff lint | Pass |
| ruff format | Pass |
| Accounting simulation (441 configs) | 0 escrow leaks |
| Validation sweep (1,440 combos) | All valid/invalid as expected |


droplister:
Recreating with correct base branch

PR #3280, opened 2026-03-29
Codex/pr3278 pool fixes

adamkrellenstein:
This follow-up PR fixes three correctness issues in feature/amm-pools.

  First, it preserves all pool_matches rows during state DB rebuilds. pool_matches is
  append-only event history, not mutable state, so consolidating it by tx_hash,
  order_tx_hash could drop legitimate swaps when a single order hit the pool multiple
  times during book/pool interleave.

  Second, it enforces min_lp_quantity consistently across all deposit paths. The
  original PR only applied slippage protection to subsequent proportional deposits; this
  patch extends the same guard to first deposits and empty-pool re-funds, so callers get
  the same guarantee regardless of pool state.

  Third, it makes pooldeposit.validate() ratio-aware for existing pools. Subsequent
  deposits are designed to treat submitted quantities as maxima and debit only the
  proportional amounts that fit the current reserve ratio. Validation now checks
  balances and XCP fee requirements against those actual debits instead of the submitted
  maxima, which avoids rejecting valid off-ratio deposits.

  In short, this PR aligns validation with execution semantics, makes slippage
  protection uniform, and ensures pool swap history is not lost by the state DB layer.


adamkrellenstein:
@droplister

droplister:
I cherry picked your fixes in #3281 

Work will be continue and be finalized there.

PR #3273, opened 2026-02-17
Add cancel-all mode to cancel message (type 70)

droplister:
## Summary

- Adds a "cancel all" mode to the existing `cancel` message (type ID 70) that cancels all open orders and bets for the source address in a single transaction
- Triggered by omitting the `offer_hash` parameter when composing a cancel message
- Wire format: `[type_id: 1 byte][flag: 1 byte (0x01)]` = 2 bytes total
- Gated behind the `cancel_all_offers` protocol change
- Charges an anti-spam fee of 0.0002 XCP per cancelled offer (same rate as dividend fees)

### Files Changed

| File | Change |
|------|--------|
| `protocol_changes.json` | Add `cancel_all_offers` protocol change entry |
| `lib/ledger/markets.py` | Add `get_open_orders_by_source()` query |
| `lib/ledger/other.py` | Add `get_open_bets_by_source()` query |
| `lib/messages/cancel.py` | Core logic: validate, compose, unpack, parse for cancel-all |
| `lib/api/compose.py` | Make `offer_hash` optional in `compose_cancel()` |
| `lib/api/apiwatcher.py` | Add `CANCEL_ALL` event to address fields |
| `test/units/messages/cancel_test.py` | Unit tests for compose, unpack, and parse paths |
| `release-notes-v11.0.5.md` | Document the new feature |

### Design Decisions

- **Reuses message type ID 70** with a 1-byte flag (`0x01`) rather than creating a new message type
- **Anti-spam fee** (0.0002 XCP per offer, matching dividend rate) prevents DoS via unbounded parse-time work
- **Pre-protocol-change guard** in `parse()` normalizes `0x01` messages back to `"invalid: could not unpack"` before the activation block, preserving identical behavior with the original code
- **Deterministic iteration** via `ORDER BY tx_index, tx_hash` on all queries
- **Fee validation** follows the dividend pattern: check balance in `validate_cancel_all()`, debit in `parse()` only after validation passes

### Edge Cases

- No open offers → `ComposeError` at compose time; `"invalid: no open offers for this address"` at parse time
- Insufficient XCP for fee → `ComposeError` at compose time; `"invalid: insufficient funds"` at parse time
- Protocol not yet enabled → `ComposeError` at compose time; treated as invalid unpack at parse time (backward compatible)
- BTC orders → `cancel_order()` handles correctly (credits non-BTC assets only)

## PR Checklist

- [x] Double-check the spelling and grammar of all strings, code comments, etc.
[...]

PR #3282, opened 2026-04-10
Add cancel-all mode to cancel message (type 70)

droplister:
## Overview

Extends the existing cancel message (type ID 70) with a **cancel-all mode** that cancels all open orders and bets for the source address in a single transaction.

Supersedes #3273 with a cleaner design: idiomatic extension of the existing cancel functions, sigmoid gas fee system instead of flat per-offer fee, and comprehensive test coverage.

---

## Design Summary

- **Reuses message type ID 70** with a 1-byte flag (`0x01`) — same approach as detach.py's placeholder byte
- **Sigmoid gas fee** via existing `fee_parameters` system (ID 70)
- **Gas counter increments by offer count** (not by 1), so large cancel-all operations proportionally increase the fee for subsequent uses
- **Gated behind `cancel_all_offers` protocol change**
- **Before activation**, the 1-byte message is treated as `"invalid: could not unpack"` — preserving identical behavior with the original code

---

## Wire Format

```text
Single cancel: [type_id: 70][32 bytes: offer_hash]
Cancel all:    [type_id: 70][1 byte: 0x01 flag]
```

The 1-byte flag follows the detach.py precedent (`b"0"` placeholder when no meaningful payload exists).

---

### Compose

`offer_hash` is now optional in `compose_cancel`. When omitted, the cancel-all flag is packed:

```text
GET /v2/addresses/{address}/compose/cancel                    → cancel all
GET /v2/addresses/{address}/compose/cancel?offer_hash=...     → single cancel (unchanged)
GET /v2/addresses/{address}/compose/cancel/estimatexcpfees    → fee estimate
```

### Parse

1. `unpack()` detects the 1-byte flag and returns `(None, "valid")`
2. `parse()` checks `offer_hash is None` and `protocol.enabled("cancel_all_offers")`
3. If enabled: validates, pays gas fee, cancels all open orders and bets
4. If not enabled: falls through to standard unpack which returns `"invalid: could not unpack"`

### Cancellation

Cancel-all delegates to the same `order.cancel_order()` and `bet.cancel_bet()` functions used by single cancel. Each cancelled order/bet produces the standard `ORDER_UPDATE`, `CREDIT`, and `BET_UPDATE` events. No new event machinery.

---

## Fee Model

Cancel-all uses the **sigmoid gas fee system** — the same infrastructure used by attach, detach, and pool operations.

### Why sigmoid (not flat per-offer)?

- Sigmoid is the direction for Counterparty fees — one consistent model
- A flat per-offer fee introduces a second fee equation alongside sigmoid
- Under normal usage the fee is zero — cancel-all is a convenience feature
- Under spam the sigmoid ramps up naturally

### Counter increment by offer count

The gas counter increments by the **number of offers cancelled**, not by 1. This means a cancel-all of 100 orders registers as 100 units of work against the sigmoid curve. If someone repeatedly creates and cancels large batches, the fee ramps proportionally.

```python
gas.increment_counter(db, ID, tx["block_index"], count=offer_count)
```

`gas.increment_counter` now accepts an optional `count` parameter (default 1, backward compatible).

### Natural rate limiting

Creating orders requires individual Bitcoin transactions (~$0.04 each today). An address with 10,000 open orders spent ~$400 in BTC fees to get there. This on-chain cost is the primary defense against abuse. The sigmoid fee provides additional protection for repeated use.

---

## Performance

Cancel-all is O(N) in the number of open offers — each cancellation requires individual journal events (`ORDER_UPDATE`, `CREDIT`) for consensus. Benchmarks on test hardware which was just a simple laptop:

| Orders | Parse time |
|--------|-----------|
| 10 | 0.03s |
| 100 | 0.11s |
| 1,000 | 1.6s |
| 10,000 | 22s |

Realistic usage (power user with <2,000 orders) completes in under 4 seconds. The O(N) writes are fundamental to the append-only journal model — the same trade-off dividends make.

---

## Files Changed

| File | Change |
|------|--------|
| `lib/messages/cancel.py` | Cancel-all mode in validate, compose, unpack, parse |
| `lib/messages/gas.py` | Add `count` parameter to `increment_counter` |
| `lib/ledger/markets.py` | Add `get_open_orders_by_source()` |
| `lib/ledger/other.py` | Add `get_open_bets_by_source()` |
| `lib/api/compose.py` | Make `offer_hash` optional, add fee estimate |
| `lib/api/routes.py` | Add fee estimate endpoint |
| `lib/api/apiwatcher.py` | Add `CANCEL_ALL` event |
| `protocol_changes.json` | Add activation gate + fee params for ID 70 |
| `test/units/messages/cancel_test.py` | 16 unit tests |
| `test/units/api/compose_test.py` | API compose + fee estimate tests |
| `test/integrations/regtest/scenarios/scenario_26_cancel_all.py` | Regtest integration scenario |
| `test/integrations/regtest/scenarios_test.py` | Wire in scenario |

---

## Tests

| Type | Count | Covers |
|------|-------|--------|
| `cancel_test.py` | 16 | compose, unpack, validate, parse for both single and cancel-all modes |
| `compose_test.py` | 2 | API cancel-all compose, fee estimate endpoint |
| `scenario_26_cancel_all` | 4 steps | Open 2 orders, cancel-all, verify events/status, error on empty |

---

## Backward Compatibility

Before `cancel_all_offers` activation:
- A 1-byte message with flag `0x01` reaches `unpack()` which sees `len(message) != 32` → `"invalid: could not unpack"`
- Recorded as `INVALID_CANCEL` — identical to any malformed cancel message
- No behavioral change to existing nodes

After activation:
- `parse()` detects the flag and routes to cancel-all logic
- Single cancel (32-byte offer_hash) is completely unchanged


droplister:
I think this is a cool feature and would be useful. I worry about the time it would take to cancel all offers if you have 100k+ orders on your address which isn't something we've seen but it's the one thing that this PR doesn't solve for completely.

adamkrellenstein (changes requested):
let's add release notes too!

Ouziel:
> let's add release notes too!

ah right! done!

adamkrellenstein (approved):
Let's be sure we update the official docs as well. A new message type requires updated specs etc.

PR #3285, opened 2026-04-13
Trustless Pool Creation with Locked Liquidity for Fairminters

droplister:
## Overview

Adds `pool_quantity` and `lp_asset` to the fairminter. When set, the protocol trustlessly creates a TOKEN/XCP AMM pool at soft cap resolution with permanently locked liquidity. No operator, no custodian, no trust required.

This gives Counterparty a **protocol-level launch pad** where issuers can guarantee both fair distribution (via soft caps and per-address limits) and instant liquidity (via locked AMM pools) in a single atomic operation.

Depends on #3281 (AMM pools).

---

## Why this matters

Existing memecoin launch platforms suffer from zero-barrier launches that fragment liquidity and produce thousands of dead tokens. Counterparty's fairmint-pool model is structurally different:

- **Soft cap as quality filter.** A token cannot launch unless enough people independently commit XCP. Failed mints refund automatically.
- **Locked liquidity as trust signal.** LP tokens go to UNSPENDABLE. The pool can never be drained. No rug pulls.
- **Per-address limits as distribution guarantee.** `max_mint_per_address` prevents whales from capturing the launch.
- **XCP as the common denominator.** Every launch locks XCP in permanent pool liquidity — a supply sink that also improves DEX depth.

Issuers who want unlocked LP can create a normal fairminter and deposit to a pool separately after the mint closes. The fairmint-pool path is specifically the trustless, locked, no-rug path.

---

## What makes this safe on Bitcoin L1

The same structural properties that make #3281 safe apply here:

- **No flash loans.** Bitcoin transactions can't call protocol logic recursively. Atomic borrow-swap-repay doesn't exist.
- **No callbacks or reentrancy.** Message parsing is linear; no function hands control to attacker-controlled code.
- **No admin keys.** The fairminter creator cannot unlock LP tokens, dilute the pool, or skim after launch. LP asset issuer is `UNSPENDABLE` — no reissuance, no lock-flip, no dividend-to-self.

What fairmint-pool adds:

- **Deterministic pool creation.** Pool is seeded from on-chain state at soft-cap resolution. No operator runs the migration. No trusted bridge.
- **Locked liquidity at resolution.** LP tokens go to `UNSPENDABLE` at creation — permanently. The only MEV primitive on Bitcoin L1 that applies (within-block tx reordering) is documented in Known Limitations below.

---

### Founder options

All existing fairminter parameters remain configurable alongside `pool_quantity`:

- **`premint_quantity`** — reserve tokens for the founder upfront. Explicit, visible to all participants before they mint. Preminted tokens are not part of the mintable supply.
- **`minted_asset_commission`** — founder receives a percentage of each mint's tokens. Grows proportionally with participation. All XCP still goes to the pool regardless of commission.
- **`max_mint_per_address`** — cap per-wallet allocation to ensure broad distribution. This is an absolute token quantity (e.g. 2,000,000 on a 100M supply = 2% per address).
- **`lp_asset`** — optionally pin the LP token name (numeric asset like `A95428956661682177`). If omitted, the protocol generates one deterministically from the sorted pair. Either way, the name is committed on-chain at fairminter open so every node sees the same LP asset.

### Walkthrough: community fair launch

A 100M token launch with 31% pool, 69% mintable, 2% per-address cap:

```
Fairminter parameters:
  hard_cap:              100,000,000
  pool_quantity:          31,000,000  (31%)
  soft_cap:               69,000,000  (= mintable, enforced)
  price / quantity_by_price:  1 / 10,000  (1 XCP per 10,000 tokens)
  max_mint_per_address:    2,000,000  (2% of total supply)
  soft_cap_deadline_block: ~3 days

What happens:
  - 1 XCP buys 10,000 tokens
  - Each address can mint up to 2M tokens for 200 XCP
  - Minimum 35 unique addresses needed to fill the mint
  - Total XCP raised if filled: 6,900 XCP

If soft cap is met:
  - Pool opens with 31M tokens + 6,900 XCP
  - Pool price: 0.000222 XCP/token (2.23x mint price)
  - LP locked permanently — nobody can drain the pool
  - Token is tradeable at the next block

If soft cap is not met:
  - All XCP returned, all tokens destroyed, no pool
```

### Sell pressure absorption (31/69 split, 6,900 XCP raise)

Pool opens at 2.23x mint price. As minters sell tokens into the pool:

| Tokens sold into pool | Pool price | Avg price received |
| --- | --- | --- |
| 5M (5%) | 1.65x mint | 1.91x mint |
| 10M (10%) | 1.27x mint | 1.68x mint |
| 15M (15%) | 1.01x mint | 1.49x mint |
| 25M (25%) | 0.68x mint | 1.23x mint |
| 50M (50%) | 0.33x mint | 0.85x mint |
| 69M (69%) — all minted | 0.21x mint | 0.69x mint |

Pool price reaches mint price after ~15% of supply is sold. Even in a worst-case full dump of all minted tokens, sellers recover 68% of their XCP and the pool permanently retains 2,147 XCP (31%) as liquidity.

---

## How it works

```
Issuer creates fairminter with pool_quantity (+ optional lp_asset)
  │
  ├─ Tokens reserved for pool escrowed at UNSPENDABLE
  ├─ lp_asset name earmarked (blocks issuance, pooldeposit, competing fairminter)
  ├─ Minters pay XCP, tokens + XCP escrowed during soft cap period
  │
  ├─ Soft cap resolution (at deadline block OR the block hard_cap is filled):
  │     Yes → Pool created end-of-block: pool_quantity tokens + all XCP
  │            LP tokens → UNSPENDABLE (locked forever)
  │            Minters receive their tokens
  │            Token is tradeable at the next block
  │     No  → All XCP refunded, all tokens destroyed, no pool
  │
  └─ lp_asset earmark released when fairminter closes.
```

Pool creation always happens in the block's `after_block` hook (never mid-tx), so orders in the same block cannot match against the fresh pool. First trade is next block.

---

## LP asset earmark

When a fairminter-pool is opened, its `lp_asset` name is reserved on-chain. Until the fairminter resolves, three other code paths reject it:

- `issuance.validate` — nobody else can directly issue the earmarked asset (even as a free numeric issuance).
- `fairminter.validate` — a second fairminter cannot request the same `lp_asset`.
- `pooldeposit.validate` — nobody can manually first-deposit using this `lp_asset` as the pool's LP token.

Released the moment the fairminter transitions to `closed` or `invalid`. Enforced by a new helper `ledger.issuances.get_active_fairminter_by_lp_asset` on a new index on `fairminters.lp_asset`.

---

## Invariants the code enforces

- `soft_cap == mintable` when `pool_quantity > 0` — all-or-nothing launches, no post-deadline minting.
- Pool reserves exactly equal what was escrowed: `reserve_a + reserve_b == pool_quantity_tokens + raised_XCP`. No slippage at creation.
- Initial LP supply = `isqrt(reserve_a * reserve_b)` credited entirely to `UNSPENDABLE`. Verified in regtest by balance check.
- `lp_asset` is committed on-chain at fairminter open and identical on every node — no process-local randomness anywhere in the pool-creation path.
- At most one active fairminter earmarks a given `lp_asset` at any block. Enforced by `get_active_fairminter_by_lp_asset` over `status IN ('open', 'pending')`.
- Pool creation runs in `after_block`, never mid-tx. First trade against the new pool is always next block.
- Every new row (`fairminters.lp_asset` column, `pools`, `pool_deposits`, LP `issuances`, LP `credits`) carries `block_index` and flows through `parser/blocks.py:TABLES` — reorg across a hard-cap-hit block is traced clean.

---

## Design constraints

| Constraint | Rationale |
| --- | --- |
| `soft_cap == mintable` enforced | All-or-nothing launches. Prevents post-deadline minting where XCP goes to issuer instead of pool — misaligned incentives and arbitrage risk. |
| LP always locked at UNSPENDABLE | The "trustless" in the title. Issuer cannot drain pool. Note: only the initial LP from fairmint resolution is locked — anyone can later add liquidity via manual `pooldeposit` and their LP is regular transferable. |
| `price > 0` required | Free mints produce no XCP for the pool. |
| `burn_payment` incompatible | Burned XCP can't seed a pool. |
| Existing supply accounted for | `mintable = hard_cap - existing_supply - premint - pool_quantity`. Works with both new and existing assets. |
| Pool must not already exist | Checked at fairminter open; manual `pooldeposit` for the pair is also blocked while a fairmint-pool is active. Reaching resolution with the pool nonetheless present is an invariant violation and halts the block — no refund-fallback (the setup is unreachable under the validators). |
| Gas fee charged upfront | Pool deposit gas fee paid at fairminter creation, even if soft cap fails. |
| Pool creation deferred to end-of-block | Even when `hard_cap` fills before the scheduled deadline, pool creation runs in `after_block`, never mid-tx. Orders in the same block see no pool. |
| `lp_asset` earmarked until resolution | Prevents same-name issuance, competing fairminter, or manual pooldeposit racing the launch. |

### Pool economics note

The pool opening price relative to mint price is determined by the ratio `mintable / pool_quantity`:

| Pool % | Mintable % | Premium |
| --- | --- | --- |
| 10% | 90% | 9.0x |
| 20% | 80% | 4.0x |
| 31% | 69% | 2.2x |
| 40% | 60% | 1.5x |
| 50% | 50% | 1.0x |

When `pool_quantity > mintable` (pool > 50%), the pool opens below mint price. This isn't prevented — issuers can choose their economics — but wallet UIs should surface the premium so minters understand what they're getting.

Since the initial LP is locked at UNSPENDABLE, the fairmint-pool also provides a **permanent liquidity floor**: even if all subsequent LPs withdraw their share, reserves proportional to the locked LP remain in the pool forever.

---

## Out of scope

Deliberately not attempting in this PR:

- **No commit-reveal on mint ordering.** Within-block last-slot snipe is accepted as inherent on-chain MEV (same frame as #3281). Protocol fix would require a reveal phase or L2.
- **No earmark duration cap.** Consistent with base fairminter, which already allows arbitrarily-long asset-name holds. Economic cost of squatting a numeric name (18 quintillion available) doesn't justify new policy.
- **No partial-fill of the last mint slot.** Fairmints that exceed hard_cap are rejected whole, matching existing fairminter semantics.
- **No bonding-curve pricing during the mint.** Price is fixed by `price / quantity_by_price` at open. A declining-price LBP-style shape would be a separate feature.
- **No whitelist / allowlist.** Fairmints are permissionless; `max_mint_per_address` is the only per-participant filter.

---

## Wire format

CBOR array grows to 21 elements. Field 19 is `pool_quantity`; field 20 is `lp_asset_id` (uint64 encoding of the LP asset, 0 when absent). Messages with fewer elements default both to 0. Backwards compatible with pre-pool and pre-earmark messages.

The LP asset name is **committed at fairminter open**, not generated at soft-cap close. Random generation at close would be non-deterministic across nodes — a consensus fork risk. Committing at open makes every node see the same LP asset name from first write.

---

## Formally-proven pool math (inherited)

Pool math in this PR (constant product, `isqrt`, k-monotonic under swap, LP-mint bounds via `min(A-basis, B-basis)`, post-swap reserve non-negativity) is formally verified in #3281 via CrossHair + Z3 over realistic input ranges. Fairmint-pool calls the same functions directly — no new math to re-prove, only new orchestration: *when* pool creation happens, *what* reserves are passed in, *where* LP goes.

---

## Rigor

Intended as a demonstration that composing protocol primitives can ship with the same care as conservative fixes.

| Layer | What |
| --- | --- |
| Unit tests | 68 tests across `fairminter_test`, `fairmint_test`, `pooldeposit_test`: CBOR roundtrip (21-element), compose/unpack/parse with `pool_quantity` + `lp_asset`, pool validation, earmark collision, realistic pro-rata refund fixture, halt-on-invariant-violation. |
| Regtest `scenario_27` | 17 steps end-to-end: single-minter happy path, multi-minter happy path with explicit `lp_asset` and UNSPENDABLE-balance verification, soft-cap-fail refund, burn_payment / free / soft_cap-mismatch rejections, earmark collision (second fairminter rejected, direct issuance rejected). |
| Cross-feature interaction audit | Every Counterparty message type × fairmint-pool state. `dispense` / `dividend` / `attach` / `destroy` / `sweep` cannot touch LP tokens because UNSPENDABLE holds them and has no signing key. Earmark check closes the three compose-time collision points (issuance, fairminter, pooldeposit). |
| Rollback / reorg audit | Every new table insert carries `block_index`; new `fairminters.lp_asset` and pool-creation writes flow through `parser/blocks.py:TABLES`. Reorg across a hard-cap-hit block traces clean: `insert_update` appends rows, rollback removes by block_index, latest-state restores to pre-hard-cap values. |
| Compose/validate symmetry scrub | Checks moved into `validate()` (lot alignment, free-fairminter-quantity rejection, earmark) so parse-time ingest rejects the same shapes that compose rejects. Closes the hand-rolled-CBOR bypass. |
| Adversarial multi-agent review | Three parallel attacker personas (greedy issuer, whale minter, miner ordering). After verifying claims against code: 0 real novel findings. Rounding-based dust-mint attack against the whale persona: unprofitable (attacker overpays, cannot recover through pool). Miner-flagged double-execution on hard-cap-at-deadline: already blocked by inline status update at `fairmint.py:333`. |
| Launchpad post-mortem mapping | pump.fun migration rug / sniper MEV, Fjord Foundry LBP weight-boundary MEV, daos.fun treasury drain, Camelot Genesis race, Odin.fun / Four.Meme: structurally inapplicable due to UNSPENDABLE-locked LP + UNSPENDABLE issuer + deferred end-of-block pool creation + no admin surface. |

### Known limitations

- **Within-block last-slot mint snipe.** A miner ordering the deadline block can insert their own fairmint as the final soft-cap-filling tx and capture the minting slot. Inherent to any on-chain permissionless fairmint with a bounded cap — no protocol fix without commit-reveal or L2 settlement. Same class as PR #3281's acknowledged within-block reordering.
- **Cross-block MEV N→N+1.** Resting orders in the pool-creation block match at N+1 alongside fresh N+1 mempool. Standard DEX MEV; mitigated by orderers using limit prices and slippage parameters.
- **Latent footgun in append-only lookups.** `get_fairminters_to_open` (`start_block` filter) and `get_fairminters_to_close` (`end_block` filter) share the same "filter inside GROUP BY, `MAX(rowid)` on subset" pattern as the query this PR fixes. Both are safe today because nothing mutates those columns, but the pattern is fragile — worth a follow-up cleanup to make all three queries robust to future mutation.

---

## How to call it

```
GET /v2/addresses/{address}/compose/fairminter
    ?asset=MYTOKEN
    &lot_price=1
    &hard_cap=100000000
    &soft_cap=69000000
    &soft_cap_deadline_block=...
    &pool_quantity=31000000
    &lp_asset=A95428956661682177   # optional; auto-generated if omitted
    &lock_quantity=True
```

All other fairminter endpoints work unchanged.


adamkrellenstein:
@droplister @ouziel-slama what do we think about including this in v11.1.0?

PR #3296, opened 2026-05-15
Handle invalid CBOR broadcast payloads

SENDNOX:
Invalid CBOR broadcast arrays can currently decode as `valid` with non-numeric fields and then raise a `TypeError` in `broadcast.parse()` when `min(value, config.MAX_INT)` runs.

This adds schema/type checks for CBOR broadcast payloads so malformed arrays are recorded as `invalid: could not unpack`, while non-CBOR data can still fall back to the legacy decoder.

Tests run:
- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/messages/broadcast_test.py -q`
- `ruff check counterparty-core/counterpartycore/lib/messages/broadcast.py counterparty-core/counterpartycore/test/units/messages/broadcast_test.py`

Broader check:
- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/messages -q` currently has one order-dependent failure in `issuance_test.py::test_reset_issuance`; rerunning that test alone passes.

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/broadcast.py
+++ b/counterparty-core/counterpartycore/lib/messages/broadcast.py
@@ -182,9 +182,25 @@ def compose(
 
 
 def load_cbor(message):
-    timestamp, value, fee_fraction_int, mime_type, text = cbor2.loads(message)
+    decoded = cbor2.loads(message)
+    if not isinstance(decoded, (list, tuple)) or len(decoded) != 5:
+        raise exceptions.UnpackError
+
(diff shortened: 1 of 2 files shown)


ouziel-slama:
Thanks for all these PRs! 
hum.. the final effect (invalid: could not unpack) is the same when falling into the fallback of the legacy decoder, right? If not, it probably needs to be attributed to a protocol change.

ouziel-slama:
after verification, already fixed by commits 909f64011

PR #3297, opened 2026-05-15
Handle invalid dispenser asset ids

SENDNOX:
## Summary

Invalid dispenser payloads with a uint64 value below the Counterparty asset-id range currently raise `AssetIDError` from `ledger.issuances.generate_asset_name()` during `dispenser.unpack()`.

That exception bypasses the existing invalid-unpack handling, so malformed on-chain dispenser data can escape parsing instead of being classified as `invalid: could not unpack`.

This PR catches `AssetIDError` in the existing invalid-unpack path and adds a regression test for asset id `2`.

## Tests

- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/messages/dispenser_test.py::test_unpack_invalid_asset_id -q`
- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/messages/dispenser_test.py -q`
- `ruff check counterparty-core/counterpartycore/lib/messages/dispenser.py counterparty-core/counterpartycore/test/units/messages/dispenser_test.py`

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/dispenser.py
+++ b/counterparty-core/counterpartycore/lib/messages/dispenser.py
@@ -332,7 +332,7 @@ def unpack(message, return_dict=False, block_index=None):
             oracle_address = address_unpack(message[read : read + 21])
         asset = ledger.issuances.generate_asset_name(assetid)
         status = "valid"
-    except (exceptions.UnpackError, struct.error):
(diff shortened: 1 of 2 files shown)


ouziel-slama:
Likewise, it is necessary to verify that this does not require a change of protocol.

PR #3298, opened 2026-05-15
Handle short subasset issuance payloads

SENDNOX:
## Summary

Short legacy subasset issuance payloads currently raise `struct.error` inside `issuance.unpack()` after CBOR decode falls back to legacy parsing.

That exception bypasses the existing invalid-unpack handling, so malformed subasset issuance data can escape parsing instead of being classified as `invalid: could not unpack`.

This PR catches `struct.error` in the existing invalid-unpack path and adds a regression test for an empty `SUBASSET_ID` payload.

## Tests

- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/messages/issuance_test.py::test_unpack_short_subasset_message -q`
- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/messages/issuance_test.py -q`
- `ruff check counterparty-core/counterpartycore/lib/messages/issuance.py counterparty-core/counterpartycore/test/units/messages/issuance_test.py`

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/issuance.py
+++ b/counterparty-core/counterpartycore/lib/messages/issuance.py
@@ -811,7 +811,7 @@ def unpack(db, message, message_type_id, block_index, return_dict=False):
         except exceptions.AssetIDError:
             asset = None
             status = "invalid: bad asset name"
-    except exceptions.UnpackError as e:
+    except (exceptions.UnpackError, struct.error) as e:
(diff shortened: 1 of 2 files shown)


ouziel-slama:
Likewise, it is necessary to verify that this does not require a change of protocol.

ouziel-slama:
after verification, already fixed in bf7e38d4a

PR #3299, opened 2026-05-15
Fix divisible dispenser prices

### LINKED ISSUE #2882: Dispenser price is off by 10**8
the unit price should be the same as the lot price except for the first entry, where it should be half (because the lot size is 2)

https://dev.counterparty.io:4000/v2/assets/FAKEASF/dispensers?offset=0&limit=10&verbose=true&sort=price:asc&status=open

SENDNOX:
## Summary

Fixes #2882.

The API-derived dispenser `price` currently uses raw atomic `give_quantity`, so divisible asset dispensers report satoshis-per-atomic-unit instead of satoshis-per-whole-asset-unit. For a divisible asset lot of `2.0` units at `12345` sats, the raw API `price` should be `6172.5`, not `0.000061725`.

This PR scales the SQL-derived raw `price` by `1e8` when `assets_info.divisible` is true, while leaving indivisible assets on the existing raw lot-size calculation.

## Tests

- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/api/queries_test.py::test_get_dispensers_by_asset_prices_divisible_lots -q`
- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/api/queries_test.py -q`
- `ruff check counterparty-core/counterpartycore/lib/api/queries.py counterparty-core/counterpartycore/test/units/api/queries_test.py`

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/queries.py
+++ b/counterparty-core/counterpartycore/lib/api/queries.py
@@ -2436,7 +2436,14 @@ def prepare_dispenser_where(status, other_conditions=None, exclude_with_oracle=F
     return where
 
 
-SELECT_DISPENSERS = "*, (satoshirate * 1.0) / (give_quantity * 1.0) AS price"
+SELECT_DISPENSERS = """
+*,
+CASE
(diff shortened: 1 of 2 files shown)


ouziel-slama:
LGTM

adamkrellenstein (approved):

PR #3300, opened 2026-05-15
Handle invalid UTXO messages

SENDNOX:
## Summary

Malformed legacy UTXO messages currently raise `UnpackError` from `utxo.unpack()` before `utxo.parse()` can record an invalid transaction.

That exception can escape block parsing instead of classifying the transaction as `invalid: could not unpack`.

This PR handles UTXO unpack failures in `parse()`, records an invalid `sends` row, marks the transaction invalid, and adds a regression test for a payload without the required four fields.

## Tests

- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/messages/utxo_test.py::test_parse_invalid_unpack -q`
- `.venv/bin/pytest counterparty-core/counterpartycore/test/units/messages/utxo_test.py -q`
- `ruff check counterparty-core/counterpartycore/lib/messages/utxo.py counterparty-core/counterpartycore/test/units/messages/utxo_test.py`

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/utxo.py
+++ b/counterparty-core/counterpartycore/lib/messages/utxo.py
@@ -154,7 +154,20 @@ def unpack(message, return_dict=False):
 
 
 def parse(db, tx, message):
-    (source, destination, asset, quantity) = unpack(message)
+    try:
+        (source, destination, asset, quantity) = unpack(message)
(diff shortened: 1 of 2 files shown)


ouziel-slama:
utxo contract is not anymore used (replaced by attach and detach)

PR #3301, opened 2026-05-15
Validate issuance CBOR payloads

SENDNOX:
Malformed taproot/CBOR issuance payloads can currently reach `validate()` with incorrectly typed fields while still marked `valid`. For example, a CBOR issuance array with a string `quantity` raises `ValueError: not enough values to unpack` in `parse()` because `validate()` returns its shorter invalid tuple contract.

This PR validates decoded issuance CBOR payload shape and field types before accepting them as unpacked messages. Decode failures and non-issuance CBOR shapes still fall back to legacy unpacking, preserving legacy payload behavior; decoded issuance arrays with invalid field types now become `invalid: could not unpack`.

Verification:
- `../.venv/bin/pytest counterpartycore/test/units/messages/issuance_test.py::test_parse_invalid_cbor_schema -q`
- `../.venv/bin/pytest counterpartycore/test/units/messages/issuance_test.py::test_parse_invalid_cbor_schema counterpartycore/test/units/messages/issuance_test.py::test_parse_too_short counterpartycore/test/units/messages/issuance_test.py::test_parse_divisible_legacy_taproot_acivated -q`
- `../.venv/bin/pytest counterpartycore/test/units/messages/issuance_test.py -q`
- `ruff check counterpartycore/lib/messages/issuance.py counterpartycore/test/units/messages/issuance_test.py`
- `git diff --check`

BTC payout address: `bc1qehva4gmx68nhktywxtcfkwkvqknc3zpe4es75a`

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/issuance.py
+++ b/counterparty-core/counterpartycore/lib/messages/issuance.py
@@ -34,6 +34,76 @@ DESCRIPTION_MARK_BYTE = b"\xc0"
 DESCRIPTION_NULL_ACTION = "NULL"
 
 
+def _is_int(value):
+    return isinstance(value, int) and not isinstance(value, bool)
+
+
+def _is_boolish(value):
+    return isinstance(value, bool) or (_is_int(value) and value in (0, 1))
+
+
+def _is_optional_text(value):
+    return value is None or isinstance(value, str)
+
+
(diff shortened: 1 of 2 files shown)


ouziel-slama:
after verification, already fixed in 909f64011

PR #3319, opened 2026-05-21
Validate broadcast compose numeric fields

vultuk:
Summary:
- Reject broadcast compose timestamps that cannot fit the legacy 32-bit wire field before packing.
- Reject non-finite broadcast values and non-finite or out-of-range fee fractions before data construction.
- Add unit coverage for these compose-time guards, including validate=false cases that previously reached packing errors.

Why:
Malformed compose inputs such as timestamp=4294967296 or a negative fee_fraction with validate=false can reach struct.pack and surface as an internal server error instead of a normal compose validation error. Non-finite values such as NaN are also not meaningful broadcast payloads and should not be emitted by the composer.

Tests:
- ./.venv/bin/pytest -q counterparty-core/counterpartycore/test/units/messages/broadcast_test.py
- ./.venv/bin/pytest -q counterparty-core/counterpartycore/test/units/messages/broadcast_test.py counterparty-core/counterpartycore/test/units/api/composer_test.py -k broadcast
- ./.venv/bin/pytest -q counterparty-core/counterpartycore/test/units/api/compose_test.py
- git diff --check

Bounty claim:
If this qualifies under the Counterparty bug bounty, please pay the BTC portion to: bc1qa5acxqc3wldzxwjhe8a65gjp6n7dxmafm6qs4j

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/broadcast.py
+++ b/counterparty-core/counterpartycore/lib/messages/broadcast.py
@@ -21,6 +21,7 @@ because it is stored as a four‐byte integer, it may not be greater than about
 
 import decimal
 import logging
+import math
 import struct
 from fractions import Fraction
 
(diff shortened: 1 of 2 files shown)


ouziel-slama:
done here https://github.com/CounterpartyXCP/counterparty-core/pull/3371/changes/c7d1a0cb9664099eb213366b3ed95d4d96e585a2

PR #3324, opened 2026-06-05
Allow valid large pool deposits (gated)

adamkrellenstein:
## Summary

Removes a redundant product-overflow validation in `pooldeposit.validate()` that rejects otherwise valid AMM pool deposits, gated behind a new protocol change so it is replay-safe after the AMM activates.

The check `quantity_a * quantity_b > config.MAX_INT` rejected every deposit whose product exceeds `MAX_INT`, even when both quantities, the resulting LP quantity, balances, and stored reserves are individually valid. For balanced divisible assets it caps each side at `floor(sqrt(MAX_INT)) = 3,037,000,499` raw units (30.37000499 divisible units) — so a normal 100-unit-per-side deposit (`10,000,000,000` each) is rejected.

The product is only consumed by arbitrary-precision Python integer arithmetic (`math.isqrt`) and is never persisted. With both quantities `<= MAX_INT`, `isqrt(quantity_a * quantity_b) <= MAX_INT`, and `quantity_minted` is bounded separately. The check protects no stored value.

## Consensus safety

The AMM activates at mainnet block **952800**. This fix lands in a later release, i.e. **after** the feature is live, so the removal is gated behind the new `fix_pool_deposit_product_overflow` protocol change. Pre-activation consensus is unchanged.

> [!IMPORTANT]
> The activation heights in `protocol_changes.json` are placeholders (`99999999`) and **must be set to the real release heights before merge**. As shipped (far-future), the fix stays disabled — safe default, no accidental fork.

## Changes

- gate the `quantity_a * quantity_b` check behind `fix_pool_deposit_product_overflow`
- add the `protocol_changes.json` entry (placeholder heights)
- replace the product-overflow regression test with one asserting individually valid large quantities are accepted

## Validation

- `pooldeposit_test.py`: 49 passed
- `ruff check` on changed Python: clean
- independently verified `isqrt(quantity_a*quantity_b) <= MAX_INT` for both `<= MAX_INT`

Bug reported by GioLealhmx.

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/pooldeposit.py
+++ b/counterparty-core/counterpartycore/lib/messages/pooldeposit.py
@@ -6,7 +6,7 @@ import struct
 from counterpartycore.lib import config, exceptions, ledger
 from counterpartycore.lib.ledger.currentstate import CurrentState
 from counterpartycore.lib.messages import gas
-from counterpartycore.lib.parser import messagetype
+from counterpartycore.lib.parser import messagetype, protocol
 from counterpartycore.lib.utils import assetnames
 
 logger = logging.getLogger(config.LOGGER_NAME)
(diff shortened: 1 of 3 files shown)


droplister:
Dang it, I am embarrassed by this one.

droplister (approved):

ouziel-slama (approved):

PR #3325, opened 2026-06-05
Fix AMM best-price routing under integer pool math (gated)

adamkrellenstein:
## Summary

Fixes AMM best-price routing under integer pool math, gated behind a new protocol change.

`compute_pool_input_for_target_price()` solved a continuous-price quadratic while execution floors the pool output to an integer (`compute_pool_output`). That mismatch broke the documented best-price routing guarantee in **two** directions:

- **Underfill** — the continuous input could floor to **zero** output, so `try_pool_fill()` rejected a pool fill that was cheaper than the next book order, and the taker filled the **worse book order** instead. (Originally reported: taker spends 4 XCP where the pool could do 2.)
- **Overfill** — bounding on the post-fill **marginal** price let the input grow until the **realized average** price (output floored) exceeded the book price, so the taker overpaid the pool versus an available book order. Demonstrated end-to-end: taker pays **2 XCP for 1 NODIVISIBLE** where current master correctly pays **1**.

The originally proposed patch addressed only the underfill case and **introduced** the overfill regression (it bounds on marginal price without trimming).

## Fix

Replace the quadratic with an integer-exact computation evaluated against the same output function execution uses:

1. Binary search the largest input whose **marginal** post-fill price still beats the target (bounds how much the pool should take — filling further buys units costing more than the book).
2. **Trim** to the smallest input that still yields that floored output (so the taker never overpays for the rounded-down units).

The realized average price of the result is then always at or better than the book target.

## Verification

- Exhaustive small-reserve sweep + random large-boundary samples (**18M+ parameter combinations**): the new function never charges the taker more than the book would for the same units in integer terms (`0` genuine overpays), and exactly matches an independent taker-optimal oracle.
- Both end-to-end directions covered by new regression tests.
- `pool_match_test.py`, `markets_test.py`, `order_test.py`, `pooldeposit_test.py`: 104 passed.
- `ruff check` on changed files: clean.

## Consensus safety

The AMM activates at mainnet block **952800**; this lands in a later release (after the feature is live), so it is gated behind the new `fix_pool_best_price_routing` protocol change. `block_index` is threaded through `try_pool_fill` to the gate. Pre-activation consensus is unchanged (old quadratic retained for replay).

> [!IMPORTANT]
> The activation heights in `protocol_changes.json` are placeholders (`99999999`) and **must be set to the real release heights before merge**. As shipped (far-future), the fix stays disabled — safe default.

## Changes

- integer-exact `_compute_pool_input_for_target_price_integer`, gated behind `fix_pool_best_price_routing`
- thread `block_index` through `try_pool_fill` and both `order.py` call sites
- `protocol_changes.json` entry (placeholder heights)
- regression tests for both the underfill and overfill directions

Bug reported by GioLealhmx.


droplister (approved):

droplister:
---

The underfill diagnosis is right, and the gating/`block_index` plumbing is good. But bounding on **marginal** price lets the **average** price the taker pays drift past the book once output is floored — so this swaps the underfill for an overfill. Reproduced on this branch:

- reserves in=5 out=6, fee 50bps; book 10 XCP for 7 NODIVISIBLE; taker gives 2 XCP for 1.
- branch fills **2 XCP → 1 NODIVISIBLE**; book would charge **1** (`round(1*10/7)=1`); master pays 1.

Worse than master, breaks the "never worse than book" claim. (Trying to make it exactly optimal is a rabbit hole — flooring makes average price non-monotone in input, no clean closed-form. Not worth it.)

**Proposed rule:** bound on average price, gate on the book's integer cost. One-directional guarantee — never overpay; may underfill a dust unit (taker takes it from the book). 0 overpays across ~836k reserve/price/fee combos.

**Scope of change** (solver body only, everything else stands):
- replace the marginal binary-search bound with a closed-form average-price crossover: `dx = (num·ff·ro − den·ri·10000) // (ff·den)`
- add a `dx ≤ book_cost(co(dx))` gate (step down to the largest non-overpaying input) before the existing trim
- new helper `_book_input_for_output()` = `round(output·num/den)` half-to-even, matching `order.py:675–679`
- verification oracle must be that integer book cost, not the continuous price — that's the blind spot that hid both the underfill and this overfill

Can push the commit + a regression test off the integer oracle if you're good with the direction. Thoughts?

ouziel-slama:
Pushed a commit with droplister's core fix — **gate on the book's integer cost** — but kept the **marginal** bound instead of the average crossover.

**The real gap was the `max_give` cap.** `try_pool_fill` caps the solver result *after* the solver, without re-checking the book price — that's why droplister's case (pool 5/6, book 10 XCP/7 NODIVISIBLE, taker has 2 XCP) overpaid. Fix: thread `max_give` into the solver and gate **after** the cap.

**Marginal + gate, not average + gate.** Both give 0 overpays — the gate is what fixes the bug. But across the small-reserve sweep, average + gate charges the taker more in ~43% of cases (up to ~2× optimal); marginal + gate matches the brute-force optimal and keeps the documented routing rule + `test_target_price_output_reaches_target` (which the average bound breaks).

Gate uses `round(output*num/den)` via `Fraction`, matching `order.match()`. **Verified:** 0 overpays across the exhaustive small sweep (every cap) + 400k large random samples vs an independent oracle. Added a regression test for droplister's case (fails on marginal+trim, passes with the gate). Merged `develop`; heights still `99999999`.

Does this direction look good to you both? @adamkrellenstein @droplister

ouziel-slama (approved):

PR #3340, opened 2026-06-09
Skip call price checks after callability parameter removal

### LINKED ISSUE #1338: Cannot lock callable asset
9.60 introduced a bug that made it impossible to lock callable assets (and likely do any other issuance).
Should not 9.61 have fixed this with the new issuance id 22 ?

Recent tx with ID 22 failed.
https://xchain.io/tx/cbbac8a8471e7bc1e5f6e06fdbbf35c41cce20e15c74ffc43d55576534cc582d
https://jpja.github.io/Electrum-Counterparty/decode_tx?tx=cbbac8a8471e7bc1e5f6e06fdbbf35c41cce20e15c74ffc43d55576534cc582d

_Note: Only a few callable assets were issued in 2014, so this bug is very rare._

@pataegrillo @jdogresorg

caydyan:
Fixes #1338.

`issuance_callability_parameters_removal` already skips the reissuance check for changing `callable`, but validation still rejected old callable assets when newer lock/reset issuance formats decoded removed callability fields as `call_price=0.0`. That left old callable assets unable to be locked because `last_issuance["call_price"] > call_price` still produced `cannot reduce call price`.

This keeps the old validation behavior before the protocol change, but skips all deprecated callability-parameter comparisons after `issuance_callability_parameters_removal` is active:
- `callable`
- `call_date`
- `call_price`

Tests:
- `.venv/bin/ruff format counterpartycore/lib/messages/issuance.py counterpartycore/test/units/messages/issuance_test.py`
- `.venv/bin/ruff check counterpartycore/lib/messages/issuance.py counterpartycore/test/units/messages/issuance_test.py`
- `.venv/bin/pytest counterpartycore/test/units/messages/issuance_test.py::test_validate`
- `.venv/bin/pytest counterpartycore/test/units/messages/issuance_test.py` — 30 passed
- `git diff --check`

BTC payout address, if applicable: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/issuance.py
+++ b/counterparty-core/counterpartycore/lib/messages/issuance.py
@@ -130,7 +130,7 @@ def validate(
             (not protocol.enabled("cip03", block_index)) or (not reset)
         ):
             problems.append("cannot change divisibility")
-        if not protocol.enabled("issuance_callability_parameters_removal", block_index):
+        if not protocol.enabled("issuance_callable_lock_fix", block_index):
             if bool(last_issuance["callable"]) != bool(callable_):
                 problems.append("cannot change callability")
             if last_issuance["call_date"] > call_date and (
--- a/counterparty-core/counterpartycore/lib/messages/issuance.py
+++ b/counterparty-core/counterpartycore/lib/messages/issuance.py
@@ -130,20 +130,19 @@ def validate(
             (not protocol.enabled("cip03", block_index)) or (not reset)
         ):
             problems.append("cannot change divisibility")
-        if (not protocol.enabled("issuance_callability_parameters_removal", block_index)) and bool(
-            last_issuance["callable"]
-        ) != bool(callable_):
-            problems.append("cannot change callability")
-        if last_issuance["call_date"] > call_date and (
-            call_date != 0
-            or (
-                not protocol.enabled("default_values_for_callable")
-                and not protocol.is_test_network()
-            )
-        ):
-            problems.append("cannot advance call date")
-        if last_issuance["call_price"] > call_price:
-            problems.append("cannot reduce call price")
+        if not protocol.enabled("issuance_callability_parameters_removal", block_index):
+            if bool(last_issuance["callable"]) != bool(callable_):
+                problems.append("cannot change callability")
+            if last_issuance["call_date"] > call_date and (
+                call_date != 0
+                or (
+                    not protocol.enabled("default_values_for_callable")
+                    and not protocol.is_test_network()
+                )
+            ):
+                problems.append("cannot advance call date")
+            if last_issuance["call_price"] > call_price:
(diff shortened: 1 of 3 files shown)


ouziel-slama:
Thanks for the fix — the bug is real and the local refactor is correct. However, this can't be merged as-is: it's a **retroactive consensus change**.

`validate()` runs in the parse path ([issuance.py#L993](https://github.com/CounterpartyXCP/counterparty-core/blob/master/counterparty-core/counterpartycore/lib/messages/issuance.py#L993)) and its `problems` drive the tx status, so relaxing the `call_price`/`call_date` checks under `issuance_callability_parameters_removal` (active since block **819300**, ~Dec 2023) flips already-confirmed txs from *invalid* → *valid* — e.g. `cbbac8a8…` from #1338. Upgraded vs. non-upgraded nodes would then compute different ledgers (fork on resync).

Consensus fixes in this repo must be gated behind a **new** protocol change with a **future** activation block (the current upgrade batch targets block **952800**). Could you please:

1. Add a new key in `protocol_changes.json` (e.g. `issuance_callable_lock_fix`) with `block_index: 952800`,
2. Gate the relaxed checks behind `protocol.enabled("issuance_callable_lock_fix", block_index)` instead of reusing `issuance_callability_parameters_removal`,
3. Add an activation test (before/after the new block).

The unit tests are a good base — they just need to cover the new gate. 🙏

caydyan:
Thanks for the review. I updated this to avoid a retroactive consensus change:

- Added a new `issuance_callable_lock_fix` protocol change at mainnet block 952800, matching the existing 11.1.0 testnet/signet activation batch.
- `issuance.validate()` now relaxes the old callable/call_date/call_price reissuance checks only behind that new gate.
- Updated the issuance validation coverage to assert the old behavior at 952799 and the relaxed behavior at 952800, with a disabled-gate regression check as well.

Tests run:

- `.venv/bin/python -m pytest counterpartycore/test/units/messages/issuance_test.py::test_validate -q`
- `.venv/bin/python -m pytest counterpartycore/test/units/messages/issuance_test.py -q`
- `.venv/bin/python -m ruff check counterpartycore/lib/messages/issuance.py counterpartycore/test/units/messages/issuance_test.py`
- `.venv/bin/python -m ruff format --check counterpartycore/lib/messages/issuance.py counterpartycore/test/units/messages/issuance_test.py`
- `.venv/bin/python -m json.tool counterpartycore/protocol_changes.json`
- `git diff --check`

caydyan:
Opened a develop-based replacement here: https://github.com/CounterpartyXCP/counterparty-core/pull/3434

It keeps the new future issuance_callable_lock_fix protocol gate requested in review, includes the before/after activation coverage, and was re-verified against current develop. Closing this master-targeted PR as superseded so review can continue on the active branch.

PR #3341, opened 2026-06-09
Skip oracle fee for dispenser closes

### LINKED ISSUE #1271: paying oracle fee when closing dispensers
In testing dispensers, I discovered that when closing a dispenser, if you specify the `oracle_address` in the API request, then the resulting TX includes the oracle fee in the tx.

CP should only collect an oracle fee on OPEN and REFILL actions and not CLOSE.

CP API should be updated to only include the oracle fee when `oracle_address` is defined and status is 0 (OPEN)

caydyan:
Fixes #1271.

Dispenser close transactions do not consume oracle pricing or enforce oracle-fee payment during parsing, but compose/validation still treated any `oracle_address` as requiring a current oracle price and a fee output.

This limits oracle price validation and oracle fee output generation to dispenser open/refill actions (`STATUS_OPEN` and `STATUS_OPEN_EMPTY_ADDRESS`). Close messages can still encode an oracle address, but they no longer add a BTC fee output or fail validation just because the oracle has no current price.

Tests:
- `.venv/bin/ruff format counterpartycore/lib/messages/dispenser.py counterpartycore/test/units/messages/dispenser_test.py`
- `.venv/bin/ruff check counterpartycore/lib/messages/dispenser.py counterpartycore/test/units/messages/dispenser_test.py`
- `.venv/bin/pytest counterpartycore/test/units/messages/dispenser_test.py::test_compose_close_with_oracle_does_not_pay_fee counterpartycore/test/units/messages/dispenser_test.py -q` — 12 passed
- `.venv/bin/pytest counterpartycore/test/units/api/compose_test.py::test_compose_dispenser`
- `git diff --check`

BTC payout address, if applicable: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/dispenser.py
+++ b/counterparty-core/counterpartycore/lib/messages/dispenser.py
@@ -219,7 +219,11 @@ def validate(
 
     cursor.close()
 
-    if oracle_address is not None and protocol.enabled("oracle_dispensers", block_index):
+    if (
+        oracle_address is not None
+        and protocol.enabled("oracle_dispensers", block_index)
+        and status in [STATUS_OPEN, STATUS_OPEN_EMPTY_ADDRESS]
+    ):
         last_price, _last_fee, _last_label, _last_updated = other.get_oracle_last_price(
             db, oracle_address, block_index
         )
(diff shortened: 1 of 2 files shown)


ouziel-slama:
Thanks for tackling #1271. The `compose()` change is correct: parse already only enforces the oracle fee for OPEN/EMPTY ([dispenser.py#L419](https://github.com/CounterpartyXCP/counterparty-core/blob/master/counterparty-core/counterpartycore/lib/messages/dispenser.py#L419)), and the on-chain message bytes are unchanged, so this just stops over-charging the wallet on closes. 👍

**Blocker — the `validate()` change is a consensus change that isn't gated.** `validate()` is also called from `parse()` ([dispenser.py#L399](https://github.com/CounterpartyXCP/counterparty-core/blob/master/counterparty-core/counterpartycore/lib/messages/dispenser.py#L399)) whenever `dispenser_parsing_validation` is active (mainnet block 751485+). Skipping the "oracle has not broadcasted any price yet" check for CLOSE flips the validity of any close tx that encodes an `oracle_address` with no current price: previously `invalid` (dispenser not closed) → now `valid` (dispenser closed). That's a ledger-state divergence between un-upgraded and upgraded nodes on the same block.

It's not a no-op: reverting only this hunk makes `test_compose_close_with_oracle_does_not_pay_fee` fail with `has not broadcasted any price yet`, which also confirms the close→validate→oracle-price path is reachable on real on-chain bytes. (Contrast with the recent "Validate …" PRs, which only add type guards unreachable from on-chain data — consensus-neutral.)

Two ways forward:

1. Gate it behind a new `protocol_changes.json` flag with a future `block_index` (mainnet + testnet3/testnet4) and a `minimum_version_major` bump, e.g. `protocol.enabled("dispenser_close_skip_oracle_price_check", block_index)`.
2. Or drop the `validate()` hunk entirely and keep only the safe `compose()` fix, adjusting the test (oracle with a price, or `skip_validation=True`).

Also worth adding: a parse-side test for a close that encodes an `oracle_address` (the consensus-relevant path is currently untested).

caydyan:
Updated the develop-based replacement here: https://github.com/CounterpartyXCP/counterparty-core/pull/3399

That branch now drops the validate()/parse behavior change called out in review and keeps only the compose-side fee-output fix, with a regression test confirming validate() still preserves the existing oracle-price error for closes. Closing this master-targeted PR as superseded.

PR #3342, opened 2026-06-09
Fix balance asset sorting with subasset longnames

### LINKED ISSUE #3021: Sorting Balances on the API
I tried to coalesce the fields here in #2209, but the asset_longname column didn't exist at the time so it didn't work and was reverted.

Now in #2786 we have asset_longname on the table, but if you sort by that you don't get the coalesced experience I was going for and think is the ideal way to sort balances in a familiar way. The null asset_longname values for named assets makes sorting by asset_longname alone insufficient. I think my failed PR could probably work now.

caydyan:
Fixes #3021.\n\nThis restores the intended coalesced asset sorting for balances now that the API balances table includes asset_longname. When clients request sort=asset on balances, the query orders by COALESCE(asset_longname, asset), so named assets and subassets sort by the user-visible asset name. The special ordering is scoped to the balances table so other tables that support sort=asset but do not have asset_longname keep using their asset column.\n\nTests:\n- .venv/bin/ruff format counterpartycore/lib/api/queries.py counterpartycore/test/units/api/queries_test.py\n- .venv/bin/ruff check counterpartycore/lib/api/queries.py counterpartycore/test/units/api/queries_test.py\n- .venv/bin/pytest counterpartycore/test/units/api/queries_test.py\n- git diff --check\n\nBTC address for any applicable bounty: bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/queries.py
+++ b/counterparty-core/counterpartycore/lib/api/queries.py
@@ -324,7 +324,9 @@ def select_rows(
                 sort_order = "ASC"
             if sort_order.upper() not in ["ASC", "DESC"]:
                 sort_order = "ASC"
-            if sort_name in SUPPORTED_SORT_FIELDS.get(table, []):
+            if table == "balances" and sort_name == "asset":
+                order_by.append(f"COALESCE(asset_longname, asset) {sort_order.upper()}")
+            elif sort_name in SUPPORTED_SORT_FIELDS.get(table, []):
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3432: https://github.com/CounterpartyXCP/counterparty-core/pull/3432

PR #3346, opened 2026-06-09
Encode MPMA text memo lengths as UTF-8 bytes

### LINKED ISSUE #1248: Special character on mpma memo is giving errors when parsing a tx
When creating a mpma and inserting a memo with a special character like "∞", the txs won't get parsed because of an utf-8 decoding error. The tx [11d105d3bb5b9a2ccce7c5c5b86a2b5333edc8f5dd7286e6acef1d4bf82e5fdf](https://blockstream.info/tx/11d105d3bb5b9a2ccce7c5c5b86a2b5333edc8f5dd7286e6acef1d4bf82e5fdf) didn't get parsed because of that error.

Right now it's not clear if there are more special characters giving the same error. Also, not sure if it's an encoding or decoding problem.

I did some tests on testnet putting that special character on a mpma and got an error too. When i removed it from the mpma it was parsed without problems, so that character is surely causing problems.

caydyan:
Fixes #1248.

MPMA text memo encoding currently stores `len(memo)` in the 6-bit memo length field, but writes `memo.encode("utf-8")` into the payload. For non-ASCII memo text such as `∞`, the encoded byte length is larger than the character count, so decoding reads a truncated UTF-8 sequence and raises a UnicodeDecodeError.

This PR encodes the text memo first, then writes the encoded byte length before appending the bytes. Hex memo handling is unchanged.

Added a regression test that composes an MPMA with `∞` as the message-wide memo and verifies it unpacks back to the same text memo for each send.

Tests run:
- `.venv/bin/pytest counterpartycore/test/units/messages/versions/mpma_test.py::test_compose_roundtrips_utf8_memo -q`
- `.venv/bin/pytest counterpartycore/test/units/messages/versions/mpma_test.py -q`
- `.venv/bin/ruff format --check counterpartycore/lib/utils/mpmaencoding.py counterpartycore/test/units/messages/versions/mpma_test.py`
- `.venv/bin/ruff check counterpartycore/lib/utils/mpmaencoding.py counterpartycore/test/units/messages/versions/mpma_test.py`
- `git diff --check`

BTC address for bounty consideration: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

### DIFF
--- a/counterparty-core/counterpartycore/lib/utils/mpmaencoding.py
+++ b/counterparty-core/counterpartycore/lib/utils/mpmaencoding.py
@@ -56,9 +56,10 @@ def _encode_memo(memo=None, is_hex=False):
             barr.append(memo)
         else:
             # signal a 0 bit for a string encoded memo
+            memo = memo.encode("utf-8")
             barr.append("0b0")
             barr.append(f"uint:6={len(memo)}")
-            barr.append(BitArray(memo.encode("utf-8")))
+            barr.append(BitArray(memo))
 
         return barr
     # if the memo is None, return just a 0 bit
--- a/counterparty-core/counterpartycore/test/units/messages/versions/mpma_test.py
+++ b/counterparty-core/counterpartycore/test/units/messages/versions/mpma_test.py
@@ -410,6 +410,26 @@ def test_compose_valid(ledger_db, defaults):
     )
 
 
+def test_compose_roundtrips_utf8_memo(ledger_db, defaults):
+    _source, _destination, data = mpma.compose(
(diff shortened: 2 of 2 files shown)


caydyan:
Opened and updated the develop-based replacement here: https://github.com/CounterpartyXCP/counterparty-core/pull/3400

It carries the same MPMA UTF-8 memo length fix, now with full MPMA test-file verification and ruff/diff checks against current develop. Closing this master-targeted PR as superseded to keep review focused on the active branch.

Issue #2965, opened 2025-01-25
Slow DB query


ouziel-slama:
slow query reported by sentry

```
SELECT COUNT(*) AS count FROM (SELECT *, rowid AS rowid FROM assets_info WHERE (asset = ?) OR (UPPER(asset_longname) = ?)  )
```
from

https://api.counterparty.io:4000/v2/assets/A7025775933017093000

caydyan:
Submitted PR #3349 for this: https://github.com/CounterpartyXCP/counterparty-core/pull/3349

Bounty payment address (BTC): `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

caydyan:
Opened a `develop`-based replacement PR for this issue: https://github.com/CounterpartyXCP/counterparty-core/pull/3403

This carries the same focused single-row query performance fix as #3349, but targets `develop` so maintainers do not need to review the older `master`-based branch.

BTC payout address if this qualifies for a contributor/security reward: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

ouziel-slama:
fixed here https://github.com/CounterpartyXCP/counterparty-core/pull/3403

PR #3349, opened 2026-06-09
Skip count query for single-row selects

caydyan:
Addresses #2965.

Summary:
- Let `select_rows()` skip the extra `COUNT(*)` query when callers do not need `result_count`.
- Use that lighter path from `select_row()`, which covers single-record endpoints such as `/v2/assets/<asset>`.
- Preserve the default counted behavior for list/paginated endpoints.

Tests:
- `.venv/bin/pytest counterpartycore/test/units/api/queries_test.py::test_select_rows_can_skip_result_count counterpartycore/test/units/api/apiserver_test.py::test_new_get_asset_info counterpartycore/test/units/api/apiserver_test.py::test_get_asset_by_longname_case_insensitive -q`
- `.venv/bin/ruff check counterpartycore/lib/api/queries.py counterpartycore/test/units/api/queries_test.py`
- `git diff --check`

Bounty payment address (BTC): `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/queries.py
+++ b/counterparty-core/counterpartycore/lib/api/queries.py
@@ -195,6 +195,7 @@ def select_rows(
     order="DESC",
     wrap_where=None,
     sort=None,
+    with_count=True,
 ):
     if offset is not None or sort is not None:
         last_cursor = None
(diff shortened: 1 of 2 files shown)


caydyan:
Closed in favor of the `develop`-based replacement PR #3403: https://github.com/CounterpartyXCP/counterparty-core/pull/3403

PR #3370, opened 2026-06-09
Support message-only compose alias

### LINKED ISSUE #2185: Create `/pack` method
Just as we can pass a datahex to `/unpack` it'd be great to have a method to which we can pass a Counterparty message and return the hex-encoded message.

caydyan:
Fixes #2185.

This adds `message_only` as a public compose parameter alias for the existing `return_only_data` behavior. When set, compose calls return only the packed Counterparty message data instead of constructing a full Bitcoin transaction, matching the message-only/pack use case discussed in the issue while keeping the existing `return_only_data` parameter unchanged.

Verification:
- `.venv/bin/pytest counterpartycore/test/units/api/composer_test.py::test_prepare_construct_params counterpartycore/test/units/api/composer_test.py::test_compose_transaction -q`
- `.venv/bin/pytest counterpartycore/test/units/api/composer_test.py -q`
- `.venv/bin/ruff check counterpartycore/lib/api/composer.py counterpartycore/test/units/api/composer_test.py`
- `.venv/bin/ruff format --check counterpartycore/lib/api/composer.py counterpartycore/test/units/api/composer_test.py`
- `git diff --check`

Bounty payment address (BTC): `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/composer.py
+++ b/counterparty-core/counterpartycore/lib/api/composer.py
@@ -1257,6 +1257,7 @@ CONSTRUCT_PARAMS = {
         "Include additional information in the result including data and psbt",
     ),
     "return_only_data": (bool, False, "Return only the data part of the transaction"),
+    "message_only": (bool, False, "Alias for `return_only_data`"),
     "segwit_dust_size": (int, None, "The dust size for segwit outputs (default is 330)"),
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3416: https://github.com/CounterpartyXCP/counterparty-core/pull/3416

PR #3376, opened 2026-06-09
Keep latest holder rows by consolidation key

caydyan:
## Summary

This fixes holder consolidation in `ledger.supplies._get_holders()`.

The function builds `record_id` from the configured id fields, but the membership check accidentally used Python's built-in `id` instead of `record_id`. As a result, duplicate holder ids were overwritten by cursor order before the rowid comparison could run. If rows are not returned in rowid order, holder calculations can keep an older balance or escrow row instead of the latest row.

The fix uses the computed `record_id` key so duplicate holder records are compared by rowid as intended.

## Why it matters

`holders()` feeds dividend validation/distribution, holder counts, and supply/held consistency checks. Keeping stale holder rows can under- or over-count balances when query order differs from latest row order.

## Tests

- `.venv/bin/pytest counterpartycore/test/units/ledger/supplies_test.py::test_get_holders_keeps_highest_rowid_for_duplicate_id -q`
- `.venv/bin/pytest counterpartycore/test/units/ledger/supplies_test.py -q`
- `.venv/bin/ruff check counterpartycore/lib/ledger/supplies.py counterpartycore/test/units/ledger/supplies_test.py`
- `.venv/bin/ruff format --check counterpartycore/lib/ledger/supplies.py counterpartycore/test/units/ledger/supplies_test.py`
- `git diff --check`

### DIFF
--- a/counterparty-core/counterpartycore/lib/ledger/supplies.py
+++ b/counterparty-core/counterpartycore/lib/ledger/supplies.py
@@ -9,7 +9,7 @@ def _get_holders(cursor, id_fields, hold_fields_1, exclude_empty_holders=False):
     save_records = {}
     for record in cursor:
         record_id = " ".join([str(record[field]) for field in id_fields])
-        if id not in save_records:
+        if record_id not in save_records:
             save_records[record_id] = record
             continue
         if save_records[record_id]["rowid"] < record["rowid"]:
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3418: https://github.com/CounterpartyXCP/counterparty-core/pull/3418

PR #3378, opened 2026-06-09
Validate v2 API boolean parameters

caydyan:
## Summary

This validates v2 API boolean query parameters instead of silently coercing unsupported values to `false`.

Currently, `prepare_args()` parses booleans with `str_arg.lower() in ["true", "1"]`, so any unsupported value becomes `False`. For example:

- `/v2/transactions?show_unconfirmed=maybe` returns `200 OK`
- `/v2/transactions?verbose=maybe` is accepted and treated as non-verbose

The fix accepts only `true`, `1`, `false`, and `0` for boolean route arguments and returns the existing `400` API error path for unsupported values. It also validates the shared non-compose `verbose` query parameter before executing the route.

## Tests

- `.venv/bin/pytest counterpartycore/test/units/api/apiserver_test.py::test_invalid_bool_param_rejected -q`
- `.venv/bin/pytest counterpartycore/test/units/api/apiserver_test.py -q`
- `.venv/bin/ruff check counterpartycore/lib/api/apiserver.py counterpartycore/test/units/api/apiserver_test.py`
- `.venv/bin/ruff format --check counterpartycore/lib/api/apiserver.py counterpartycore/test/units/api/apiserver_test.py`
- `git diff --check`

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/apiserver.py
+++ b/counterparty-core/counterpartycore/lib/api/apiserver.py
@@ -132,6 +132,15 @@ def set_cors_headers(response):
         response.headers["Access-Control-Allow-Methods"] = "*"
 
 
+def parse_bool_arg(arg_name, str_arg):
+    value = str_arg.lower()
+    if value in ["true", "1"]:
+        return True
+    if value in ["false", "0"]:
+        return False
+    raise ValueError(f"Invalid boolean: {arg_name}")
+
+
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3420: https://github.com/CounterpartyXCP/counterparty-core/pull/3420

PR #3380, opened 2026-06-09
Validate inputs_set values

caydyan:
## Summary

`prepare_inputs_set()` accepted explicit UTXO values outside Bitcoin's valid output amount range. A caller could pass negative values such as `-1`, or values above the 21M BTC money supply limit, and those values were stored in the composed UTXO list.

This PR validates explicit `inputs_set` values before they are used for transaction input selection and change calculation.

## Impact

Invalid caller-provided UTXO values can distort compose-time BTC input totals, fee/change calculation, and later transaction-building behavior. The API should return a clear `ComposeError` instead of carrying impossible UTXO amounts deeper into the composer.

## Changes

- Add `MAX_BTC_OUTPUT_VALUE = 21_000_000 * config.UNIT`.
- Reject `inputs_set` values below 0 or above the BTC output limit.
- Add regression coverage for negative and too-large explicit UTXO values.

## Tests

- `python -m pytest counterpartycore/test/units/api/composer_test.py::test_prepare_inputs_set`

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/composer.py
+++ b/counterparty-core/counterpartycore/lib/api/composer.py
@@ -35,6 +35,7 @@ from counterpartycore.lib.parser import deserialize, messagetype, utxosinfo
 from counterpartycore.lib.utils import helpers, multisig, script
 
 MAX_INPUTS_SET = 100
+MAX_BTC_OUTPUT_VALUE = 21_000_000 * config.UNIT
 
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3422: https://github.com/CounterpartyXCP/counterparty-core/pull/3422

PR #3382, opened 2026-06-09
Validate dust size parameters

caydyan:
## Summary

Deprecated but still-active compose parameters `regular_dust_size` and `multisig_dust_size` accepted negative values. Those values are used directly when creating default destination/change/data outputs, so a negative dust size can become an invalid transaction output amount.

This PR rejects negative dust size overrides with a clear `ComposeError`.

## Impact

Invalid caller-provided dust sizes can flow into `TxOutput` creation and transaction sanity checks instead of failing at the parameter boundary. This can produce confusing compose behavior and malformed transaction output objects.

## Changes

- Reject negative `regular_dust_size`.
- Reject negative `multisig_dust_size`.
- Add regression coverage for both parameters while preserving the existing default and positive override behavior.

## Tests

- `python -m pytest counterpartycore/test/units/api/composer_test.py::test_dust_size`

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/composer.py
+++ b/counterparty-core/counterpartycore/lib/api/composer.py
@@ -123,12 +123,16 @@ def create_tx_output(value, address_or_script, unspent_list, construct_params):
 
 def regular_dust_size(construct_params):
     if construct_params.get("regular_dust_size") is not None:
+        if construct_params["regular_dust_size"] < 0:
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3424: https://github.com/CounterpartyXCP/counterparty-core/pull/3424

PR #3384, opened 2026-06-09
Validate confirmation target

caydyan:
## Summary

`prepare_fee_parameters()` accepted any `confirmation_target` value before passing it to Bitcoin Core fee estimation. The backend helper documents the valid range as 1 to 1008 blocks, but compose-time callers could pass 0, negative values, or values above 1008 and let the error surface later at the RPC layer.

This PR validates the confirmation target before calling `estimatesmartfee`.

## Impact

Invalid caller-provided confirmation targets can produce unclear compose failures from Bitcoin Core instead of a direct `ComposeError` at the parameter boundary.

## Changes

- Add `MAX_CONFIRMATION_TARGET = 1008`.
- Reject `confirmation_target` values below 1 or above 1008 when fee estimation will use the parameter.
- Add regression coverage for 0, negative, and above-range values.

## Tests

- `python -m pytest counterpartycore/test/units/api/composer_test.py::test_prepare_fee_parameters`
- `git diff --check`

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/composer.py
+++ b/counterparty-core/counterpartycore/lib/api/composer.py
@@ -35,6 +35,7 @@ from counterpartycore.lib.parser import deserialize, messagetype, utxosinfo
 from counterpartycore.lib.utils import helpers, multisig, script
 
 MAX_INPUTS_SET = 100
+MAX_CONFIRMATION_TARGET = 1008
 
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3426: https://github.com/CounterpartyXCP/counterparty-core/pull/3426

PR #3386, opened 2026-06-09
Reject duplicate inputs_set UTXOs

caydyan:
## Summary

`prepare_inputs_set()` accepted duplicate UTXOs in the caller-provided `inputs_set` list. The duplicated entries are returned as separate inputs, so later input selection can count the same UTXO value more than once and build a transaction with duplicate outpoints.

This PR rejects duplicate UTXOs while parsing `inputs_set`.

## Impact

Duplicate explicit inputs can distort compose-time BTC input totals and produce an invalid transaction input set. The composer should fail at the parameter boundary instead of carrying duplicated outpoints deeper into transaction construction.

## Changes

- Track seen `<txid>:<vout>` identifiers while parsing `inputs_set`.
- Reject repeated UTXOs with a clear `ComposeError`.
- Add regression coverage for duplicate explicit inputs.

## Tests

- `python -m pytest counterpartycore/test/units/api/composer_test.py::test_prepare_inputs_set`
- `git diff --check`

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/composer.py
+++ b/counterparty-core/counterpartycore/lib/api/composer.py
@@ -619,6 +619,7 @@ def complete_unspent_list(unspent_list):
 def prepare_inputs_set(inputs_set):
     unspent_list = []
     utxos_list = inputs_set.split(",")
+    seen_utxos = set()
     if len(utxos_list) > MAX_INPUTS_SET:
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3428: https://github.com/CounterpartyXCP/counterparty-core/pull/3428

PR #3388, opened 2026-06-09
Validate fairmint quantity before comparisons

caydyan:
## Summary

`fairmint.validate()` recorded `quantity must be an integer` for non-integer quantities, but then continued into fairminter-specific arithmetic and comparisons when the fairminter existed. For a paid fairminter, a value like `"10"` could reach `quantity <= 0` and raise a Python `TypeError` instead of returning the compose-level validation error.

This PR stops fairmint validation before any quantity arithmetic when the quantity is not an integer.

## Impact

Invalid caller-provided fairmint quantities can leak lower-level Python exceptions on otherwise valid fairminter assets. The compose path should return a clear `ComposeError` from validation.

## Changes

- Track invalid non-integer fairmint quantity before looking up the fairminter.
- Preserve the existing `fairminter not found` behavior when the asset is missing.
- Return the recorded quantity validation problem before arithmetic/comparison logic for existing fairminters.
- Add regression coverage for validate and compose paths.

## Tests

- `python -m pytest counterpartycore/test/units/messages/fairmint_test.py::test_validate counterpartycore/test/units/messages/fairmint_test.py::test_compose`
- `python -m pytest counterpartycore/test/units/messages/fairmint_test.py`
- `python -m ruff check counterpartycore/lib/messages/fairmint.py counterpartycore/test/units/messages/fairmint_test.py`
- `python -m ruff format --check counterpartycore/lib/messages/fairmint.py counterpartycore/test/units/messages/fairmint_test.py`
- `git diff --check`

## Bounty note

If this fix qualifies for a bounty, please send it to:

### DIFF
--- a/counterparty-core/counterpartycore/lib/messages/fairmint.py
+++ b/counterparty-core/counterpartycore/lib/messages/fairmint.py
@@ -26,12 +26,16 @@ def validate(
 ):
     problems = []
 
-    if not isinstance(quantity, int):
+    invalid_quantity = not isinstance(quantity, int)
+    if invalid_quantity:
         problems.append("quantity must be an integer")
 
     fairminter = ledger.issuances.get_fairminter_by_asset(db, asset)
     if not fairminter:
         problems.append(f"fairminter not found for asset: `{asset}`")
+        return problems
+
+    if invalid_quantity:
         return problems
 
     if fairminter["status"] != "open":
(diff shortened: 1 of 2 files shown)


caydyan:
Closing this master-based PR in favor of the develop-based replacement #3430: https://github.com/CounterpartyXCP/counterparty-core/pull/3430

Issue #3396, opened 2026-06-09
counterparty-server resident memory reaches ~12 GiB on mainnet (v11.1.0)


adamkrellenstein:
## Environment

- counterparty-core **v11.1.0** (`counterparty/counterparty:v11.1.0`), fully caught up on mainnet
- Two replicas on GKE (Kubernetes), each with bitcoind v30 + electrs v3.3.0 backends
- Relevant flags: `--wsgi-server=waitress --waitress-threads=20 --memory-profile --json-log`

## Observed memory usage

30-day peak resident memory (Cloud Monitoring, non-evictable, through 2026-06-09):

| Pod | Peak RSS | Current RSS | Uptime at measurement |
|---|---|---|---|
| counterparty-0 | 11.6 GiB | 11.8 GiB | ~2 days |
| counterparty-1 | 8.0 GiB | 8.1 GiB | ~3 days |

Both pods reach this within a couple of days of a restart (0 container restarts), and the younger pod is the larger one, so this looks like growth to a large steady-state working set under API load rather than a slow uptime-correlated leak. For comparison, the same version on signet peaks at ~2.1 GiB and on testnet4 at ~0.4 GiB.

~12 GiB for an index whose database is ~35 GB on disk seems much higher than the workload should need, and it is expensive to provision for: we just had to raise our Kubernetes memory requests/limits to 12/16 GiB per replica to stop flirting with OOM kills (UnspendableLabs/Infrastructure#243).

## Prior art

Earlier reports were closed: #783 ("Unbounded RAM usage?"), #1929 ("Testnet Catchup Memory Usage"), #698. #1975 ("In-Memory DB" mode) is open but goes the other way (spend RAM for catch-up speed). There doesn't appear to be an open issue tracking steady-state memory reduction, hence this one.

`--memory-profile` is enabled on these pods — happy to provide profile output, tracemalloc snapshots, or any other diagnostics that would help pin down where the resident set goes.

adamkrellenstein:
Diagnostic follow-up — per-process breakdown of the ~11.8 GiB pod (mainnet, v11.1.0, `--memory-profile` enabled):

| Component | Size | Notes |
|---|---|---|
| OS page cache (SQLite files) | ~4.6 GiB | Counted in the k8s working-set metric but kernel-reclaimable; not process memory |
[...]

PR #3444, opened 2026-06-23
perf: cache the enriched API response, not the raw QueryResult

adamkrellenstein:
## Summary

Caches the **enriched, ready-to-serve** API response instead of the raw `QueryResult`, so a cache *hit* skips per-row verbose enrichment (and its DB lookups) that previously re-ran on **every** request. This collapses the repeated-request cost that dominates the heavy-endpoint latency tail. No change to API output or status codes — purely where the work happens.

## Problem

`BLOCK_CACHE` stored the raw `QueryResult`. The `QueryResult` unwrap + `verbose.inject_details` / `clean_api_result` ran in `handle_route` on every request, **including cache hits**. `inject_details` is per-row and issues its own DB lookups, so a hot, repeated key (e.g. a verbose listing for a confirmed block) paid the full enrichment cost on every hit.

This shows up in real production latency (per-pod request logs, live gateway traffic): p50 is ~3 ms (cheap/point lookups), but heavy endpoints carry the tail — e.g. p95 ~65–116 ms, p99 ~330–690 ms, and verbose `limit=1000` requests up to ~1.7 s. Those repeated heavy keys re-enrich every time.

## Change

- On a cache **miss**, `execute_api_function` returns an `UncachedResult` (raw result + cache key + cachability); `handle_route` enriches it once via `enrich_result()` and stores the enriched `CachedResponse`.
- On a cache **hit**, `execute_api_function` returns the stored `CachedResponse` directly — no enrichment, no DB lookups, no re-unwrap.
- `verbose` is part of `request.url` and therefore of the cache key, so verbose / non-verbose forms never collide.

## Behavior preserved (this is a pure perf change)

- **Output is byte-identical**: the same `return_result` serializes the same enriched payload. Added regression test `test_cache_hit_returns_identical_body` asserts a cache hit returns a byte-identical body to the miss, for both verbose and non-verbose. The repo's regtest `apicache.json` golden test + `verbose_test.py` are the success-path guard.
- **Error semantics unchanged**: enrichment still runs in `handle_route` *outside* the function-call `try`, so an enrichment error still surfaces as **500** and a function-execution error as **400/503**, exactly as before (covered by the existing `test_sentry_context_*` tests). Only the *location* of the work moved.

## Relationship to other PRs

- Independent of the `--api-cache-size` PR (#3443), though both touch `execute_api_function`'s insert/evict region (trivial merge).
- Changes *what* is cached (enriched payload, larger than raw rows), so a follow-up row/byte-budget bound should be sized against these entries.

Part of #3396 (latency side).


adamkrellenstein:
**Empirical mainnet A/B: the aggregate latency win does not hold — converting to draft.**

Validated on a live mainnet replica (`v11.1.0` + this change) vs an unmodified `v11.1.0` control, both serving ~50/50 real gateway traffic at the same 1000-entry cap.

- An initial window showed a large p95/mean improvement, but that was a **traffic-mix artifact** of the canary pod's post-restart NEG ramp (it briefly received a cheaper, mostly-non-verbose mix). In a **fresh balanced window** (matched verbose share, ~equal n), overall **p95 ≈ 58 vs 52, mean ≈ 16.7 vs 18.1 (this change slightly _higher_), p99 unchanged-to-slightly-worse** — bootstrap CIs span zero. Not significant; the headline did not replicate.
- What *is* real: a controlled equal-request hit test shows ~78 ms mean / ~158 ms p95 saved on **heavy `verbose` endpoints** (the change genuinely skips `inject_details` on cache hits). But those are a small share of real traffic, so the saving washes out of aggregate p90/p95/mean.
- Caching the enriched payload also makes entries larger than the raw `QueryResult` — a small **memory increase** for no measured aggregate latency gain.

The change is correct (output byte-identical, CI green) but **not justified as a latency optimization by production data**. Drafting until there's a validated win — most likely the deferred variant that caches the *serialized bytes* (serialization, not enrichment, dominates heavy responses) — together with an accepted memory cost.

adamkrellenstein:
the above was written by an AI. I think this is a good change

PR #3445, opened 2026-06-23
perf: bound the API response cache by total rows (--api-cache-max-rows)

adamkrellenstein:
## Summary

Bounds the API response cache (`BLOCK_CACHE`) by **total rows** in addition to entry count, via `--api-cache-max-rows` (`config.API_CACHE_MAX_ROWS`, default `50000`; `0` disables the row bound). This lets the cache hold **many small entries** (high hit rate → low tail latency) while keeping its memory bounded.

> Stacked on #3444 (cache the enriched payload). Review/merge that first. Composes with #3443 (`--api-cache-size`).

## Why

An entry-count cap alone forces a memory ↔ hit-rate tradeoff: a small cap bounds memory but lowers the hit rate and **raises the tail latency** (we measured p95/p99 roughly doubling at `--api-cache-size 100` on real traffic). Most cached responses are tiny (real p50 ≈ 3 ms), so a **row budget** caps memory (driven by the few large result sets) while letting thousands of small entries stay cached — bounded memory *without* sacrificing hit rate. With #3443 you can then raise `--api-cache-size` for a bigger cache, and this row budget keeps its memory in check.

## How

- `blockcache.cache_insert(key, value, max_entries, max_rows)` is the single insert path: it maintains a running `BLOCK_CACHE_ROWS` total + per-key `BLOCK_CACHE_SIZES`, and FIFO-evicts (`popitem(last=False)`) until within **both** the entry-count cap and the row budget. `MAX_BLOCK_CACHE_SIZE` remains an entry-count backstop.
- `estimate_rows()` is an **O(1)** proxy (length of the cached result list). True byte sizing would be `O(rows × cols)` on every cache miss — adding latency to the path we want cheap; the entry-count backstop bounds memory if the proxy ever undercounts.
- `--api-cache-max-rows` flag; negatives clamp to 0. Profiler reports `BLOCK_CACHE_rows`.
- Unit tests: `estimate_rows` mapping, row-budget FIFO eviction, entry-cap backstop, overwrite-no-double-count, running-total invariant.

## Concurrency note

`BLOCK_CACHE` is mutated lock-free under ~20 waitress threads (the existing model). The `BLOCK_CACHE_ROWS += rows` read-modify-write is therefore approximate under concurrency, but the drift is bounded and safe: over-count evicts slightly early (more conservative memory), under-count is backstopped by `MAX_BLOCK_CACHE_SIZE`, and the total self-heals (clamped ≥ 0). No correctness impact on responses; only the budget is approximate. Adding a lock would serialize all workers on every cache write for no correctness benefit, so it is intentionally avoided.

Part of #3396 (memory/latency).


adamkrellenstein:
Stacked on #3444, which is now **draft** (its latency premise didn't hold in a balanced production A/B). Drafting this too. The row-budget bound itself is a sound memory-safety mechanism and could be rebased onto #3443's configurable entry cap independently if we want it without #3444.

PR #3458, opened 2026-07-15
fix: include every output in transaction fee

droplister:
## Problem

The Rust transaction parser stops walking outputs at the first ordinary output after Counterparty data. That output is usually change, but valid Bitcoin transactions can contain additional outputs after it. Those later outputs are then omitted from the miner-fee calculation.

For example, transaction `1db7a85e9bbbcd9f60a62411e94f1ae8d3851642d0e3ca73e095d522bf234293` is reported as paying 19,388,665 sats, while its inputs minus all outputs is 46,970 sats. The reported value is exactly the input total minus only the outputs visited before the parser breaks.

## Change

- Once Counterparty output parsing is complete, continue walking the remaining Bitcoin outputs solely to subtract their values from the fee.
- Do not interpret those later outputs as Counterparty destinations or data.
- Gate the corrected behavior behind a protocol-change height because `fee` participates in the transaction-list consensus hash.
- Add a regression using an existing transaction fixture with a later 5,324,538-sat output.

Mainnet/testnet activation heights intentionally remain placeholders for release coordination; regtest and signet enable the behavior immediately.

## Validation

- `rustfmt --check` passes for both modified Rust files.
- `git diff --check` passes.
- The local Windows Python suite is blocked before collection by python-bitcoinlib/OpenSSL loading; the added parser regression is intended to run in Linux CI.
- A full cold `cargo check` exceeded the local four-minute command window while compiling dependencies; no compiler error was emitted.


ouziel-slama:
moved here https://github.com/CounterpartyXCP/counterparty-core/pull/3470

PR #3475, opened 2026-08-09
Normalize max_mint_per_address and pool_quantity

### LINKED ISSUE #3474: verbose=true never normalizes max_mint_per_address or pool_quantity on fairminters
# `verbose=true` never normalizes `max_mint_per_address` or `pool_quantity`

Every other quantity on a fairminter gets a `_normalized` companion under `verbose=true`. These two
never do, on any fairminter, because neither is registered in the `quantity_fields` map in
`counterpartycore/lib/api/verbose.py`.

## Evidence

`GET /v2/fairminters?limit=100&verbose=true` against `api.counterparty.io`, 100 rows:

| Field | Rows carrying `_normalized` |
|---|---|
| `price`, `quantity_by_price`, `hard_cap`, `soft_cap`, `max_mint_per_tx`, `premint_quantity`, `earned_quantity`, `paid_quantity`, `commission` | 97 / 97 |
| **`max_mint_per_address`** | **0 / 100** |
| **`pool_quantity`** | **0 / 100** |

A single fairminter, for a concrete example — `GET /v2/assets/LAUNCHCOIN/fairminters?verbose=true`:

```json
"max_mint_per_tx": 100000000000000,
"max_mint_per_tx_normalized": "1000000.00000000",
"max_mint_per_address": 100000000000000,     // no companion
"pool_quantity": 3100000000000000,           // no companion
"soft_cap": 6900000000000000,
"soft_cap_normalized": "69000000.00000000"
```

## Cause

`inject_normalized_quantities` in `counterpartycore/lib/api/verbose.py` normalizes exactly the
fields listed in its `quantity_fields` dict. The fairminter group there is:

```python
"price":               {"asset_field": "asset_info", "divisible": None},
"hard_cap":            {"asset_field": "asset_info", "divisible": None},
"soft_cap":            {"asset_field": "asset_info", "divisible": None},
"quantity_by_price":   {"asset_field": "asset_info", "divisible": None},
"max_mint_per_tx":     {"asset_field": "asset_info", "divisible": None},   # <- registered
"premint_quantity":    {"asset_field": "asset_info", "divisible": None},
"earned_quantity":     {"asset_field": "asset_info", "divisible": None},
"earn_quantity":       {"asset_field": "asset_info", "divisible": None},
"commission":          {"asset_field": "asset_info", "divisible": None},
"paid_quantity":       {"asset_field": "asset_info", "divisible": None},
```

`max_mint_per_address` sits one line away from `max_mint_per_tx` and is simply absent, as is
`pool_quantity`. This reads as an oversight rather than a decision: both are asset-denominated
quantities of exactly the same kind as their neighbours, and `max_mint_per_address` is
consensus-enforced in `messages/fairmint.py` (`validate` rejects a mint that would push
`already_minted + quantity` past it).

## Why it matters to consumers

The asymmetry is the problem, not the absence. A client reading fairminter limits reasonably writes

```js
if (fairminter.max_mint_per_address_normalized) applyPerAddressCap(...)
```

by symmetry with `max_mint_per_tx_normalized` — and that branch is then dead on every fairminter
ever created. It fails silently and in the permissive direction: the cap is skipped, a "max
mintable" figure comes out too high, and the resulting `fairmint` is rejected at consensus.

That is not hypothetical — it is the defect this issue came out of, in XCP Wallet's fairmint form.
It is also easy to get wrong the other way: normalizing by hand means dividing by 1e8 *only* for a
divisible asset, which is exactly the branch `inject_normalized_quantity` already owns.

Clients can and should work around it, and we have. But every client has to independently discover
that two of eleven quantities behave differently from the rest.

## Suggested fix

Two lines in `quantity_fields`. PR to follow.

I checked `apicache.json` is generated by `test/integrations/regtest/genapidoc.py` rather than
hand-maintained, so it would want regenerating alongside — I have not run the regtest suite, so
that part is untested on my side.

droplister:
# Normalize `max_mint_per_address` and `pool_quantity`

Closes #3474.

Two fairminter quantities are never given a `_normalized` companion under `verbose=true`, because
neither is registered in the `quantity_fields` map that `inject_normalized_quantities` iterates.
Every other quantity on the same row is. Across 100 fairminters from
`GET /v2/fairminters?limit=100&verbose=true`, `max_mint_per_address_normalized` and
`pool_quantity_normalized` appear **0** times, while `max_mint_per_tx_normalized` appears on all 97
rows that carry `max_mint_per_tx`.

```diff
         "max_mint_per_tx": {"asset_field": "asset_info", "divisible": None},
+        "max_mint_per_address": {"asset_field": "asset_info", "divisible": None},
         "premint_quantity": {"asset_field": "asset_info", "divisible": None},
         "earned_quantity": {"asset_field": "asset_info", "divisible": None},
         "earn_quantity": {"asset_field": "asset_info", "divisible": None},
         "commission": {"asset_field": "asset_info", "divisible": None},
         "paid_quantity": {"asset_field": "asset_info", "divisible": None},
+        "pool_quantity": {"asset_field": "asset_info", "divisible": None},
```

Both are asset-denominated quantities of exactly the same kind as their neighbours, so they take the
same `{"asset_field": "asset_info", "divisible": None}` treatment — divisibility comes from the
fairminter's own `asset_info`, which is what makes the 1e8 division conditional. `max_mint_per_tx`
is placed immediately above its per-address counterpart; `pool_quantity` joins the rest of the
fairminter group.

## Why this is worth fixing rather than documenting

The asymmetry is the hazard, not the absence. A client reading fairminter limits writes

```js
if (fairminter.max_mint_per_address_normalized) applyPerAddressCap(...)
```

by symmetry with the per-tx cap directly above it, and that branch is dead on every fairminter that
has ever existed. It fails silently and permissively: the cap is dropped, a "maximum mintable"
figure comes out too high, and the `fairmint` composed from it is rejected at consensus by
`messages/fairmint.py`, which does enforce `max_mint_per_address`. That is the defect this came out
of, in XCP Wallet.

Normalizing by hand is also the kind of thing worth not asking every client to do: it means dividing
by 1e8 *only* when the asset is divisible, which is the branch `inject_normalized_quantity` already
owns.

## Compatibility

Additive. Existing fields keep their values and no field is renamed or removed, so a client reading
only base units is unaffected. Clients already working around this by normalizing
`max_mint_per_address` themselves keep working — they just no longer need to.

## Testing

- Verified the patched module parses and the map is well-formed.
- **Not** run: the regtest integration suite. `test/integrations/regtest/apidoc/apicache.json`
  contains recorded API responses and is generated by `test/integrations/regtest/genapidoc.py`, so
  it will want regenerating for the new fields to appear in the API docs. I did not have a regtest
  environment to do that in — happy to follow up if you point me at the invocation, or for a
  maintainer to regenerate as part of the merge.
- The 100-row sample above is against mainnet `api.counterparty.io` and is what established the
  omission; it is evidence for the bug rather than a test of the fix.


ouziel-slama:
moved here https://github.com/CounterpartyXCP/counterparty-core/pull/3476

PR #3488, opened 2026-08-25
Fix slow API shutdown and cold startup

### LINKED ISSUE #3486: Force-killed API child causes multi-minute State DB recovery on restart
## Summary

Counterparty Core v11.3.0 repeatedly force-kills its API child during otherwise graceful Kubernetes termination. On the next start, mainnet then spends 2.5–6.8 minutes inside the State DB migration/check phase even though no migration is pending. This turned a routine GKE node upgrade into an approximately 8 minute 44 second public API outage.

This is distinct from, but compounds, the State DB reorg issue in #3485.

## Environment

- Counterparty Core: v11.3.0
- Current commit reported by the process: cd827e0fafe62b3166b2a186749e4191e1897170
- Image digest: sha256:e546a869f4ad0cf049e5480bad6b7246e790be6b310981297f78391f62ca44bd
- Network: Bitcoin mainnet
- State DB size: approximately 7.1 GiB
- Ledger DB size: approximately 16.9 GiB
- Deployment: singleton StatefulSet with a 180 second Kubernetes termination grace period

## Latest restart timeline

- 07:13:39Z: SIGTERM received by the main process
- 07:13:52Z: API Server process did not stop in time. Terminating forcefully...
- 07:14:24Z: replacement container starts
- 07:14:33Z: databases opened; ledger migrations checked
- 07:15:17Z: Applying migrations to /data/state.db...
- 07:22:07Z: Starting API Watcher thread / WSGI / health server
- 07:22:23Z: pod Ready

No State DB migration was actually applied. All 15 migration records were already present from August 5. On the healthy running database, migration-file loading takes approximately 8.5 ms and a read-only query of Yoyo metadata takes approximately 3 ms.

## Repeated correlation

Every retained mainnet restart shows the same forced API-child termination followed by a slow State DB startup phase:

- August 21: 3m02s
- August 23: 2m29s
- August 25: 6m50s

By comparison, the same State DB migration phase takes approximately 0.7 seconds on signet and 2.5–5.3 seconds on testnet4.

## Suspected cause

APIServer.stop() sends SIGTERM, waits a hard-coded 10 seconds, and then calls process.kill(). The child signal handler itself is a no-op. The following startup delay occurs within apply_outstanding_migration(), likely while opening/recovering/checkpointing the large SQLite WAL after the forced child exit. The current log line does not distinguish get_backend(), read_migrations(), to_apply(), apply_migrations(), or connection close.

## Proposed approach

1. Add elapsed-time and pending-count logs around each apply_outstanding_migration() phase.
2. Make API-child graceful-stop timeout configurable and substantially longer than 10 seconds.
3. Ensure APIWatcher, State DB connections, and WAL checkpoint/close finish before the child exits.
4. Log WAL/checkpoint state during shutdown and startup.
5. Test repeated stop/start cycles against a restored mainnet State DB snapshot and compare clean versus forced shutdown.

I can implement and test this against an isolated restored mainnet snapshot with no public route.

adamkrellenstein:
## Summary

- interrupt API watcher and node-status SQLite reads during shutdown so the API child exits inside the existing 10-second parent budget
- enforce one aggregate shutdown deadline across health, WSGI, watcher, and Gunicorn-worker cleanup, while retaining the bounded force-kill fallback
- preserve latest-event semantics while replacing the cold full-result `BLOCK_PARSED` sort with a reverse scan of `parsed_events_event_index_idx`
- add phase-level migration/WAL and API-watcher startup timing logs

Closes #3486.

## Why

During a GKE node upgrade, the mainnet API child did not exit within 10 seconds and was killed. Its next startup then spent several minutes initializing the API watcher, causing a real singleton outage.

Profiling an isolated disk cloned from the current mainnet snapshot showed:

- ledger migration check: 0.12s
- state migration check: 0.08s
- pre-fix `APIWatcher.__init__`: approximately 132.45s
- expensive query: latest `BLOCK_PARSED` ordered by `event_index`, planned as an event-index scan plus temporary B-tree sort

## Validation

- original full suite: `1604 passed`
- post-red-team focused shutdown/query suite: `120 passed`
- post-red-team full API unit suite: `463 passed`
- Ruff/pre-commit: passed
- deterministic local integration test uses the production Waitress backend and a genuinely blocked APSW query; shutdown interrupts it and completes inside a 2-second component deadline, with no external backend and no skip path
- isolated production-equivalent GKE test against a copied mainnet snapshot, with no Service or public route:
  - first patched start: API watcher initialized in 0.01s; ledger/state migration checks 0.10s/0.05s
  - clean SIGTERM: API child stopped in 1.22s; no force-kill
  - second patched start: API watcher initialized in 0.01s; ledger/state migration checks 0.07s/0.03s; catch-up completed immediately
  - both starts became Ready with zero container restarts

The final query preserves the original `event_index` ordering explicitly and has a regression assertion that SQLite uses `parsed_events_event_index_idx` without a temporary B-tree. No consensus code or production deployment configuration is changed by this PR.


adamkrellenstein:
CI note: the two `Audit` failures are unrelated to this PR. Running the exact workflow command locally reports `RUSTSEC-2026-0258` against `h2 0.4.15` in the unchanged `counterparty-client/Cargo.lock`; the advisory was published on 2026-08-17, after the current `develop` head last passed this workflow on 2026-08-16. The advisory recommends `h2 >= 0.4.16`.

This PR does not modify Counterparty Client dependencies or its lockfile. That dependency update should be handled separately rather than mixed into the API shutdown/startup fix.

PR #3489, opened 2026-08-25
Fix slow address history endpoints

### LINKED ISSUE #3482: Slow responses on 3 API endpoints
The below v2 API endpoints are often extremely slow and/or time out, even on addresses with short histories:

 https://api.counterparty.io:4000/v2/addresses/bc1qdnaw4cxefuplcw3a68sw9w4f4puua3dnz5kprh/credits?limit=10
 https://api.counterparty.io:4000/v2/addresses/bc1qdnaw4cxefuplcw3a68sw9w4f4puua3dnz5kprh/debits?limit=10
 https://api.counterparty.io:4000/v2/addresses/bc1qdnaw4cxefuplcw3a68sw9w4f4puua3dnz5kprh/sends?limit=10

adamkrellenstein:
## Summary

- resolve the credits, debits, and sends address-history OR branches through their existing indexes
- bound each rowid branch before UNION/deduplication so hot-address pages do not materialize the full history
- use SQLite native multi-index OR for exact counts instead of a second complete UNION
- preserve generic address resolution behavior outside the explicitly indexed paths
- cap address-history offsets at 10,000, reject arbitrary sorting on the two sends-by-address routes, and normalize/deduplicate CSV send types
- return exact result counts on initial/offset responses while skipping repeated full counts on later cursor pages

Closes #3482.

## Root cause and safety

SQLite could choose a reverse full-table scan for OR-shaped address predicates followed by `ORDER BY rowid DESC LIMIT`, scanning millions of unrelated rows even though every branch had an index. The normal rowid path now takes only the page-sized top slice from each branch, then UNION-deduplicates and applies the existing outer pagination semantics. Arbitrary global sorts cannot be bounded by that theorem, so the affected public routes reject them explicitly instead of silently recreating unbounded work.

## Validation

- 184 focused query tests and 250 post-rebase query/API tests pass
- full Python 3.11 unit suite before the final #3488 rebase: 1,622 passed
- synthetic 1M-row/all-four-leg hot case: row page 0.5036s -> 0.0001s; exact count 0.5242s -> 0.2031s
- coverage includes ASC/DESC cursors, offset cap, action/asset/send-type filters, duplicate branch matches, unknown addresses, hot credits/debits/sends, custom-sort rejection, CSV amplification, count suppression on later pages, and EXPLAIN plans
- Ruff, format, and commit hooks pass

This remains stacked on #3488 so its single commit is reviewed against the hardened shutdown/startup base.


adamkrellenstein:
Additional validation: the complete API unit suite passes (464 passed).

adamkrellenstein:
Final cross-PR validation after rebasing and squashing: a local integration tree containing #3488 + #3489 + #3490 + #3491 passed the complete Python 3.11 unit suite (**1,706 passed**) and the complete API unit suite (**557 passed**). Ruff/format/commit hooks also passed. The temporary combined branch was not pushed. A Python 3.13 run reached 1,702 passes with the same four pre-existing Rust-extension/global-config failures seen outside these changes; Python 3.11 is the repository known-good environment used for the clean result.

PR #3490, opened 2026-08-25
Reject pending issuance compose conflicts

adamkrellenstein:
## Summary

- reject issuance and fairminter composition when Counterparty parsed-mempool state already proves a conflicting asset creation, ownership/lock/reset transition, cumulative supply overflow, or active fairminter
- use real parser event bundles and fairminter `open`/`pending` statuses while allowing compatible reissuances and parameter-compatible zero-quantity operations
- recheck confirmed and parsed-mempool state after slow input selection; refreshed compose bytes are authoritative
- atomically reserve selected UTXOs during late validation and release only the caller reservation on failure
- return HTTP 409 for conflicts, handle malformed/non-object bindings safely, and preserve `validate=false`

Closes #3478.

## Scope

This is an API preflight guard, not consensus logic. It prevents compositions already known to conflict without making consensus depend on local mempool contents. Counterparty mempool parsing is asynchronous, so the parser batching window and the normal post-response sign/broadcast race remain explicitly best-effort. UTXO reservations retain the existing process-local scope; the current threaded API deployment gets atomic selection, while a future multi-process composition deployment would need shared locking.

## Validation

- focused compose/composer suite: 151 passed
- post-rebase compose/composer/API suite: 216 passed
- full Python 3.11 unit suite: 1,629 passed
- tests cover real creation/fairminter bundles, cumulative MAX_INT overflow, compatible reissuance, lock parameter sensitivity, malformed bindings, issuance/fairminter HTTP 409, authoritative subasset/full-transaction recomposition, reservation ownership/expiry, concurrent selection, and rejection cleanup
- Ruff, format, and commit hooks pass

This remains stacked on #3488 so its single commit is reviewed against the hardened shutdown/startup base.

PR #3491, opened 2026-08-25
Keep API available during State DB reorg rebuilds

### LINKED ISSUE #3485: State DB rollback after one-block reorg blocks API for ~33 minutes
## Summary

A one-block Bitcoin mainnet reorganization caused Counterparty Core v11.3.0 to spend approximately **32 minutes 51 seconds** rebuilding its State DB. The public API became unavailable for roughly the final six minutes of that operation, despite the process remaining alive and having substantial CPU and memory headroom.

## Environment

- Counterparty Core: `v11.3.0`
- Container image digest: `sha256:e546a869f4ad0cf049e5480bad6b7246e790be6b310981297f78391f62ca44bd`
- Network: Bitcoin mainnet
- Deployment: single Counterparty API instance on GKE
- Incident date: 2026-08-24 UTC

## Timeline

- `11:54:05.156Z`: `Blockchain reorganization detected at block 963854`
- Ledger DB rollback to block 963853 completed in approximately 38 seconds, after which block parsing resumed.
- `11:55:07.495Z`: API watcher detected the reorganization at block 963853.
- `11:55:07.500Z`: `Rolling back State DB to block index 963853...`
- Approximately `12:22–12:28Z`: the API's readiness/liveness endpoints timed out and public requests returned 5xx responses.
- `12:27:58.943Z`: `State DB rolled back in 1971.44 seconds`
- `12:28Z`: API readiness and public service recovered.

## Observed impact

During the approximately six-minute public degradation window:

- 2,387 5xx responses were recorded at the edge.
- Cacheable GETs could be served stale by an external cache, but uncached endpoints, mempool endpoints, and `/v2/healthz` failed.
- The Counterparty process did not restart and was not OOM-killed.
- Peak CPU was approximately 2.22 cores with an 8-core limit.
- Peak non-evictable memory was approximately 9.4 GiB with a 16 GiB limit.
- The WSGI worker pool did not appear saturated; reported busy workers remained low and the queue was generally empty.

This suggests the outage was caused by the State DB rollback/rebuild path rather than ordinary request load or resource exhaustion.

## Suspected expensive path

`counterpartycore/lib/api/dbbuilder.py::rollback_state_db` deletes rollbackable rows and then reapplies the migrations in `MIGRATIONS_AFTER_ROLLBACK`. That appears to reconstruct multiple materialized tables, views, and indexes even for a one-block reorganization.

## Expected behavior

A shallow reorganization should not block the public API for tens of minutes. Ideally one or more of the following would apply:

1. Roll back State DB changes incrementally for the affected block range.
2. Build a repaired State DB separately and atomically switch readers to it.
3. Continue serving the last internally consistent State DB snapshot while rebuilding.
4. Keep the dedicated health server responsive throughout recovery and expose a distinct degraded/rebuilding state.

We can provide additional sanitized logs or test a proposed patch if useful.

adamkrellenstein:
## Summary

- keep the existing internally consistent State DB available only through pre-reorg cached public GET hits while a replacement is built from a transactionally stable Ledger snapshot
- fail closed for cache misses, API metadata, compose, mempool, Bitcoin/live routes, authenticated requests, POST, and other fresh-state work
- keep the rebuild gate and age in parent-owned shared memory across Gunicorn workers, API-child crashes, and the deliberate activation restart; clear it only after lineage verification and catch-up
- validate a durable lineage/migration/builder-commit manifest, checkpoint/fsync the replacement, and activate it through ordered rename/fsync barriers
- cancel backup/migration/verification/vacuum work on shutdown; enforce a rebuild watchdog and restart the child after critical watcher failure
- reclaim `.previous` only when preflight proves its bytes are needed, clean every partial sidecar/journal, and rebuild address-event history in the fallback path
- detect latest same-height one-block reorgs and mid-block branch changes before appending events

Closes #3485.

## Failure safety

- incomplete, stale, cross-image, corrupt, or lineage-mismatched builds are never activated
- activation-resume validates a missing-staged/live-present state against the manifest before consuming the marker
- the parent restarts an unexpectedly dead child or failed critical watcher while preserving the gate
- a cache-cold replacement child, disabled cache, authenticated-only API, missing State DB, or over-age rebuild reports unready
- a failed staged build falls back while the API remains gated; migration 0001 is included so `address_events` cannot remain permanently incomplete
- no consensus code is changed

## Production-sized exercise

An isolated GKE pod with no Service/public route used the production 2 CPU / 12 GiB request and 8 CPU / 16 GiB limit on a 60 GB SSD cloned from the retained mainnet snapshot. The trigger was synthetic: one copied `BLOCK_PARSED` hash was altered; it was not a true two-fork reorg.

- 16.9 GB consistent Ledger snapshot: 2m27s
- State DB migration: 2,210.8s
- total staging: 2,398.8s (~40m)
- readiness probes during backup: 61–95ms
- old API child graceful stop: 2.90s
- full API-child switchover: 13.15s
- peak disk use: 88%, 7.1 GB free
- post-activation height: 964039 / 964039

Local tests now add real same-height replacement fixtures with fewer/equal/more events, a mid-block data-version change, stale-builder/lineage rejection, cancellation, repeated rebuild space, every activation fsync boundary, missing-staged recovery, fallback address-event completeness, shared-process gating, cold-cache readiness, and watcher/watchdog restart behavior.

## Validation

- full post-rebase API suite: 500 passed
- focused rebuild/shutdown/health/database suite: 228 passed
- combined #3488/#3489/#3490/#3491 API tree: 557 passed
- Ruff, format, and commit hooks pass

This remains stacked on #3488 because clean bounded shutdown and fast cold API initialization are prerequisites. It contains one additional commit.

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/apiserver.py
+++ b/counterparty-core/counterpartycore/lib/api/apiserver.py
@@ -57,6 +57,18 @@ def verify_password(username, password):
 
 
 def is_server_ready():
+    # A staged State DB rebuild deliberately serves the last internally
+    # consistent snapshot until the replacement is complete. Keep the
+    # singleton routable during that degraded-but-readable period.
+    if CurrentState().state_db_rebuilding():
+        return (
+            not CurrentState().state_db_rebuild_cache_cold()
+            and not config.DISABLE_API_CACHE
+            and config.API_PASSWORD is None
+            and CurrentState().state_db_rebuild_age()
+            <= config.DEFAULT_STATE_DB_REBUILD_MAX_READY_SECONDS
+        )
+
     backend_height = CurrentState().current_backend_height()
     block_index = CurrentState().current_block_index()
 
@@ -109,6 +121,24 @@ def is_cachable(rule, route=None, result=None):
     if request.path == "/v2/addresses/mempool":
         return False
     if "show_unconfirmed=true" in request.url:
+        return False
+    return True
+
+
+def can_serve_stale_during_state_db_rebuild(rule, route):
+    """Whether a request may use a response cached before rebuilding."""
+    if rule in {"/v2/healthz", "/healthz"}:
+        return True
+    if request.method != "GET" or config.API_PASSWORD is not None:
+        return False
+    path = request.path.lower()
+    if any(fragment in path for fragment in ("/compose/", "/mempool")):
+        return False
+    if request.args.get("show_unconfirmed", "false").lower() in {"true", "1"}:
+        return False
+    if route is not None and route["category"] in {"bitcoin", "compose", "mempool", "v1"}:
+        return False
+    if route is not None and route["function"].__module__.endswith(".compose"):
         return False
     return True
 
@@ -198,7 +228,13 @@ def return_result(
     response.headers["X-COUNTERPARTY-READY"] = is_server_ready()
     response.headers["X-COUNTERPARTY-VERSION"] = config.VERSION_STRING
     response.headers["X-BITCOIN-HEIGHT"] = CurrentState().current_backend_height()
-    response.headers["X-LEDGER-STATE"] = CurrentState().ledger_state()
+    rebuilding = CurrentState().state_db_rebuilding()
+    response.headers["X-LEDGER-STATE"] = (
+        "State DB Rebuilding" if rebuilding else CurrentState().ledger_state()
+    )
+    if rebuilding:
+        response.headers["X-COUNTERPARTY-STALE"] = "true"
+        response.headers["X-STATE-DB-REBUILD-AGE"] = str(int(CurrentState().state_db_rebuild_age()))
     response.headers["Content-Type"] = "application/json"
     set_cache_control_headers(response, http_code, result)
     set_cors_headers(response)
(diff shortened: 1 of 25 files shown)


adamkrellenstein:
Final cross-PR validation after rebasing and squashing: a local integration tree containing #3488 + #3489 + #3490 + #3491 passed the complete Python 3.11 unit suite (**1,706 passed**) and the complete API unit suite (**557 passed**). Ruff/format/commit hooks also passed. The temporary combined branch was not pushed. A Python 3.13 run reached 1,702 passes with the same four pre-existing Rust-extension/global-config failures seen outside these changes; Python 3.11 is the repository known-good environment used for the clean result.

ouziel-slama:
Closing: superseded by #3487, with the part that survives extracted into #3493.

This PR and #3487 answered #3485 from opposite ends. #3487 made rollbacks incremental, so a reorganization no longer triggers a full State DB rebuild at all, and it argued against the staged-rebuild approach here: once the rebuild only happens at startup, before any listener exists, there are no readers left to keep serving, and the atomic-swap machinery carries real risk for a case that no longer arises. With #3487 merged, everything built here — the Ledger snapshot, the lineage manifest, the cache gating, the rename/fsync activation, `.previous` reclamation, the rebuild watchdog and the activation restart — is reachable only through the fallbacks in `rollback_state_db()`: a rollback deeper than 1,000 blocks, a State DB predating that release, or an unexpected failure in the incremental path. For those, #3487 already sheds the pod with `503 rebuilding` rather than returning a storm of 5xx.

What did not overlap is three things that have nothing to do with the rebuild strategy, and they are now in #3493, rewritten against `develop`:

- `check_reorg()` comparing the block *before* the last one parsed, which hides a tip block replaced at the same height;
- the check running only when the watcher is idle and at most every five seconds, which misses a rollback landing mid-backlog or mid-block — keyed on `PRAGMA data_version` there instead;
- a watcher thread dying into `threading.excepthook` with its traceback going nowhere, while the process kept serving a frozen State DB — logged at `CRITICAL` and reported as `503 watcher_stopped` there.

The one piece deliberately left out of #3493 is the parent-side restart of a failed API child. Readiness sheds the pod; it does not bring the watcher back. If we want the automatic restart, it belongs in its own PR without the rebuild machinery attached to it.

PR #3499, opened 2026-09-04
v11.4.0

ouziel-slama:


john-a-zoidburg:
Independent review of the v11.4.0 release. This covers the non-security findings — correctness, packaging, lifecycle, and documentation items. Two security-relevant findings (a regression in the non-finite broadcast fix, and an uncovered backend-error boundary) are reported separately through the private security channel per SECURITY.md.

Credit: source-level analysis by OpenAI GPT-6 Astra Max; independent validation by Kimi K3; research coordination by John A Zoidburg.

## Summary

| # | Finding | Area | Issue |
|---|---|---|---|
| 1 | State refresh/rollback advances the replay cursor past missing address history | API correctness | #3503 |
| 2 | Health readiness can report ready before the public API finishes initialization | Lifecycle | #3504 |
| 3 | sdist cannot build a wheel (OpenAPI force-include outside the archive) | Packaging | #3505 |
| 4 | Compose image / integration-test identity mismatch | Release checklist | #3506 |
| 5 | OpenAPI missing 409 declarations for issuance/fairminter composition | Documentation | #3507 |

All findings were validated against the release head (`230302d30`) and master base (`67e10db3e`). Details, evidence, and proposed fixes are in the linked issues. A brief overview of each below.

---

### 1. State refresh/rollback cursor mismatch (API correctness)

The new automatic State refresh (and the full-rollback fallback) can advance the parsed-event replay cursor past the completion point of the `address_events` projection. `MIGRATIONS_AFTER_ROLLBACK` rebuilds migrations 0002–0016 but **not** `0001.create_and_populate_address_events`, while `ROLLBACKABLE_TABLES` deletes `address_events` and `0002` copies the parsed-event cursor to the Ledger tip. Catch-up then sees no work, and the API serves incomplete address history (`/v2/addresses/.../events`, `/v2/addresses/.../credits`) until a full State rebuild. Ledger consensus and balances are unaffected. See the linked issue for the full mechanism, evidence, and proposed fix.

### 2. Early health readiness (lifecycle)

The dedicated health listener now starts before State maintenance and WSGI construction, and the readiness predicate has no service-initialized condition — so it can report ready while WSGI construction is incomplete and the public-ready value is still 0. The API-only branch reports caught-up unconditionally. An orchestrator can route traffic to a pod that isn't yet serving. The base created the health object *after* WSGI construction, so this is a release-introduced exposure.

### 3. sdist → wheel packaging regression

`pyproject.toml:45-49` configures an unconditional wheel force-include from `../openapi.json`, a path valid in the repo layout but absent from a standalone sdist extraction. `pip wheel` on a fresh sdist fails with `FileNotFoundError: Forced include not found: .../openapi.json`. Reproduced with both the `build` and `pip` frontends; the master-base control passes. Release-blocking for the source-distribution channel.

### 4. Compose image / test identity (release checklist)

`docker-compose.yml` selects `counterparty/counterparty:v11.3.0` while `config.VERSION_STRING` is `11.4.0`. The integration test's search string matches nothing at this head, and its replacement would substitute `v11.0.4` — so the test doesn't verify the candidate executable. This is a pre-publication alignment item (the image tag is updated at release-prep time per the existing process).

### 5. OpenAPI missing 409 declarations (documentation)

`compose_issuance` and `compose_fairminter` call `_compose_with_pending_asset_guard`, which can return HTTP 409 on a parsed pending asset conflict. `openapi.json` has 197 operations and zero 409 declarations; the generator discards the observed example status. A documentation gap for schema-driven consumers.

---

Happy to provide the full evidence bundle (reproduction runners, validation notes) for any of these. The two security findings are being handled through the private channel per SECURITY.md.

adamkrellenstein (approved):

PR #3509, opened 2026-09-08
Address the external review of the v11.4.0 release candidate

### LINKED ISSUE #3503: State refresh and full rollback can advance the replay cursor past missing address history
# State refresh and full rollback can advance the replay cursor past missing address history

**Area:** API correctness (State DB projection)
**Affected release:** `230302d303b9c25ba01e45edeccba0d30a2c188f` (v11.4.0, PR #3499)
**Master comparison:** `67e10db3ee266068c1effc4e83653df39ace5ca8`

## Summary

When the State DB is behind the Ledger, the new automatic v11.4.0 refresh can mark unprocessed events as parsed without adding their address associations. The same projection mismatch occurs during full rollback fallback when the Ledger already contains a replacement suffix. Subsequent catch-up sees no work, so address history remains incomplete even while the node continues processing new events.

The underlying refresh/full-rebuild behavior is inherited; the release newly schedules the refresh automatically on applicable upgrades and retains full rebuild as the correctness fallback for incremental rollback. Both paths violate the same invariant: the replay cursor advances beyond the completion point of a dependent projection.

## Mechanism

`MIGRATIONS_AFTER_ROLLBACK` (`lib/api/dbbuilder.py:16-43`) lists migrations 0002, 0003, …, 0016 — but **not** `0001.create_and_populate_address_events`. `ROLLBACKABLE_TABLES` (`dbbuilder.py:44-48`) includes `address_events`, so it is deleted on rollback. `0002.create_and_populate_parsed_events` (`lib/api/migrations/0002...:71-80`) copies `parsed_events` from the full `ledger_db.messages`, advancing the cursor to the Ledger tip. `apiwatcher.get_next_event_to_parse` (`lib/api/apiwatcher.py:226-231`) then selects `WHERE message_index > cursor` — so after the cursor is at the tip, no catch-up work is seen, and the deleted `address_events` associations are never rebuilt.

**Result:** `address_events` is deleted, the cursor is advanced, and the migration that rebuilds it (0001) is not re-applied. Address history served by the API (`/v2/addresses/.../events`, `/v2/addresses/.../credits`) is incomplete until a full State rebuild.

## Expected behavior

After refresh, or rollback followed by catch-up, address history should contain the associations for all relevant Ledger events through the recorded State cursor. If a projection has not been rebuilt, its missing interval must remain replayable. A successful HTTP response should not silently omit confirmed history that remains available in the Ledger.

## Affected conditions

- **Automatic refresh:** an established Ledger beyond `BLOCK_FIRST`, a State version mismatch, `FORCE` false, an applicable v11.4.0 upgrade action, and address history behind the Ledger. Exists for mainnet, testnet3, testnet4, and signet.
- **Full fallback:** State has copied events at or beyond the rollback point; the Ledger already contains the replacement/reparsed suffix; and the full path is selected (e.g. missing incremental marker or depth > the stock 1,000-block threshold).
- A synchronized refresh is a passing control. A rollback that also removes the suffix from the Ledger retains prefix history only (also a passing control).

## Evidence

The finding was reproduced with real migrations, old-version upgrade inputs, actual reorg detection, and API checks (full evidence bundle available on request). Observed:

| Case | Address associations after action/catch-up | Replay work | Control |
|---|---:|---:|---|
| Lagging upgrade, each of four networks | 4 of 12 | 0 | Synchronized upgrade retains 12 of 12 |
| Partially copied mainnet State | 7 of 12 | 0 | Old-version continuation catches up normally |
| Legacy full fallback, replacement suffix present | 4 of 12 | 0 | Incremental rollback restores 12 of 12 |
| Full fallback at depth 1,002 | 4 of 12 | 0 | Stock threshold unchanged at 1,000 |
| Ledger suffix removed, no replacement | 4 prefix associations | 0 | Correct prefix-only result |

In the lagging upgrade, the cursor moved 5 → 17 while ordinary CREDIT associations remained 1 of 3. After full fallback, the replacement recipient had zero address CREDIT events while the direct Ledger-credit query returned two. The base implementation reproduced the full-rollback omission as well, confirming its inheritance.

A fresh-process recovery test used the actual Flask routes `/v2/addresses/events?addresses=...&event_name=CREDIT` and `/v2/addresses/<address>/credits`. Both returned HTTP 200. Missing event IDs `[9, 15]` survived restart, another new fixture block, and another refresh. A full State rebuild restored them.

## Root cause

| Location | Role |
|---|---|
| `lib/config.py:43-83` | New automatic State refresh actions |
| `lib/parser/check.py:199-220`, `lib/api/apiserver.py:645-661` | Version dispatch and real refresh callback |
| `lib/api/dbbuilder.py:16-48, 256-289` | Re-applied migration list omits address-history builder |
| `lib/api/migrations/0002...:71-80` | Copies parsed-event cursor through complete Ledger history |
| `lib/api/dbbuilder.py:116-124` | Prunes address history from the rollback point |
| `lib/api/apiwatcher.py:226-239, 784-794` | Selects only messages beyond the advanced cursor |

## Proposed fix

Make address-history completion part of the refresh/full-fallback invariant. Rebuild or backfill address associations to the same checkpoint as the other copied projections before publishing that checkpoint as fully parsed. Use a shared, runtime-equivalent association resolver, including UTXO-owner associations. Keep the incremental fast path separate.

Adding the existing address builder to the migration list addresses the demonstrated ordinary-address omission, but is not a complete equivalence proof: the fresh builder has inherited owner-alias differences. Alternatively, bound every rebuilt projection below the rollback point and replay the suffix. Do not merely rewind the global cursor after copying balances/counters to the tip, which can reapply already-copied state.

Also address already-affected State DBs: their recorded version can already be 11.4.0, so version-mismatch dispatch alone will not repair them. A dedicated repair migration/completion marker or an explicit supported State rebuild procedure is needed. This repair should leave Ledger messages and consensus history unchanged.

## Required regression acceptance

Retain all four upgrade networks, synchronized and partial-copy inputs, real changed-suffix watcher detection, natural marker/depth selection, and the Ledger-pruned control. Assert API-visible event IDs against Ledger data after catch-up and after restart. A repair test must recover missing associations without changing balances or double-counting events. Add UTXO-owner cases before claiming complete projection equivalence.

## Impact limits

Ledger messages remained intact; the expected balance was retained, and later balances stayed correct during repair checks. No asset loss, irreversible data loss, production incident, or live Bitcoin reorg was demonstrated. This is an API State DB projection issue, not a consensus or fund-loss bug.

---

*Credit: source-level analysis by OpenAI GPT-6 Astra Max; independent validation by Kimi K3; research coordination by John A Zoidburg.*

### LINKED ISSUE #3504: Early health readiness can become true before the public API finishes initialization
# Early health readiness can become true before the public API finishes initialization

**Area:** Operational / lifecycle
**Affected release:** `230302d303b9c25ba01e45edeccba0d30a2c188f` (v11.4.0, PR #3499)
**Master comparison:** `67e10db3ee266068c1effc4e83653df39ace5ca8`

## Summary

The dedicated health listener now starts before State maintenance and WSGI construction. Its readiness predicate can report ready while WSGI construction remains incomplete and the public-ready value is still zero. API-only mode takes this path without backend readings; normal mode can also take it once the shared backend height and State readings are available. An orchestrator can therefore receive a readiness signal that has not established public-service availability.

The base created the health object *after* WSGI construction, so this is a release-introduced exposure.

## Expected behavior

Early liveness should remain available during maintenance. Readiness should remain false until the required public API initialization/serving conditions have completed, then also reflect maintenance, watcher failure, lag, and saturation. API-only mode may bypass backend-lag checks but should not bypass service initialization.

## Root cause

`lib/api/apiserver.py:731-802` starts health, runs maintenance, installs shared backend height, constructs WSGI, and only then updates public readiness. `lib/api/healthz_server.py:259-283` combines maintenance/lag/failure/saturation conditions **without a service-initialized condition**. Its API-only branch (`healthz_server.py:359-368`, `_compute_caught_up`) immediately reports caught-up status (`return True, None, None, None, None`).

## Observed behavior

| Release configuration | Before migrations | During migration/refresh | Inside unfinished WSGI constructor | Public-ready value there |
|---|---|---|---|---:|
| API-only, backend known | Ready | Unready | Ready | 0 |
| API-only, backend unknown | Ready | Unready | Ready | 0 |
| Normal, backend/State readings available | Unready | Unready | Ready | 0 |
| Normal, backend unknown | Unready | Unready | Unready | 0 |

The actual base startup was tested with analogous stubs: its health object did not exist during migrations or inside WSGI construction; it was created afterward. This confirms the early-start exposure introduced by the release.

## Proposed fix

Introduce an explicit service lifecycle condition, initially false, that readiness requires in both modes. Publish it only after the relevant backend has completed initialization required to serve requests, and clear it before teardown/on failure. Preserve early liveness and the existing reasons for maintenance, watcher failure, lag, and saturation.

The existing public-ready value would close some early windows, but should not be treated as complete worker availability without validation: it is set before `wsgi_server.run()`, and Gunicorn workers still have their own initialization. Dispatcher attachment alone is also insufficient because Gunicorn/Werkzeug return no shared dispatcher. Use meaningful per-backend readiness hooks rather than moving the maintenance listener later.

## Required regression acceptance

Cover pre-maintenance, maintenance, delayed construction, worker initialization, successful serving, startup error, and shutdown for each supported WSGI backend. Keep known/unknown-backend controls in both modes. Verify early liveness remains available while readiness stays false.

## Impact limits

This demonstrates the production lifecycle's allowed state transition under controlled scheduling, not real HTTP availability, OS scheduling frequency, a production routing incident, or an outage duration. Operational/availability concern, not a security vulnerability.

---

*Credit: source-level analysis by OpenAI GPT-6 Astra Max; independent validation by Kimi K3; research coordination by John A Zoidburg.*

### LINKED ISSUE #3505: Core's v11.4.0 sdist cannot build a wheel because its OpenAPI force-include is outside the archive
# Core's v11.4.0 sdist cannot build a wheel because its OpenAPI force-include is outside the archive

**Area:** Packaging — release-blocking for the source-distribution channel
**Affected release:** `230302d303b9c25ba01e45edeccba0d30a2c188f` (v11.4.0, PR #3499)
**Master comparison:** `67e10db3ee266068c1effc4e83653df39ace5ca8`

## Summary

The new wheel rule includes `../openapi.json` from the repository parent directory. The Core sdist does not contain that file or an archive-local replacement. Building a wheel from a fresh extraction therefore fails. A direct repository-to-wheel build succeeds because the parent resource exists there.

## Expected behavior

The sdist should contain the source inputs required to build its wheel. The resulting wheel should include the same OpenAPI resource as a direct repository build, without depending on a neighboring checkout.

## Root cause

`counterparty-core/pyproject.toml:45-49` configures an unconditional wheel force-include from `../openapi.json`:

```toml
[tool.hatch.build.targets.wheel.force-include]
"../openapi.json" = "counterpartycore/openapi.json"
```

This path is valid in the repository layout but not in a standalone sdist extraction.

## Reproduction

Build an sdist, extract it into a fresh directory with no neighboring repository files, then build its wheel:

```sh
python -m build --sdist --no-isolation          # from counterparty-core/
python -m pip wheel --no-index --no-deps --no-build-isolation --no-cache-dir dist/*.tar.gz
```

## Observed results

| Operation | Release | Master base |
|---|---|---|
| Repository → sdist | Succeeds | Succeeds |
| Fresh sdist → wheel, `build` frontend | **Fails** | Succeeds |
| Sdist → wheel, pip frontend | **Fails** | Succeeds |
| Direct repository → wheel | Succeeds; correct OpenAPI bytes included | (control) |

The release sdist contains zero OpenAPI members. The wheel failure is:

```text
FileNotFoundError: Forced include not found: .../openapi.json
```

## Proposed fix

Make the OpenAPI resource an input carried inside the sdist and consumed through an archive-local path when building its wheel. An in-package generated resource or an explicit Hatch build hook can preserve the repository's single source of truth. Remove the unconditional dependence on an absent parent file. Merely adding the file to the sdist while leaving the same parent-path force-include is insufficient unless the build rule is also adapted.

## Required regression acceptance

Build an sdist, extract it into a fresh directory with no neighboring repository files, then build its wheel. Run the same check through pip with cache and dependency resolution disabled. Assert the wheel contains the expected spec bytes, and retain a direct repository-wheel control.

## Environment and limits

Verified with Python 3.11.11, Hatchling 1.32.0, hatch-requirements-txt 0.4.1, build 1.6.0, and pip 26.2.1 on macOS ARM64. Both frontends used that resolved declared backend. This does not claim identical failure behavior in every historical/future Hatchling release, failure of direct wheels or already-deployed containers, or a published-installation incident.

---

*Credit: source-level analysis by OpenAI GPT-6 Astra Max; independent validation by Kimi K3; research coordination by John A Zoidburg.*

### LINKED ISSUE #3506: Release checklist: align the Compose image and make its test verify the candidate executable
# Release checklist: align the Compose image and make its test verify the candidate executable

**Area:** Release checklist / delivery
**Affected candidate:** `230302d303b9c25ba01e45edeccba0d30a2c188f` (v11.4.0, PR #3499)
**Master base:** `67e10db3ee266068c1effc4e83653df39ace5ca8`

## Summary

The release source and upgrade instructions target 11.4.0, but Compose selects 11.3.0. The Docker integration's version replacement consequently matches nothing at this head. Updating the Compose version alone would make that same test substitute 11.0.4, so it still would not demonstrate that the candidate executable was exercised.

The PR is an open draft, so this should be treated as unfinished release alignment rather than evidence of an already-deployed regression.

## Values at the pinned head

| Item | Value |
|---|---|
| `config.VERSION_STRING` | `11.4.0` |
| Compose image | `counterparty/counterparty:v11.3.0` |
| Test search string | `image: counterparty/counterparty:v11.4.0` |
| Search occurrences | `0` |
| Test replacement, if the search matched | `image: counterparty/counterparty:v11.0.4` |

After launching the generated Compose file, the test asserts `server_ready` and `network == regtest`. It does not later upgrade to the reviewed head, build that head into the tested image, or assert its version/commit.

## Sources

- `docker-compose.yml:8-9`
- `counterpartycore/test/integrations/dockercompose_test.py:55-85, 109-115`
- API root version/commit fields: `lib/api/apiserver.py:92-107`

## Proposed fix

Update the executable image tag/digest as part of final release preparation. Have the integration explicitly select an image built from the candidate commit and verify its runtime `version` and `current_commit`, backed by the intended image digest/build provenance. If retaining an old-version scenario, identify the starting and target images, perform the upgrade, then assert the target identity and behavior. Readiness alone is not an identity assertion.

Use isolated, test-owned containers and cleanup when revising the integration; the current test includes host-wide prune operations.

## Required acceptance

Assert the expected image selection before execution; assert the candidate identity after startup; fail clearly on a zero-match or unintended replacement. Keep any compatibility/upgrade scenario distinct from candidate-image validation.

## Impact and limits

Users following the current Compose file could select an older executable than intended, and this test cannot prove candidate-image correctness. No running image or production deployment was inspected. The v11.3 bootstrap snapshot-data pin is a separate data-artifact choice and is not classified as an executable-image defect. The old-version substitution/test design is inherited; its mismatch with the candidate version is the release-checklist concern.

---

*Credit: source-level analysis by OpenAI GPT-6 Astra Max; independent validation by Kimi K3; research coordination by John A Zoidburg.*

### LINKED ISSUE #3507: Document the new 409 response for issuance and fairminter composition in OpenAPI
# Document the new 409 response for issuance and fairminter composition in OpenAPI

**Area:** Documentation / API contract
**Affected release:** `230302d303b9c25ba01e45edeccba0d30a2c188f` (v11.4.0, PR #3499)
**Master base:** `67e10db3ee266068c1effc4e83653df39ace5ca8`

## Summary

The release returns HTTP 409 when a supported asset composition conflicts with parsed pending asset changes. The prose describes this behavior, but the affected OpenAPI operations declare only 200. Schema-driven consumers cannot discover this known response from the operation contract.

## Expected behavior

Operations that can raise the new composition-conflict exception should document their 409 response and error-body schema. This does not require adding 409 to every compose operation.

## Details

At this head there are **197 operations and zero operation-level 409 declarations**. The direct callers of `_compose_with_pending_asset_guard` are:

| Operation | Current responses |
|---|---|
| `GET /v2/addresses/{address}/compose/issuance` | `200` |
| `GET /v2/addresses/{address}/compose/fairminter` | `200` |

`compose_fairmint` does not call this guard. The conflict requires the applicable validation path and a parsed pending conflict; not every request to these operations returns 409.

## Source locations

- `lib/api/compose.py:203-234, 420-456, 644-727`
- `lib/api/apiserver.py:488-489`
- `counterpartycore/test/integrations/regtest/genapidoc.py:308-339`

The generator initializes a 200 response, then attaches example content to that response; it discards the observed example status. Updating only the generated JSON would therefore be vulnerable to being lost on regeneration.

## Proposed fix

Add error-response metadata for the two guarded operations in the generator/source and regenerate the artifact. A reusable response can describe an object containing the string `error` field returned by the handler. For example:

```yaml
"409":
  description: A parsed pending asset change conflicts with this composition.
  content:
    application/json:
      schema:
        type: object
        required: [error]
        properties:
          error:
            type: string
```

This is proposed schema content, not an implemented change. Retain the existing success response and prose.

## Required acceptance

Regeneration must preserve the 409 declarations for both guarded operations. Add a schema-level assertion tied to their operation IDs and a check of the actual error shape. Do not blindly add the response to unrelated compose operations.

## Limits

No conflicting transaction or client failure was executed. This is not a claim that the OpenAPI document is syntactically invalid or that the runtime conflict rejection is wrong. Suitable for a documentation follow-up or batching with related API-contract work.

---

*Credit: source-level analysis by OpenAI GPT-6 Astra Max; independent validation by Kimi K3; research coordination by John A Zoidburg.*

ouziel-slama:
Fixes the five non-security findings [reported on #3499](https://github.com/CounterpartyXCP/counterparty-core/pull/3499#issuecomment-5564446623) by John A. Zoidburg. All five were verified against this head before being fixed.

Closes #3503, closes #3504, closes #3505, closes #3506, closes #3507.

## #3503 — Address history lost past the replay cursor

A State DB refresh, and the full rebuild that a deep or ineligible rollback still falls back to, re-applies migration `0002`, which repopulates `parsed_events` from the *entire* Ledger history and so moves the replay cursor to the Ledger tip. `address_events` is not among the tables `MIGRATIONS_AFTER_ROLLBACK` rebuilds — and deliberately so: re-applying `0001` would cost as much as a full State DB build and would drop the UTXO-owner aliases only `apiwatcher.update_address_events()` records. The result was that everything the State DB had not parsed yet, and after a rollback everything `rollback_tables()` had just deleted, was declared parsed with no associations behind it. `get_next_event_to_parse()` never came back for it, and `/v2/addresses/<address>/events`, `/credits` and `/debits` returned `200` while silently omitting confirmed history.

`refresh_state_db()` and `full_rollback_state_db()` now backfill the missing associations before publishing the new cursor, through the same resolver the watcher uses. The repair is defined by what is *missing* — every address-bearing Ledger event with no association at all — rather than by what this particular run would have removed:

- nothing is counted twice: by construction no association exists for the events it re-derives, and `rollback_tables()` prunes whole blocks, so an event either kept all of its rows or lost all of them;
- existing rows, including the UTXO-owner aliases, are left alone;
- the one-time v11.4.0 refresh therefore also closes the equivalent holes left behind by earlier releases' refreshes, which the issue asks for.

The detection query is index-driven on both sides (`SEARCH ledger_db.messages USING COVERING INDEX messages_event_idx`, `SEARCH address_events USING COVERING INDEX address_events_event_index_idx`), and the missing set is materialised into a temp table first — SQLite leaves a query's result undefined if the table it is reading is written to while it is still open.

Ledger data, balances and consensus are untouched.

Three regression tests, two of which fail on `develop`: the refresh path (a State DB behind the Ledger), the full-rollback path, and a no-op assertion that a complete table gains nothing.

## #3504 — Readiness before the public API serves

The dedicated listener starts before State DB maintenance — that is what it is for — but the readiness predicate had no service-lifecycle condition. Between the end of maintenance and `wsgi_server.run()` (watcher startup, Flask app construction, binding the socket) a caught-up ledger was enough to report ready, and on an `--api-only` node, which skips the backend-lag comparison entirely, readiness was green from the first sample.

`HealthSampler` takes a `serving_provider`; `run_apiserver` wires it to `server_ready_value.value == 1`, which is set immediately before `wsgi_server.run()` and reset to `2` on teardown. Readiness reports `503 starting` until then, in both modes. Liveness is unchanged and stays `200` — a pod that is still coming up must not be killed. An unreadable signal sheds rather than guesses.

## #3505 — sdist could not build its own wheel

`[tool.hatch.build.targets.wheel.force-include]` pointed at `../openapi.json`, valid in the repository but absent from an extracted sdist, so `pip wheel` on a fresh sdist failed with `FileNotFoundError: Forced include not found`. Reproduced with both the `build` and `pip` frontends before the fix.

A Hatch build hook (`counterparty-core/hatch_build.py`) now carries the document into the sdist and resolves it from whichever layout the wheel build is running in — archive-local first, repository parent second. Verified end to end: sdist → wheel and repository → wheel both succeed and embed byte-identical spec content (`sha256 73ce098f…`). A missing document is a warning rather than a hard failure, so a checkout that has never generated it stays installable.

## #3506 — Compose pinned the previous release

`docker-compose.yml` selected `v11.3.0` while `config.VERSION_STRING` was `11.4.0`, so the Docker integration's version substitution matched nothing and the scenario exercised whatever the stale tag pointed at. The tag is updated to `v11.4.0` and the test now asserts the Compose file pins the release under test before substituting, so the same drift fails loudly at the next release instead of passing quietly.

Not addressed here: asserting the candidate image's runtime `version`/`current_commit`. That needs an image built from the candidate commit, which nothing in the integration workflow produces today; it belongs with the release/publish pipeline rather than this test.

## #3507 — 409 missing from the API contract

`compose_issuance` and `compose_fairminter` go through `_compose_with_pending_asset_guard` and can return the `409` added in #3490, but the document declared `200` alone across all 197 operations — including in the release notes, which already claimed otherwise. The generator only ever observes the happy path, so the declaration comes from `genapidoc.ERROR_RESPONSES` and survives regeneration. It is added to those two operations only; `compose_fairmint` and the other compose routes do not call the guard.

Two tests pin it: one on the operation IDs and the error-body shape, one that re-derives the guard's callers from the source so the mapping cannot silently fall out of step.

## Verification

- Full unit suite: 1898 passed, 1 pre-existing unrelated failure (`chain_halt_test.py::test_c1_rust_never_emits_a_none_dispenser_slot`, also failing on a clean checkout).
- `ruff check` / `ruff format --check` clean.
- Packaging verified by building both distributions, as described above.

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/addressevents.py
+++ b/counterparty-core/counterpartycore/lib/api/addressevents.py
@@ -1,0 +1,101 @@
+"""Address associations for Ledger events.
+
+``address_events`` is the projection behind ``/v2/addresses/<address>/events``,
+``/credits`` and ``/debits``. It is derived in three places -- the watcher as
+events stream in, migration ``0001`` on a full State DB build, and
+``dbbuilder.backfill_address_events()`` when a refresh or a full rollback
+advances the replay cursor past events it never associated -- and all three
+have to agree, down to the UTXO-owner alias, or the API answers differently
+depending on how the State DB was built.
+
+They therefore share these rules rather than each restating them. This lives
+outside ``apiwatcher`` because ``apiwatcher`` imports ``dbbuilder``, so
+``dbbuilder`` cannot import it back.
+"""
+
+import json
+
+from counterpartycore.lib.parser import utxosinfo
+
+EVENTS_ADDRESS_FIELDS = {
+    "NEW_TRANSACTION": ["source", "destination"],
+    "DEBIT": ["address"],
+    "CREDIT": ["address"],
+    "ENHANCED_SEND": ["source", "destination"],
+    "MPMA_SEND": ["source", "destination"],
+    "SEND": ["source", "destination"],
+    "ASSET_TRANSFER": ["source", "issuer"],
+    "SWEEP": ["source", "destination"],
+    "ASSET_DIVIDEND": ["source"],
+    "RESET_ISSUANCE": ["source", "issuer"],
+    "ASSET_ISSUANCE": ["source", "issuer"],
+    "ASSET_DESTRUCTION": ["source"],
+    "OPEN_ORDER": ["source"],
+    "ORDER_MATCH": ["tx0_address", "tx1_address"],
+    "BTC_PAY": ["source", "destination"],
+    "CANCEL_ORDER": ["source"],
+    "ORDER_EXPIRATION": ["source"],
+    "ORDER_MATCH_EXPIRATION": ["tx0_address", "tx1_address"],
+    "OPEN_DISPENSER": ["source", "origin", "oracle_address"],
+    "DISPENSER_UPDATE": ["source"],
+    "REFILL_DISPENSER": ["source", "destination"],
+    "DISPENSE": ["source", "destination"],
+    "BROADCAST": ["source"],
+    "BURN": ["source"],
+    "NEW_FAIRMINT": ["source"],
+    "NEW_FAIRMINTER": ["source"],
+    "ATTACH_TO_UTXO": ["source", "destination_address"],
+    "DETACH_FROM_UTXO": ["source_address", "destination"],
+    "UTXO_MOVE": ["source_address", "destination_address"],
+    "OPEN_POOL": ["source"],
+    "POOL_UPDATE": [],
+    "NEW_POOL_DEPOSIT": ["source"],
+    "NEW_POOL_WITHDRAWAL": ["source"],
+    "POOL_MATCH": ["source"],
+}
+
+
+def search_address_from_utxo(state_db, utxo):
+    cursor = state_db.cursor()
+    sql = "SELECT utxo_address FROM balances WHERE utxo = ? LIMIT 1"
+    cursor.execute(sql, (utxo,))
+    address = cursor.fetchone()
+    if address is not None:
+        return address["utxo_address"]
+    return None
+
+
+def update_address_events(state_db, event):
+    if event["event"] not in EVENTS_ADDRESS_FIELDS:
+        return
+    event_bindings = json.loads(event["bindings"])
+    cursor = state_db.cursor()
+    for field in EVENTS_ADDRESS_FIELDS[event["event"]]:
+        if field not in event_bindings:
+            continue
+        address = event_bindings[field]
+        sql = """
+            INSERT INTO address_events (address, event_index, block_index, event)
+            VALUES (:address, :event_index, :block_index, :event)
+            """
+        cursor.execute(
+            sql,
+            {
+                "address": address,
+                "event_index": event["message_index"],
+                "block_index": event["block_index"],
+                "event": event["event"],
+            },
+        )
+        if utxosinfo.is_utxo_format(address):
+            utxo_address = search_address_from_utxo(state_db, address)
+            if utxo_address is not None:
+                cursor.execute(
+                    sql,
+                    {
+                        "address": utxo_address,
+                        "event_index": event["message_index"],
+                        "block_index": event["block_index"],
+                        "event": event["event"],
+                    },
+                )
--- a/counterparty-core/counterpartycore/lib/api/apiserver.py
+++ b/counterparty-core/counterpartycore/lib/api/apiserver.py
@@ -697,6 +697,7 @@ class ConnectionPoolMonitor(threading.Thread):
 def run_apiserver(
     args, server_ready_value, stop_event, shared_backend_height, parent_pid, log_stream
 ):
+    api_owner_pid = os.getpid()
     logger.info("Starting API Server process...")
 
     def handle_interrupt_signal(_signum, _frame):
@@ -838,7 +839,10 @@ def run_apiserver(
             memory_profiler.stop_memory_profiler()
 
         logger.info("API Server stopped.")
-        server_ready_value.value = 2
+        # Retiring Gunicorn workers also unwind through this finally block.
+        # Only the API owner can mark the shared server state as stopped.
+        if os.getpid() == api_owner_pid:
+            server_ready_value.value = 2
 
 
 # This thread is used for the following two reasons:
--- a/counterparty-core/counterpartycore/lib/api/apiserver.py
+++ b/counterparty-core/counterpartycore/lib/api/apiserver.py
@@ -749,6 +749,11 @@ def run_apiserver(
                     config.DEFAULT_HEALTHZ_SATURATION_GRACE_SECONDS,
                 ),
                 stop_event=stop_event,
+                # Readiness must not go green during the startup this listener
+                # deliberately runs ahead of: `server_ready_value` is 0 until
+                # the WSGI server is built and about to run, and 2 once it is
+                # stopping (issue #3504).
+                serving_provider=lambda: server_ready_value.value == 1,
             )
             health_server.start()
 
--- a/counterparty-core/counterpartycore/lib/api/apiwatcher.py
+++ b/counterparty-core/counterpartycore/lib/api/apiwatcher.py
@@ -7,7 +7,7 @@ import apsw
 
 from counterpartycore.lib import config
 from counterpartycore.lib.api import dbbuilder
-from counterpartycore.lib.parser import utxosinfo
+from counterpartycore.lib.api.addressevents import update_address_events
 from counterpartycore.lib.utils import database, hashcodec
 from counterpartycore.lib.utils.helpers import deadline_timeout, format_duration
 
@@ -114,43 +114,6 @@ UPDATE_EVENTS_ID_FIELDS = {
     "ADDRESS_OPTIONS_UPDATE": ["address"],
     "FAIRMINTER_UPDATE": ["tx_hash"],
     "POOL_UPDATE": ["asset_a", "asset_b"],
-}
-
-EVENTS_ADDRESS_FIELDS = {
-    "NEW_TRANSACTION": ["source", "destination"],
-    "DEBIT": ["address"],
-    "CREDIT": ["address"],
-    "ENHANCED_SEND": ["source", "destination"],
-    "MPMA_SEND": ["source", "destination"],
-    "SEND": ["source", "destination"],
-    "ASSET_TRANSFER": ["source", "issuer"],
-    "SWEEP": ["source", "destination"],
-    "ASSET_DIVIDEND": ["source"],
-    "RESET_ISSUANCE": ["source", "issuer"],
-    "ASSET_ISSUANCE": ["source", "issuer"],
-    "ASSET_DESTRUCTION": ["source"],
-    "OPEN_ORDER": ["source"],
-    "ORDER_MATCH": ["tx0_address", "tx1_address"],
-    "BTC_PAY": ["source", "destination"],
-    "CANCEL_ORDER": ["source"],
-    "ORDER_EXPIRATION": ["source"],
-    "ORDER_MATCH_EXPIRATION": ["tx0_address", "tx1_address"],
-    "OPEN_DISPENSER": ["source", "origin", "oracle_address"],
-    "DISPENSER_UPDATE": ["source"],
-    "REFILL_DISPENSER": ["source", "destination"],
-    "DISPENSE": ["source", "destination"],
-    "BROADCAST": ["source"],
-    "BURN": ["source"],
-    "NEW_FAIRMINT": ["source"],
-    "NEW_FAIRMINTER": ["source"],
-    "ATTACH_TO_UTXO": ["source", "destination_address"],
-    "DETACH_FROM_UTXO": ["source_address", "destination"],
-    "UTXO_MOVE": ["source_address", "destination_address"],
-    "OPEN_POOL": ["source"],
-    "POOL_UPDATE": [],
-    "NEW_POOL_DEPOSIT": ["source"],
-    "NEW_POOL_WITHDRAWAL": ["source"],
-    "POOL_MATCH": ["source"],
 }
 
 EXPIRATION_EVENTS_OBJECT_ID = {
@@ -406,52 +369,6 @@ def event_to_sql(event, ledger_db=None):
     if event["command"] in ["update", "parse"]:
         return update_event_to_sql(event, ledger_db=ledger_db)
     return None, []
-
-
-def search_address_from_utxo(state_db, utxo):
-    cursor = state_db.cursor()
-    sql = "SELECT utxo_address FROM balances WHERE utxo = ? LIMIT 1"
-    cursor.execute(sql, (utxo,))
-    address = cursor.fetchone()
-    if address is not None:
-        return address["utxo_address"]
-    return None
-
-
-def update_address_events(state_db, event):
-    if event["event"] not in EVENTS_ADDRESS_FIELDS:
-        return
-    event_bindings = json.loads(event["bindings"])
-    cursor = state_db.cursor()
-    for field in EVENTS_ADDRESS_FIELDS[event["event"]]:
-        if field not in event_bindings:
-            continue
-        address = event_bindings[field]
-        sql = """
-            INSERT INTO address_events (address, event_index, block_index, event)
-            VALUES (:address, :event_index, :block_index, :event)
-            """
-        cursor.execute(
-            sql,
-            {
-                "address": address,
-                "event_index": event["message_index"],
-                "block_index": event["block_index"],
-                "event": event["event"],
-            },
-        )
-        if utxosinfo.is_utxo_format(address):
-            utxo_address = search_address_from_utxo(state_db, address)
-            if utxo_address is not None:
-                cursor.execute(
-                    sql,
-                    {
-                        "address": utxo_address,
-                        "event_index": event["message_index"],
-                        "block_index": event["block_index"],
-                        "event": event["event"],
-                    },
-                )
 
 
 def update_all_expiration(state_db, event):
--- a/counterparty-core/counterpartycore/lib/api/dbbuilder.py
+++ b/counterparty-core/counterpartycore/lib/api/dbbuilder.py
@@ -208,6 +208,96 @@ def record_balances_copied_block(state_db):
         )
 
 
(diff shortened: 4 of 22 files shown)


john-a-zoidburg:
Reviewed #3509 at `c19a9992`. The State backfill, source-distribution builds and OpenAPI declarations passed focused validation. I found one remaining Gunicorn readiness regression.

A normally retiring worker unwinds through [`run_apiserver()`'s outer `finally`](https://github.com/CounterpartyXCP/counterparty-core/blob/c19a9992bd5bc36c0abd748ee4ebf37e418f4193/counterparty-core/counterpartycore/lib/api/apiserver.py#L841) and marks the shared API state as **stopped** (`server_ready_value.value = 2`). The [new readiness check](https://github.com/CounterpartyXCP/counterparty-core/blob/c19a9992bd5bc36c0abd748ee4ebf37e418f4193/counterparty-core/counterpartycore/lib/api/apiserver.py#L756) accepts only **ready** (`1`), so it reports the entire instance as unready even though the master and replacement workers continue serving requests. The health listener's PID guard does not cover this assignment.

In an isolated Linux test with one Gunicorn worker, API-only mode, `max_requests=1` and zero jitter, a benign request triggered retirement. The replacement worker successfully served another request and liveness remained 200, while readiness returned **503 `starting`** across four observations spanning 3.3 seconds. The equivalent #3499 control stayed ready. This can remove a functioning instance from readiness-based routing; the default Waitress backend is not affected by this retirement path.

The proposed fix in [#3510](https://github.com/CounterpartyXCP/counterparty-core/pull/3510), targeting this branch, restricts the stopped-state write to the API owner PID. The actual patched package retained readiness through repeated worker replacements and preserved the tested shutdown, startup-error and Waitress/Werkzeug behavior. The new lifecycle tests fail on the unpatched head in both worker-retirement cases and pass all eight cases with the fix; the six owner shutdown/startup-failure controls pass on both versions.

adamkrellenstein (approved):

ouziel-slama:
PR merged, thank you @john-a-zoidburg !

PR #3512, opened 2026-09-09
Bound database-close mutex waits during API shutdown

adamkrellenstein:
The watcher and node-status shutdown paths acquire `db_lock` without a timeout before their bounded joins. The worker holds the same lock during SQLite close, so a slow close can consume the entire parent shutdown budget before the join is reached.

Bound mutex acquisition by the supplied deadline. If close still owns the lock, log the condition and let the daemon finish independently; never call SQLite interrupt without the lock. This preserves serialization of interrupt and close.

Follow-up to #3486. Production still emitted forced-kill logs on September 7; this fixes a reproducible deadline gap, but the logs do not prove it caused those particular kills.

Validation: a dependency-free harness exercising the exact production stop methods passes both contended-lock cases; Ruff passes. Added parametrized pytest regressions. Full repository pytest execution is pending: a fresh test environment cannot resolve the unreleased counterparty-rs==11.4.0 requirement from the package registry. This PR stays draft until CI and review complete. No production deployment or protocol change is included.

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/apiwatcher.py
+++ b/counterparty-core/counterpartycore/lib/api/apiwatcher.py
@@ -976,16 +976,21 @@ class APIWatcher(threading.Thread):
         # Under the lock: apsw checks the connection is open and then calls
         # sqlite3_interrupt() while holding the GIL, but close() releases it
         # around sqlite3_close_v2(), so an unsynchronised interrupt can reach a
-        # handle that is already being freed. Held only for the duration of the
-        # interrupts, which never block, so the closing thread waits on it for
-        # microseconds and the join() below cannot deadlock against it.
-        with self.db_lock:
+        # handle that is already being freed. The closing thread also holds
+        # this lock during SQLite close, which can wait on disk I/O. Include
+        # lock acquisition in the shutdown budget; never interrupt without it.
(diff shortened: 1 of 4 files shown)


ouziel-slama:
@adamkrellenstein opened #3513 as an alternative. Same bug, fixed at its source rather than by bounding the wait.

The root of the problem is that the closing thread holds `db_lock` across the `close()` itself, which is disk-bound: the watcher's `state_db` is the last writer once the pools are closed, so closing it checkpoints the WAL and fsyncs. #3513 hands the connections off under the lock and closes them outside it, so both critical sections become a couple of assignments, `stop()` goes back to its original `with self.db_lock:`, and the only wait left is the join it already bounded.

That avoids the side effects of the timed acquire: the budget is no longer spent on a lock that cannot be used for anything, the early `return` that skips the join and both terminal log lines disappears, the double `wsgi_server.stop()` can no longer produce a spurious warning on a clean shutdown, and there is no new `consider-using-with` alert.

The regressions in #3513 fail on develop at 5.01s against a 0.05s deadline, on both threads.

ouziel-slama:
replaced by https://github.com/CounterpartyXCP/counterparty-core/pull/3513

PR #3514, opened 2026-09-12
Let an MPMA send pay a Taproot destination

### LINKED ISSUE #3208: Taproot addresses not supported for MPMA
I was playing around with MPMA and found out that it doesn't support taproot addresses.

`https://api.counterparty.io:4000/v2/addresses/bc1qy25gwr9vdrvxx0w35wa9vrpc3hc5h6rd0p56f4/compose/mpma?address=bc1qy25gwr9vdrvxx0w35wa9vrpc3hc5h6rd0p56f4&destinations=bc1p2q9t8xv4y3tf62lddgjkhzanma0xdr6a205hfq8a2r9f2n70eevszuvqsh&assets=A10207619717431880000&quantities=1&fee_per_kb=3000&encoding=taproot&return_psbt=true&memo=&memo_is_hex=False`

`{"error": "Address not supported by MPMA send: bc1p2q9t8xv4y3tf62lddgjkhzanma0xdr6a205hfq8a2r9f2n70eevszuvqsh"}`

droplister:
Closes #3208.

## What was blocking it

An MPMA's address lookup table packs every entry with `address.pack_legacy`, which emits a fixed 21 bytes — a one byte prefix plus a 20 byte hash or witness program. A Taproot witness program is 32 bytes, so it never fit. `mpmaencoding._decode_decode_lut` matched with a hard-coded `bytes_per_address = 21`, and `compose_mpma` refused the destination up front:

```python
if len(address.pack(destination)) > 22:
    raise exceptions.ComposeError(f"Address not supported by MPMA send: {destination}")
```

## Nothing new had to be designed

`taproot_support` (v11.0.0) already added a self-describing packing in `counterparty_rs` — `0x01`/`0x02` plus a hash, `0x03` plus a witness version and program — and enhanced sends (`enhancedsend.py:190`) and sweeps (`sweep.py:168`) have used it since. **MPMA was the last message type still packing a destination the legacy way.** It was left behind because its table is a *list*, and a list of variable-length entries needs lengths.

(Dispenser oracle addresses are the other `pack_legacy` caller. They are deliberately untouched here.)

The Rust packer needs no change at all.

## The encoding

Entries are emitted in groups of one address kind: a shared prefix, an entry length, a `uint16` count, then the entries with that prefix omitted — so the kind and the length are paid for once per group rather than once per address.

```
uint16 entry_count
per group:  uint8 prefix_len | prefix | uint8 entry_len | uint16 count | count × entry_len
```

The grouping is applied in `_encode_construct_lut`, before indices are handed out, because the send lists index into the table by position. Within a group the existing sorted order is preserved, so the table stays a deterministic function of the send list.

**Measured**, encoder to decoder, distinct addresses of each kind:

| destinations | grouped | per-address length byte | saving |
|---|---|---|---|
| 100 × Taproot | 3,208 | 3,502 | 8.4% |
| 500 × Taproot | 16,008 | 17,502 | 8.5% |
| 500 × P2WPKH | 10,008 | 11,502 | 13.0% |
| 500 mixed | 13,019 | 14,377 | 9.4% |

An MPMA composed with `encoding=taproot` carries that table in the witness, where it is discounted four to one.

## Gating

Everything is behind `mpma_taproot_support` (`minimum_version` 11.4.0, activation heights TBD). Below the activation height the table is written and read by the **original code path, untouched** — including the exception a truncated table raises, because `parse` records that in a status string, which makes even the error classification consensus-visible. An earlier revision of this branch let a new length check leak into that branch and turned `UnpackError("truncated data")` into `DecodeError`; the existing `test_unpack` caught it.

`/v2/transactions/unpack` now passes its `block_index` through to `mpma.unpack`, as it already did for every other message type. Without that the endpoint decoded a historical MPMA under the current rules and returned 503.

## Tests

Unit tests run with `REGTEST` set, where `protocol.enabled()` answers True for every change regardless of block index — so every hard-coded MPMA vector in the suite silently became a post-activation assertion. The six tests built on pre-activation encodings now say which era they are testing, via the existing `ProtocolChangesDisabled` helper.

New coverage:

- a Taproot destination composes, with the exact bytes pinned (two vectors)
- the same compose is still refused before activation
- every kind `address.pack` represents round-trips through the table — P2PKH, P2SH, P2WPKH, Taproot and **P2WSH**, which the same `pack_legacy` branch refused ("p2wsh still not supported for sending")
- the table does not depend on the order destinations were given in
- a truncated table is refused rather than read out of whatever bytes follow

`counterpartycore/test/units`: **1750 passed**, against 1744 on `develop` at the same commit, with the same 4 pre-existing failures and 31 pre-existing errors. `ruff check` and `ruff format` are unchanged from `develop`.

A regtest scenario step is appended to `scenario_19_mpma.py`. Its controls assert the credited balances rather than event indexes, so it does not shift the indexes the existing steps assert.

## Verified on a chain

Regtest, against this branch — an MPMA paying two Taproot destinations and one legacy address:

```
Block 113 - MPMA_SEND [... destination=bcrt1ps7 quantity=1000 status=valid memo=taproot airdrop]
Block 113 - MPMA_SEND [... destination=bcrt1pw9 quantity=2000 status=valid memo=taproot airdrop]
Block 113 - MPMA_SEND [... destination=mkbCDH3y quantity=3000 status=valid memo=taproot airdrop]

balance bcrt1ps7gfq0h0… : 1000 XCP
balance bcrt1pw9jnqadn… : 2000 XCP
```

`/v2/transactions/<hash>` unpacks it back off the chain correctly.

## The alternative

`feat/mpma-taproot-lenbyte` is the same change with a **one byte length per entry** instead of groups — same gate, same tests, same regtest verification, ~30 fewer lines, and it leaves the table's ordering alone (this branch reorders it by kind, which is what lets a group share a header). It is 8.5–13% larger at airdrop scale, per the table above.

I am happy to switch this PR to that branch if the smaller consensus surface is worth the bytes — the sizes are the only thing that differs.

### DIFF
--- a/counterparty-core/counterpartycore/lib/api/compose.py
+++ b/counterparty-core/counterpartycore/lib/api/compose.py
@@ -1172,7 +1172,7 @@ def unpack(db, datahex: str, block_index: int = None):
         # MPMA send
         elif message_type_id == messages.versions.mpma.ID:
             message_type_name = "mpma_send"
-            mpma_message_data = messages.versions.mpma.unpack(message)
+            mpma_message_data = messages.versions.mpma.unpack(message, block_index=block_index)
             message_data = []
             for asset_name, send_info in mpma_message_data.items():
                 message_data.append(
--- a/counterparty-core/counterpartycore/lib/messages/versions/mpma.py
+++ b/counterparty-core/counterpartycore/lib/messages/versions/mpma.py
@@ -29,9 +29,9 @@ def py34_tuple_append(first_elem, t):
 
 
 ## expected functions for message version
-def unpack(message):
+def unpack(message, block_index=None):
     try:
-        unpacked = _decode_mpma_send_decode(message)
+        unpacked = _decode_mpma_send_decode(message, block_index=block_index)
     except struct.error as e:
         raise exceptions.UnpackError("could not unpack") from e
     except (exceptions.AssetNameError, exceptions.AssetIDError) as e:
@@ -142,11 +142,12 @@ def compose(
 
     cursor = db.cursor()
 
-    for send in asset_dest_quant_list:
-        destination = send[1]
+    if not protocol.enabled("mpma_taproot_support"):
+        for send in asset_dest_quant_list:
+            destination = send[1]
 
-        if len(address.pack(destination)) > 22:
-            raise exceptions.ComposeError(f"Address not supported by MPMA send: {destination}")
+            if len(address.pack(destination)) > 22:
+                raise exceptions.ComposeError(f"Address not supported by MPMA send: {destination}")
 
     if memo and not isinstance(memo, str):
         raise exceptions.ComposeError("`memo` must be a string")
@@ -184,7 +185,7 @@ def compose(
 
 def parse(db, tx, message):
     try:
-        unpacked = unpack(message)
+        unpacked = unpack(message, block_index=tx["block_index"])
         status = "valid"
     except struct.error:
         status = "invalid: truncated message"
--- a/counterparty-core/counterpartycore/lib/utils/mpmaencoding.py
+++ b/counterparty-core/counterpartycore/lib/utils/mpmaencoding.py
@@ -4,9 +4,17 @@ import struct
 
 from bitstring import BitArray, ConstBitStream
 from counterpartycore.lib import config, exceptions, ledger
+from counterpartycore.lib.parser import protocol
 from counterpartycore.lib.utils import address
 
 logger = logging.getLogger(config.LOGGER_NAME)
+
+# Size of a legacy address lookup table entry: a one byte prefix and a 20 byte hash.
+LEGACY_BYTES_PER_ADDRESS = 21
+
+# A witness entry's length depends on the witness version as well as the 0x03 marker (20 bytes of
+# program for V0 P2WPKH, 32 for V1 Taproot), so the version belongs in the group key.
+WITNESS_PREFIX = 0x03
 
 ## encoding functions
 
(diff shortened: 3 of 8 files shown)


droplister:
Dropped the regtest scenario step in 4a64e66 — it was the one red check.

`$EVENT_INDEX_n` resolves to the *n-th event of the current block*, so appending a transaction to the shared scenario chain moved the positional event expectations of scenarios running after it: `scenario_20_fairminter`'s "Create subasset PARENTB" control failed on an ASSET_ISSUANCE whose event index had shifted by three.

Everything the step covered is still pinned:

- the unit tests assert the **exact wire bytes** of a Taproot lookup table, in both the mixed and all-Taproot shapes — which is the consensus-critical part
- the behaviour was verified on a real regtest chain (compose → broadcast → parse → credit, two Taproot destinations plus one legacy)

A scenario step is still worth adding; it should be placed by someone who can re-derive the fixture ordering it disturbs, rather than guessed at from here.

droplister:
Closing in favour of #3515, which carries the same change with a one-byte length per table entry instead of grouped headers.

On review the grouped table was not worth its cost for a message sent once per airdrop: it reorders the table by address kind (a second ordering rule encoder and decoder must share forever), the encoding is not canonical, the decoder has five consensus-visible failure branches instead of two, and it had a latent flaw — P2WPKH and P2WSH share a witness-v0 prefix but differ in length, so a table mixing them fragmented into a header per run (50 + 50 → 49 groups) and lost the whole saving.

The branch stays at `feat/mpma-taproot-grouped` in case the bytes ever matter; it could be layered on later behind its own gate.

PR #3516, opened 2026-09-13
Fix coinbase commit resolution for inscription reveals

adamkrellenstein:
Fix inscription reveal source resolution when the commit transaction is a coinbase. A coinbase has one null-prevout input, so checking for an empty input list incorrectly attempts to fetch a funding transaction that does not exist. Both the Rust parser and Python fallback now retain the reveal's actual prevouts for coinbase commits. Ordinary commit-parent resolution and recovery from failed RPC lookups are preserved.

Replace the empty-input coinbase fixture with the real structure, and cover complete, partial and failed prevout resolution. The Python regression uses a serialized coinbase decoded by the Rust extension and checks that source and input values agree across all resolution paths.

Also clarify the three v11.4.0 protocol changes: non-finite/unstorable broadcast numbers, negative broadcast fee fractions, and Ordinals metadata-map support. Their activation heights are unchanged.

Validation:
- Rust suite: 35 passed, 1 ignored.
- Python unit suite: 1,922 passed, 3 skipped.
- Python functional suite: 19 passed, run separately as in CI.
- Removing the Rust coinbase guard makes both new coinbase regression tests fail.
- Ruff and rustfmt checks pass for changed code; `git diff --check` passes.

Local Python tests use isolated temporary databases/logs and a fixed hostname for Yoyo audit logging to avoid a macOS reverse-DNS stall. A combined unit/functional invocation produced functional fixture setup errors; all functional tests pass in a separate invocation.

* [x] Double-check spelling and grammar
* [x] Check determinism of consensus code
* [x] Add tests for revised logic
* [x] Run local unit, functional and Rust suites (GitHub integration checks pending)
* [x] Update release notes
* [x] Check documentation needs (release notes cover this correction; no separate documentation change)

cc @ouziel-slama


john-a-zoidburg:
Reviewed #3516 at `7de16d7d8`, including the Rust parser, Python fallback and a source comparison with v11.3.0.

A coinbase contains one input with a null previous outpoint, so the previous empty-input check did not recognize it. The reveal fallback could then treat that placeholder as a funding reference and attempt a lookup that cannot succeed. The [Python check](https://github.com/CounterpartyXCP/counterparty-core/blob/7de16d7d81b0d9ff75d3f28ed6e78c5593ee2288/counterparty-core/counterpartycore/lib/backend/bitcoind.py#L955) and [Rust resolver](https://github.com/CounterpartyXCP/counterparty-core/blob/7de16d7d81b0d9ff75d3f28ed6e78c5593ee2288/counterparty-rs/src/indexer/bitcoin_client.rs#L741) now distinguish a coinbase with no funding parent from an actual lookup failure. Coinbase commits retain their ordinary prevouts, while funded commits retain the existing all-input rewrite and failure recovery.

In isolated Linux/Docker tests, I built the PR and its unmodified base separately with their stock native extensions and ran the same upstream regression cases. On `cafc4b672`, four coinbase cases failed on the expected null-parent reference or lookup, while three controls passed. With this fix, all seven passed.

The [new Python regression](https://github.com/CounterpartyXCP/counterparty-core/blob/7de16d7d81b0d9ff75d3f28ed6e78c5593ee2288/counterparty-core/counterpartycore/test/units/parser/reveal_source_test.py#L157) exercises the representation that the previous fixture missed: a serialized coinbase decoded by Rust. Its two inputs spend different output indices with different values, and the fixed resolver returns the expected source and 33,000-satoshi input total across fully resolved, unresolved and mixed cases. Parent lookups are controlled in this comparison, so the evidence is specifically for resolver behavior.

The complete backend and parser unit directories also passed (**460 tests**), along with the native Rust suite (**35 passed, 1 ignored**) and Python functionals (**19 passed**). No blocking findings from this review. The fix addresses the coinbase resolution issue, and I recommend merging this PR.

